Issue #16, opened 2022-10-13
define minimal interface every event detection algorithm will have to use


dkrako:
## Description of the problem
We will integrate a number of event detection algorithms.
If we don't define a common interface, the user will possibly have to workaround this when applying several detection methods.

## Description of a solution
The signature should use keyword arguments.
For convenience we should also define a minimal type signature with `Callable`.
Reference: https://docs.python.org/3/library/typing.html#callable

Parameter:
- position: np.ndarray
- velocity: np.ndarray

Return:
- list[events]

I'm not sure about the parameters.
There could be methods which just use one of those.
We need to decide on what makes most sense.

## Sample Code
A first proposal would be something like this:

Signature:
```
def event_detection_function(
    pos: np.ndarray,
    vel:np.ndarray,
) -> list[Event]:
```
Types:
```
Event = Tuple[int, int]
EventDetection = Callable[[np.ndarray, np.ndarray], list[Event]]
```

dkrako:
alternative event definition
```
class Event:
    def __init__(self, name, onset, offset):
    
    @property.getter
    def duration():
```

PR #46, opened 2022-10-26
Add event base classes

dkrako:
## Description

This PR adds the first event base classes.
This will help in unifying the output of the event detection functions.
Proposed in #16

## Implemented changes
- [x] Added `Event` class with `duration` property
- [x] Added child classes `Fixation` and `Saccade`
- [x] Added basic tests

## Type of change

- [x] This change requires a documentation update

## How Has This Been Tested?

- [x] Tests correct duration property for three different cases
- [x] Tests correct name attribute

## Checklist:

### DIFF
--- a/src/pymovements/events/base.py
+++ b/src/pymovements/events/base.py
@@ -1,15 +1,60 @@
+"""
+This module holds all the main Event classes used for event detection.
+"""
+
+
 class Event:
+    """
+    Base Event class.
+
+    Parameters
+    ----------
+    name: str
+        Name of event.
+    onset: int
+        Starting index of event (included).
+    offset: int
+        Ending index of event (excluded).
+
+    Attributes
+    ----------
+    name: str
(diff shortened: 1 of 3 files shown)


src/pymovements/events/base.py
```diff
@@ -0,0 +1,78 @@
...
+    def duration(self) -> int:
+        """
+        Get sample duration of event.
+
+        Returns
+        -------
+        int
+            duration in samples.
+        """
+        return self.offset - self.onset
```
SiQube, line 44: since time here is samples -- when do we plan to add ms/sec

SiQube (approved):

src/pymovements/events/base.py
dkrako: You are totally right that this will have to be extended somehow to increase usability.

I addressed this issue in #50

PR #51, opened 2022-10-30
Add position property to fixation class

jakobchwastek:
## Description

Fixation detection algorithms may output the position of a fixation. Therefore I added this property to Fixation event class.

## Implemented changes
- Add position property for fixation event class

## Type of change

### DIFF
--- a/src/pymovements/events/base.py
+++ b/src/pymovements/events/base.py
@@ -1,6 +1,7 @@
 """
 This module holds all the main Event classes used for event detection.
 """
+from __future__ import annotations
 
 
 class Event:
@@ -24,7 +25,6 @@ class Event:
         Starting index of event (included).
     offset: int
         Ending index of event (excluded).
-
     """
     def __init__(self, name: str, onset: int, offset: int):
         self.name = name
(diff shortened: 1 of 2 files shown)


src/pymovements/events/base.py
```diff
@@ -0,0 +1,92 @@
...
+    onset: int
+        Starting index of event (included).
+    offset: int
+        Ending index of event (excluded).
+    position: tuple[float]
+        (x, y) position of fixation
+    """
+    _name = 'fixation'
+
+    def __init__(self, onset: int, offset: int, position: tuple[float]):
```
dkrako, line 62: change `position` type hint to `tuple[float, float]`
right now this won't show as an error because we don't have mypy integrated yet.

but just  for clarification:
- `tuple[float]` refers to a tuple of length 1 with a single `float` like `(1, )`
- `tuple[float, float]` refers to a tuple of length 2 with both elements `float` like `(1, 1)`
- `tuple[float, str]` refers to a tuple of length 2 with a `float` and a `str` like `(1, 'abc')`
- `tuple[float, ...]` refers to a tuple of variable length like `(1, 1, 1, 1, 1)`

see for reference:
https://docs.python.org/3/library/typing.html#typing.Tuple

src/pymovements/events/base.py
```diff
@@ -0,0 +1,92 @@
...
+        super().__init__(name=self._name, onset=onset, offset=offset)
+
+    @property
```
dkrako, line 65: save as attribute in `__init__` and delete property

dkrako (changes requested):
some minor changes need to be addressed, but looks good otherwise

dkrako (approved):

src/pymovements/events/base.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    Fixation class.
+
+    Parameters
+    ----------
+    onset: int
+        Starting index of event (included).
+    offset: int
+        Ending index of event (excluded).
+    position: tuple[float]
```
dkrako, line 57: `tuple[float]` -> `tuple[float, float]`

dkrako (approved):

dkrako:
LGTM
Thank you for your effort

Issue #20, opened 2022-10-14
Integrate Identification by Velocity Threshold (I-VT)


dkrako:
## Description of the problem
The quite simple algorithm is defined here:
https://dl.acm.org/doi/10.1145/355017.355028
There exist several implementations of this on the internet where you can get inspiration.

## Description of a solution
- [x] implement basic algorithm
- [ ] create synthetic test data and test for expected result

## Minimum acceptance criteria
- [ ] passing tests with synthetic test data

jakobchwastek:
Missing:

- [ ] Add documentation for errors

PR #52, opened 2022-10-30
Add basic i-vt implementation

jakobchwastek:
## Description

Implementation of I-VT algorithm. This contributes to #20


src/pymovements/events/ivt.py
```diff
@@ -0,0 +1,86 @@
...
+from typing import Optional
+import numpy as np
+
+from pymovements.transforms import vnorm
+from pymovements.events import Fixation
+
+
+def ivt(
+        x: list[list[float]] | np.ndarray,
+        v: Optional[list[list[float]] | np.ndarray],
```
SiQube, line 12: since we already import annotations, you can rewrite this as None | list[........

src/pymovements/events/ivt.py
```diff
@@ -0,0 +1,86 @@
+from __future__ import annotations
+
+from typing import Optional
+import numpy as np
+
+from pymovements.transforms import vnorm
+from pymovements.events import Fixation
+
+
+def ivt(
```
SiQube, line 10: should we include `eye`as an argument or rather decide which eye to use via input? I'm unsure, maybe include @dkrako

src/pymovements/events/ivt.py
```diff
@@ -0,0 +1,86 @@
...
+    fixations: array
+        List of fixations
+
+    """
+    x = np.array(x)
+    v = np.array(v)
+
+    # Check if threshold is greater 0
+    if not t > 0:
+        raise ValueError(
```
SiQube, line 38: does not fit on one line?

src/pymovements/events/ivt.py
```diff
@@ -0,0 +1,86 @@
...
+            'velocity threshold must be greater than 0'
+        )
+
+    # make sure x and v have shape (n, 2)
+    if x.ndim != 2 and x.shape[1] != 2:
+        raise ValueError(
```
SiQube, line 44: does not fit on one line?

src/pymovements/events/ivt.py
```diff
@@ -0,0 +1,86 @@
...
+
+    if x.ndim != 2 and x.shape[1] != 2:
+        raise ValueError(
+            'v needs to have shape (n, 2)'
+        )
+
+    # Check matching shape for x and v
+    if not x.shape == v.shape:
+        raise ValueError(
+            "shape x {} doesn't match shape v {}".format(x.shape, v.shape)
```
SiQube, line 56: f-string it please :) we're python 3.7+

src/pymovements/events/ivt.py
```diff
@@ -0,0 +1,86 @@
...
+    Identification of fixations based on velocity-threshold
+
+    Parameters
+    ----------
+    x: array-like
+        Continuous 2D position time series.
+    v: array-like
+        Corresponding continuous 2D velocity time series.
+    t: float
+        Velocity threshold.
```
SiQube, line 25: Maybe add description what exactly t is?

src/pymovements/events/ivt.py
```diff
@@ -0,0 +1,86 @@
...
+from pymovements.events import Fixation
+
+
+def ivt(
+        x: list[list[float]] | np.ndarray,
+        v: Optional[list[list[float]] | np.ndarray],
+        t: float
+) -> list[Fixation]:
+    """
+    Identification of fixations based on velocity-threshold
```
SiQube, line 16: Please explain better how exactly ivt works, especially how `t` is used

src/pymovements/events/ivt.py
SiQube: velocity threshold in what unit? especially since the event(fixation) class has duration in samples

src/pymovements/events/ivt.py
```diff
@@ -0,0 +1,86 @@
...
+    # Check matching shape for x and v
+    if not x.shape == v.shape:
+        raise ValueError(
+            "shape x {} doesn't match shape v {}".format(x.shape, v.shape)
+        )
+
+    total_velocity = vnorm(v)
+
+    # Map velocities lower than threshold to 1 and greater equals to 0
+    fix_map = np.array(list(map(lambda point: 1 if point < t else 0, total_velocity)))
```
SiQube, line 62: I think this is a reasonable usage of list map lambda, but in a lot of cases list map lambda is not easily readable for users outside of the bubble/package, maybe if there is a neat way to reformulate it do it -- not a must

src/pymovements/events/ivt.py
```diff
@@ -0,0 +1,86 @@
...
+            'v needs to have shape (n, 2)'
+        )
+
+    # Check matching shape for x and v
+    if not x.shape == v.shape:
+        raise ValueError(
+            "shape x {} doesn't match shape v {}".format(x.shape, v.shape)
+        )
+
+    total_velocity = vnorm(v)
```
SiQube, line 59: is total_velocity the right name?

src/pymovements/events/ivt.py
SiQube: or maybe explicit mention that `1` means fixation; 0 saccade

src/pymovements/events/ivt.py
```diff
@@ -0,0 +1,86 @@
...
+
+    fixations = []
+
+    for onset, offset in fix_groups:
+        fixation_points = x[onset:offset]
+        centroid = np.sum(fixation_points, axis=0) / len(fixation_points)
+
+        fixations.append(Fixation(onset, offset, centroid))
+
+    return fixations
```
SiQube, line 86: do we only want to return fixations or also saccades? @dkrako

SiQube (commented):
moreover, tests are missing

src/pymovements/events/ivt.py
dkrako: Also add a citation to original paper.

For this to look best, do the following:
- Add a bib entry to https://github.com/aeye-lab/pymovements/blob/main/docs/source/bibliography.bib
- Copy the complete bibtex from https://dl.acm.org/doi/10.1145/355017.355028
- Use `Salvucci2000` as the bib-key.
- Create a citation in the docstring using :cite:t:`Salvucci2000`.

Here's a reference for using bibtex in sphinx:
https://sphinxcontrib-bibtex.readthedocs.io/en/latest/usage.html

src/pymovements/events/ivt.py
dkrako: do we need to specify which unit it actually is, as the unit is determined by the `velocity` argument?
so maybe just write that it's the same unit as for `velocity`

src/pymovements/events/ivt.py
```diff
@@ -0,0 +1,78 @@
...
+        raise ValueError('v needs to have shape (n, 2)')
+
+    # Check matching shape for x and v
+    if not x.shape == v.shape:
+        raise ValueError(f"shape x {x.shape} doesn't match shape v {v.shape}")
+
+    velocity_norm = vnorm(v)
+
+    # Map velocities lower than threshold to 1 and greater equals to 0
+    fix_map = np.array(list(map(lambda point: 1 if point < t else 0, velocity_norm)))
```
dkrako, line 54: as discussed, simplify this line

src/pymovements/events/ivt.py
dkrako: we will only return fixations for the current milestone.
i will create an issue for "filling up" with saccades / fixations, depending on the algorithm

src/pymovements/events/ivt.py
```diff
@@ -0,0 +1,78 @@
...
+
+    # Stack onsets and offsets and filter out fixation groups
+    groups = np.stack((ind_group_onsets, ind_group_offsets), axis=1)
+    fix_groups = groups[[fix_map[group[0]] == 1 for group in groups]]
+
+    fixations = []
+
+    for onset, offset in fix_groups:
+        fixation_points = x[onset:offset]
+        centroid = np.sum(fixation_points, axis=0) / len(fixation_points)
```
dkrako, line 74: reformulate using `np.mean`

src/pymovements/events/ivt.py
```diff
@@ -0,0 +1,78 @@
...
+        x: list[list[float]] | np.ndarray,
+        v: Optional[list[list[float]] | np.ndarray],
+        t: float
+) -> list[Fixation]:
+    """
+    Identification of fixations based on velocity-threshold
+
+    Parameters
+    ----------
+    x: array-like
```
dkrako, line 20: add shape information to parameters `(N, 2)`

src/pymovements/events/ivt.py
```diff
@@ -0,0 +1,78 @@
...
+    x = np.array(x)
+    v = np.array(v)
+
+    # Check if threshold is greater 0
+    if not t > 0:
+        raise ValueError('velocity threshold must be greater than 0')
+
+    # make sure x and v have shape (n, 2)
+    if x.ndim != 2 and x.shape[1] != 2:
+        raise ValueError('x needs to have shape (n, 2)')
```
dkrako, line 42: change shape to `(N, 2)`

src/pymovements/events/ivt.py
```diff
@@ -0,0 +1,78 @@
...
+    -------
+    fixations: array
+        List of fixations
+
+    """
+    x = np.array(x)
+    v = np.array(v)
+
+    # Check if threshold is greater 0
+    if not t > 0:
```
dkrako, line 37: this condition is simpler to parse for a human: `if t <= 0:`

dkrako (commented):
Thank you very much for your work!
There are some changes that need to be done, but all in all it looks very good.

dkrako (changes requested):
Also please do a `git rebase -i origin/main` and get rid of the commits that are already present in the main branch.
That way it's easier to see what this PR is actually changing.

src/pymovements/events/ivt.py
dkrako: we will stay monocular for the current milestone and add binocular functionality later on.

dkrako:
There is a new PR for a step function generator: #63 

Here is the code for the function: 
https://github.com/aeye-lab/pymovements/blob/feature/step_function/src/pymovements/synthetic/step_function.py

And some tests for example use:
https://github.com/aeye-lab/pymovements/blob/feature/step_function/tests/synthetic/test_step_function.py

To include these changes into your own branch, please run
```
git rebase -i origin/feature/step_function
```
An editor opens, then pick only the commits from your own branch and remove the other lines.

dkrako (approved):
LGTM!
Thanks a lot for this work!

PR #64, opened 2022-11-06
Parametrize tests for transformation functions

jakobchwastek:
## Description

Merged the single tests for `transforms.pos2vel` and `transforms.deg2vel` into one test by pytest parametrization:

## Implemented changes
- `test_pix2deg_raise_error` function for `pix2deg` raising error `expected_error` with input `kwargs`
- `test_pos2vel_raise_error` function for `pos2vel` raising error `expected_error` with input `kwargs`
- `test_pix2deg_equals` to check if `pix2deg` returns `expected_value` with input `kwargs` 
- `test_pos2vel_returns` to check if `pos2vel` returns `expected_value` with input `kwargs` and calculation method `method`

### DIFF
--- a/tests/test_transformations.py
+++ b/tests/test_transformations.py
@@ -13,6 +13,7 @@ screen_cm_1d = 100
 screen_px_2d = [100, 100]
 screen_cm_2d = [100, 100]
 
+
 @pytest.mark.parametrize(
     'kwargs, expected_error',
     [
@@ -24,7 +25,7 @@ screen_cm_2d = [100, 100]
                 'distance_cm': 1
             },
             TypeError,
-            id='none_coords_raise_type_error'
+            id='none_coords_raises_type_error'
         ),
         pytest.param(
             {
@@ -34,7 +35,7 @@ screen_cm_2d = [100, 100]
                 'distance_cm': 1
             },
             TypeError,
-            id='none_screen_px_raise_type_error'
+            id='none_screen_px_raises_type_error'
         ),
         pytest.param(
             {
@@ -44,7 +45,7 @@ screen_cm_2d = [100, 100]
                 'distance_cm': 1
             },
             TypeError,
-            id='none_screen_cm_raise_type_error'
+            id='none_screen_cm_raises_type_error'
         ),
         pytest.param(
             {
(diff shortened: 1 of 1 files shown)


jakobchwastek:
Should we set a similar name for return tests `test_pix2deg_equals` and `test_pos2vel_returns` ?

dkrako:
> Should we set a similar name for return tests `test_pix2deg_equals` and `test_pos2vel_returns` ?

Yes, good idea.
Let's name it `test_pix2deg_returns` and `test_pos2vel_returns` then.

Also rename the functions which test for exception raising:
`test_pix2deg_raise_error` to `test_pix2deg_raises_error` and the same for the one for `pos2vel`.

There are three small flake8 errors left.

Otherwise, this looks nice and can be merged after these small fixes.

tests/test_transformations.py
```diff
@@ -7,262 +7,437 @@
...
+            id='linear_input_returns_constant_velocity'
+        )
+    ]
+)
+def test_pos2vel_returns(method, kwargs, padding, expected_value):
+    actual_value = pos2vel(**kwargs, method=method)
+    assert (actual_value[padding[0]:padding[1]] == expected_value[padding[0]:padding[1]]).all()
 
 
 def test_pos2vel_stepped_input_returns_alternating_velocity_method_preceding():
```
dkrako, line 443: one last thing:
can you parametrize these three functions too?

just call it `test_pos2vel_stepped_input_returns`

leave the creation of `N` and `x` in the function and have params like this:

```
@pytest.mark.parametrize(
    'method, expected',
    [
        pytest.param(
            'preceding', np.array([2.0, 0.0] * (N // 2)),
            id='method_preceding_alternating_velocity',
        ),
```

dkrako (changes requested):
the refactoring of the last three methods is remaining

tests/test_transformations.py
SiQube: @jakobchwastek @dkrako is this resolved and mergeable?

dkrako: Although this is not resolved yet, I decided to merge this PR.
I will create another issue for this remaining task.

dkrako (approved):
Although this is not resolved yet, I decided to merge this PR.
I will create another issue for this remaining task.

dkrako:
Ah sorry, I see this was already fixed in november and I just didn't get it.
Must have been blind or so.

Issue #41, opened 2022-10-19
integrate mypy into github workflow


dkrako:
## Description of the problem
We have typehints so we can use them

## Description of a solution
integrate mypy into github workflow

dkrako:
The following steps are necessary:

- [ ] add mypy to tox environments: https://github.com/aeye-lab/pymovements/blob/main/tox.ini
- [ ] add mypy to gitworkflow checks: https://github.com/aeye-lab/pymovements/blob/ad9c0174b22aa98a0048063622d2c45ca313bc07/.github/workflows/tests.yml#L50
- [ ] fix all the warnings and errors that come from adding mypy to the CI

SiQube:
we could solve this using pre-commit, see #67

dkrako:
I think actually it's the other way around.
We need mypy in a transparent way in the github workflow, otherwise reviewers have to run it themselves or trust that the author didn't use `--no-verify`.

SiQube:
no -- actually pre-commit-ci can do it for us :-) runtime way faster

dkrako:
I forgot to add the hard requirement for this issue, that this needs to be integrated into the tox pipeline.
We use the tox environment to be able to run all tests (including all supported python versions) locally.
By simply calling `tox` you can automatically run all necessary tests on your local machine (this will be including mypy).

So for consistency, all definitions are required to be in the `tox.ini` file.
Otherwise, developers are needed to run two different commands to check if all tests are passing for the current working state.
To run all `mypy` checks you would then just call `tox -e mypy`.

If you want pymovements to switch to another testing environment, please create a new issue for this, as it is out of scope with this specific issue.
I'm open to suggestions and that will be a better place to discuss further about what alternative solutions there are.
If we decide to switch, we can then move all the definitions from the `tox.ini` to a new system.

dkrako:
done in #68

SiQube:
as seen in PR #198 we run mypy twice, should we remove mypy from tox? pre-commit can also run locally

dkrako:
As I have already encountered a bug where pre-commit.ci didn't run correctly I would be against it

PR #68, opened 2022-11-09
add mypy to tox.ini

SiQube:
## Description

integrate mypy into tox environment -- fixes issue #41 

## Implemented changes
- Insert a description of the changes implemented in the pull request.
  - [x] mypy.ini (exclude missing imports, `pymovements/plot.py`)
  - [x] tox.ini -- call mypy tests
  - [x] fix mypy error in `transform.py`

## Type of change

Please delete options that are not relevant.

- [x] new python code check

## How Has This Been Tested?

- [x] it's a mypy -- a test

## Checklist:


tox.ini
```diff
@@ -66,6 +66,14 @@ commands =
     pylint --rcfile=pylintrc --output-format=parseable {toxinidir}/src/pymovements {toxinidir}/tests
 
 
+[testenv:mypy]
+deps =
+    mypy
```
dkrako, line 71: I think when you add this line with a single dot, this will install the package dependencies and we're not forced to ignore the missing imports.

```suggestion
    mypy
    .
```

Also see this dependency definition for running pytest as an example with additional extras:
https://github.com/aeye-lab/pymovements/blob/07f13f60b8b0669ccb5eeda903b0b48d9a93843d/tox.ini#L7-L10

In theory we don't need to specify mypy as a dependency this way in the `setup.cfg`. Then it will be only installed for the venv which is used for running mypy checking.
Although I think it would be best to include it here:
https://github.com/aeye-lab/pymovements/blob/07f13f60b8b0669ccb5eeda903b0b48d9a93843d/setup.cfg#L32-L36

It would be very nice if you could also put `flake8` and `pylint` together with `mypy` into the `setup.cfg`, as they seem to be missing from the test-extra dependency.

Also, while you're at it, please remove `pytest-cov` from the definition here, 
https://github.com/aeye-lab/pymovements/blob/07f13f60b8b0669ccb5eeda903b0b48d9a93843d/tox.ini#L7-L10
it is redundant as it's already in the `setup.cfg`.

I think then specification of test dependencies would be a bit more consistent than in the current state.
May be a bit out of the original scope of the PR, though, but the changes are so tiny :man_shrugging:

mypy.ini
```diff
@@ -0,0 +1,14 @@
+[mypy]
+exclude = src/pymovements/plot.py
+
+[mypy-matplotlib.*]
```
dkrako, line 4: I think we can get rid of these ignores.
See the comment for `tox.ini`

src/pymovements/transforms.py
```diff
@@ -213,7 +214,7 @@ def pos2vel(
     return v
 
 
-def vnorm(arr: np.ndarray, axis: int | None = None) -> np.ndarray:
+def vnorm(arr: np.ndarray, axis: int | None = None) -> np.ndarray | Any:
```
dkrako, line 217: is it Any because numpy doesn't fully support typehinting yet?

tox.ini
```diff
@@ -66,6 +66,14 @@ commands =
     pylint --rcfile=pylintrc --output-format=parseable {toxinidir}/src/pymovements {toxinidir}/tests
 
 
+[testenv:mypy]
+deps =
```
dkrako, line 70: please add these two lines
```suggestion
basepython = python3.8
changedir = {toxinidir}
deps =
```

maybe it's also a good idea to use `python3.9` for all `basepython` definitions.

dkrako (changes requested):
Thank you very much for the effort!
I think at least the stuff with the imports should be fixed.

Stuff for refactoring the `tox.ini` can be also done in a separate issue/PR with a different assignee, just as you like.

tox.ini
```diff
@@ -66,6 +66,14 @@ commands =
...
+    mypy
+commands =
+    mypy src
```
dkrako, line 73: use a more safe style by using `{toxinidir}`to be sure to use the correct tox environment 
including `{posargs}` will let you pass additional positional arguments to mypy.

example command defintion for flake8:

https://github.com/aeye-lab/pymovements/blob/07f13f60b8b0669ccb5eeda903b0b48d9a93843d/tox.ini#L55-L56

src/pymovements/transforms.py
SiQube: yes -- they did not type linalg (output) and hence it is Any or float....

mypy.ini
SiQube: unfortunately we can't get rid of scipy ignore -- scipy does not type...

SiQube (commented):
changed everything according to your feedback =)

dkrako:
LGTM thanks for your work!

PR #69, opened 2022-11-09
remove type from tests/events/test_events_base.py to conform with rest of test suite

SiQube:
## Description
I removed type hints from one test file to match the rest of the tests -- generally I think we do not need to type tests because we're checking type hints with type checkers and tests are for functionality. functionallity should fail given the wrong type.

## Implemented changes
- Insert a description of the changes implemented in the pull request.
  - [x] removed typing hint from tests/events/test_events_base.py 

## Checklist:

### DIFF
--- a/tests/events/test_events_base.py
+++ b/tests/events/test_events_base.py
@@ -1,9 +1,6 @@
 """
 This module tests functionality of base event classes.
 """
-from __future__ import annotations
-from typing import Any
-
 import pytest
 
 from pymovements.events import Event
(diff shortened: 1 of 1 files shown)


dkrako:
yes you're totally right, this is a good idea.

tests/events/test_events_base.py
```diff
@@ -2,7 +2,6 @@
 This module tests functionality of base event classes.
 """
 from __future__ import annotations
```
dkrako, line 4: we can also remove this import then

dkrako (changes requested):
Thank you very much for your effort!

tests/events/test_events_base.py
SiQube: did it in newest commit

dkrako (commented):
LGTM

Issue #83, opened 2023-01-25
Add public datasets


dkrako:
## Description of the problem
Referencing issue #58 

A lot of data loading stuff has to be dragged through all of our projects.
It would be nice to have a common interface for this.

## Description of a solution
We would create a pymovements.datasets module where publicly available eye tracking datasets could me downloaded and read into memory automatically.

This way we can foster the visibility in two ways:

- by including a lot of datasets in the module, eye tracking researchers are drawn to the pymovements package as they get all this integration for free
- by including all the datasets from our group, pymovement users can instantly integrate these into their own research

## Minimum acceptance criteria
- [x] dataset needs to be automatically downloaded and extracted
- [x] datasets need to be loaded into memory

## Context

I was inspirired by torchvision which is doing a similar thing with its visiondatasets module:
https://pytorch.org/vision/stable/datasets.html

torchvision also supports transformations, we can add this functionality later in a subsequent issue

PR #91, opened 2023-01-27
Datasets: Add JuDo1000 dataset

dkrako:
## Description

This PR adds the public JuDo1000 dataset to our package.
This way it is very easy to download, extract, read and process this and hopefully upcoming datasets.

Fixes issue #83 

## Implemented changes

- [x] Add datasets.Dataset
- [x] Add datasets.PublicDataset
- [x] Add datasets.JuDo1000
- [x] Add datasets.Dataset.read()
- [x] Add datasets.Dataset.compute_dva()
- [x] Add docstrings to all methods.

## Type of change

## How Has This Been Tested?

I have written a basic test which downloads the dataset with a plan to do some more testing but switched to a jupyter notebook midway.

The main reasons for this is that the test directory should be reserved for unit tests only.
Downloading a dataset and then checking for stuff is actually more like an integration test / functional tests.

We will have to dig deeper how we implement our functional tests, which are tests that span acomplete use case / user story.
There's an overlap to the tutorials section, where we will present the main use cases and how to get stuff done with pymovements.
Maybe there will be a way to integrate this too? I don't know and we will have to see.

We further need to evaluate if it's really a sensible approach to download the whole dataset for each call of the test suite.
We will generate a lot of traffic this way which is somehow wasteful. I don't now about a good alternative approach though. The same issue will actually also arise for adding example code.

I'm not quite sure but I think pytorch is using some internal magic to mitigate this issue:
https://github.com/pytorch/vision/blob/5dd95944c609ac399743fa843ddb7b83780512b3/torchvision/_internally_replaced_utils.py

## Checklist:


dkrako:
Finally ready for review.

Support for monocular datasets will be added with the upcoming GazeBase PR.

setup.cfg
```diff
@@ -38,6 +38,9 @@ install_requires =
     numpy>=1.10.0
     scipy>=1.5.4
     matplotlib>=3.0.0
+    pandas>1.0.0
+    polars>=0.15.0
```
SiQube, line 42: do we need pandas and polars? I think the less dependencies the better

src/pymovements/datasets/base.py
```diff
@@ -0,0 +1,375 @@
...
+        ------
+        AttributeError
+            If no regular expression for parsing filenames is defined.
+        RuntimeError
+            If an error occured during matching filenames.
+        """
+        if self._filename_regex is not None:
+            filename_regex = re.compile(self._filename_regex)
+        else:
+            raise AttributeError("no regular expression for filenames is defined.")
```
SiQube, line 91: this could also be
```python
assert self._filename_regex is not None, "no regular expression for filenames is defined."
filename_regex = re.compil(self._filename_regex)
```

don't know if it might be faster

src/pymovements/datasets/base.py
```diff
@@ -0,0 +1,375 @@
...
+        for filepath in csv_filepaths:
+
+            # All csv_filepaths already match the filename_regex.
+            match = filename_regex.match(filepath.name)
+
+            # This actually should never happen but mypy will complain otherwise.
+            if match is None:
+                raise RuntimeError(
+                    f"file {filepath} did not match regular expression {self._filename_regex}",
+                )
```
SiQube, line 110: could also be an assert

src/pymovements/datasets/base.py
```diff
@@ -0,0 +1,375 @@
...
+            If `fileinfo` is None or the `fileinfo` dataframe is empty.
+        """
+        file_dfs: list[pl.DataFrame] = []
+
+        if self.gaze is None:
+            raise AttributeError(
+                "fileinfo was not read yet. please run read() or read_gaze_files() beforehand",
+            )
+        if len(self.gaze) == 0:
+            raise AttributeError("no files present in gaze attribute")
```
SiQube, line 147: could also be assert

src/pymovements/datasets/base.py
```diff
@@ -0,0 +1,375 @@
...
+            if experiment is None.
+        """
+        if self.gaze is None:
+            raise AttributeError(
+                "gaze files were not read yet. please run read() or read_gaze_files() beforehand",
+            )
+        if len(self.gaze) == 0:
+            raise AttributeError("no files present in gaze attribute")
+        if self.experiment is None:
+            raise AttributeError('experiment must be specified for this method.')
```
SiQube, line 194: could be asserts

src/pymovements/datasets/base.py
```diff
@@ -0,0 +1,375 @@
...
```
SiQube, line 240: could be assert

SiQube (commented):
additionally to my comments above we also have a merge conflict

setup.cfg
dkrako: for pandas export both pandas and pyarrow are required.
so a potential user who wants to export the polars dataframe to pandas would be confused about the error of missing pyarrow and would need to install it manually.

you're right, the less dependencies the better, but right now I tend to think that most users who work with dataframes in python would want to use pandas and then they get this missing module error.

I also would like to add this export function to a cell in a tutorial notebook to show this functionality to the polars-uninitiated. so at least for the `docs` extras pandas would be required.

As discussed in slack, let's keep it there and see if it becomes an issue in the future.

src/pymovements/base.py
```diff
@@ -178,5 +190,40 @@ def __init__(
...
             height_cm=screen_height_cm,
             distance_cm=distance_cm,
+            origin=origin,
         )
         self.sampling_rate = sampling_rate
+
+    def pos2vel(
+        self,
+        arr: list[float] | list[list[float]] | np.ndarray,
+        method: str = 'smooth',
```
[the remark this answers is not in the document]
dkrako: yes you're totally right and i thought about this also.
I read a much better one-word description in a paper for this method, but naturally i didn't write it down and forgot where it was.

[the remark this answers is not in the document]
dkrako: I'll create an issue for this.

src/pymovements/datasets/base.py
dkrako: Asserts shouldn't be used in production code as these can get optimized away.

The problem is, that assertions are deactivated when you use the optimization parameter `-O` on launching python. So as an example if a user wants to optimize the interpreter/compiler for memory-efficiency, all of the assertions in this package will be ignored.
Most people don't use this optimization parameter, but I think we should be aware of this issue and maybe even enforce this for newly written code.

Here are some discussions on stackoverflow which can be helpful:
https://stackoverflow.com/questions/944592/best-practice-for-using-assert
https://stackoverflow.com/questions/28608385/assert-asserting-when-debug-false
https://stackoverflow.com/questions/56990925/is-this-assert-for-development-not-for-production


TLDR:

* use exceptions for the user-side or external system-side of the package, where things can go wrong by unintended input/output, or maybe errors in servers, sockets or filesystem (like an OSError will be raised when you try to download but there's no space left on your device)
* use assertions for the developer-side of the package, to assert conditions that should absolutely never go wrong, independent of user input. in production this is more like checking your loop-invariants because you're paranoid. otherwise, asserts should be exclusively used in testing code.

src/pymovements/base.py
```diff
@@ -66,10 +76,11 @@ def __init__(
         ...     width_cm=38,
         ...     height_cm=30,
         ...     distance_cm=68,
+        ...     origin='lower left',
         ... )
         >>> print(screen)  # doctest: +NORMALIZE_WHITESPACE
         Screen(width_px=1280, height_px=1024, width_cm=38, height_cm=30, distance_cm=68,
-         x_max_dva=15.60, y_max_dva=12.43, x_min_dva=-15.60, y_min_dva=-12.43)
+        origin=lower left, x_max_dva=15.60, y_max_dva=12.43, x_min_dva=-15.60, y_min_dva=-12.43)
```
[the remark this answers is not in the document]
dkrako: i struggled with this too, but actually this will fail with the current `auto_str` implementation

SiQube (approved):

PR #104, opened 2023-02-03
Add GazeBase

dkrako:
## Description

This PR adds the GazeBase dataset.

PRs #91 #103 #102 and #105 need to be merged first.

## Implemented changes

- [x] Add class GazeBase to datasets
- [ ] Add bibliography item to docs

## Type of change

## How Has This Been Tested?

No testing done for the same reasons as for PR #91

## Checklist:

### DIFF
--- a/src/pymovements/datasets/gazebase.py
+++ b/src/pymovements/datasets/gazebase.py
@@ -15,7 +15,7 @@ class GazeBase(PublicDataset):
 
     In each of the two sessions per round, participants are instructed to complete a series of
     tasks, including a fixation task (FIX), a horizontal saccade task (HSS), a random saccade task
-    (RAN), a reading task (TEX), two free viewing video task (VD1 and VD2) and a gaze-driven gaming
+    (RAN), a reading task (TEX), two free viewing video tasks (VD1 and VD2) and a gaze-driven gaming
     task (BLG).
 
     Check the respective paper for details :cite:p:`GazeBase`.
--- a/src/pymovements/datasets/gazebase.py
+++ b/src/pymovements/datasets/gazebase.py
@@ -21,13 +21,12 @@ class GazeBase(PublicDataset):
         },
     ]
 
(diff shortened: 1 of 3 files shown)


src/pymovements/datasets/gazebase.py
```diff
@@ -0,0 +1,66 @@
+"""This module provides an interface to the GazeBase dataset."""
+from pymovements.base import Experiment
+from pymovements.datasets.base import PublicDataset
+
+
+class GazeBase(PublicDataset):
+    """GazeBase dataset :cite:p:`GazeBase`.
+
+    Check the respective paper for details.
+    """
```
SiQube, line 10: I thought about this quite a bit when I went over the JuDo dataclass. I think adding a real description here is tough -- maybe we should just add it like here and link the paper / osf / wherever the dataset has any details.

SiQube: maybe even more important would be the unit in which the samples are returned

dkrako: I think a brief description of the datasets is necessary to be a useful feature.

From the JuDo1000 implementation I would say this information would be useful:

- number of users (in case of gazebase I would simply add that not all of them are present across all rounds)
- number of sessions (in case of gazebase the distinction between rounds and sessions need to made clear otherwise the user will be very confused)
- sampling rate
- which eyes are recorded
- in which units is the gaze provided
- which stimuli does the participant see. in case of gazebase a very very brief description (one word) of the stimuli would be useful.

Moreover, all this information would be necessary for a feature I want to add in the future, where the user can specify to read in only a specific user, session, round or stimulus.

dkrako: But right now nothing of this is present in the GazeBase docstring, I'll have to add this and then we can see if it's too verbose.

SiQube: I agree we should add a small docstring documentation of the code. I think your list above seems reasonable

SiQube (approved):

PR #107, opened 2023-02-03
Transforms: refactor center_origin to origin in pix2deg()

dkrako:
This PR changes some functionality from `pix2deg()`.

Before, `center_origin` was used to center pixel coordinates from the lower left to the center.
Now we have an `origin` argument to specify the location of the origin.
The meaning of the argument is much more understandable this way.
This way we will also be more flexible in the future.

Currently we still only support two cases:

- `center` which is the same as `center_origin=False` 
- `lower left` which is the same as `center_origin=True`

### DIFF
--- a/src/pymovements/base.py
+++ b/src/pymovements/base.py
@@ -65,7 +65,7 @@ class Screen:
         distance_cm : float
             Eye-to-screen distance in centimeters
         origin : str
-            Specifies the origin location of pixel coordinates.
+            Specifies the screen location of the origin in the pixel coordinate system.
 
         Examples
         --------
@@ -178,7 +178,7 @@ class Experiment:
         distance_cm : float
             Eye-to-screen distance in centimeters
         origin : str
-            Specifies the origin location of pixel coordinates.
+            Specifies the screen location of the origin in the pixel coordinate system.
         sampling_rate : float
             Sampling rate in Hz
 
--- a/src/pymovements/base.py
+++ b/src/pymovements/base.py
@@ -29,7 +29,7 @@ class Screen:
     distance_cm : float
         Eye-to-screen distance in centimeters
     origin : str
-        Specifies the origin location of pixel coordinates.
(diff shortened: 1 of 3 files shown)


src/pymovements/base.py
```diff
@@ -29,6 +28,8 @@ class Screen:
         Screen height in centimeters
     distance_cm : float
         Eye-to-screen distance in centimeters
+    origin : str
+        Specifies the origin location of pixel coordinates.
```
SiQube, line 32: Specifies the origin location in pixel coordinates.

src/pymovements/transforms.py
```diff
@@ -91,8 +97,10 @@ def pix2deg(
     distance_px = distance_cm * (screen_px / screen_cm)
 
     # center screen coordinates such that 0 is in the center of the screen
```
SiQube, line 99: I think we should delete this comment or adjust it.

SiQube (commented):
I don't know if anyone would ever use this but we could additionally add issues for the other natural origins.

src/pymovements/base.py
dkrako: I see the issue in my formulation.. but I would interpret the origin location IN pixel coordinates to be `(0, 0)`, because technically, this is the origin specified in pixel coordinates.

What do you think about that?

``` 
Specifies the screen location of the origin in the pixel coordinate system.
```

src/pymovements/transforms.py
dkrako: yes, that leftover doesn't make any sense. I'll adjust it.

src/pymovements/base.py
SiQube: sounds good to me

dkrako: After some googling I think it's the origin of a space/coordinate system instead of in.
So I used that.

SiQube (approved):

PR #124, opened 2023-02-13
CI: Use py37+ for pupgrade

dkrako:


dkrako:
already did this accidentaly in a rebase in #121 
i'll close this then

PR #133, opened 2023-02-14
CI: Move mypy definitions to pyproject.toml

dkrako:
Currently our root directory is a bit cluttered with configuration files.

I made a start in resolving this issue by moving the few definitions in the `mypy.ini` to the `pyproject.toml`.

The definitions from `setup.cfg`, `tox.ini` and `pylintrc` will be migrated in some future sprints.
pre-commit and readthedocs unfortunately are stubborn and not willing to support `pyproject.toml`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,6 +5,14 @@ requires = [
 ]
 build-backend = "setuptools.build_meta"
 
+[tool.autopep8]
+max_line_length = 100
+in-place = true
+recursive = true
+aggressive = 1
+
+[tool.doc8]
+max-line-length = 100
 
 [tool.mypy]
 exclude = ['src/pymovements/plot.py']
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -5,13 +5,18 @@ requires = [
...
 in-place = true
 recursive = true
 aggressive = 1
 
-
 [tool.doc8]
 max-line-length = 100
+
+[tool.mypy]
+exclude = ['src/pymovements/plot.py']
```
SiQube, line 18: need to get rid of this asap

SiQube (approved):

pyproject.toml
dkrako: yes I'll create an issue in the future

PR #134, opened 2023-02-14
CI: Add pre-commit to GitHub workflows

dkrako:
I added pre-commit to the GitHub workflows.
I'm not sure if it makes sense in addition to the pre-commit.ci?

Also, this way flake8 and mypy will be run two times.
Maybe exclude it here by using `stages`? This could also be a reasonable setup for the pre-commit ci.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -68,6 +68,7 @@ jobs:
           - flake8
           - pylint
           - mypy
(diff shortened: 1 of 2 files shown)


dkrako:
Ok, I see pre-commit is added as a separate workflow already. So we won't need this in the workflow defintions now.

PR #135, opened 2023-02-14
CI: Add configuration for pre-commit-ci

dkrako:
regarding the issue stated in PR #134 

pre-commit-ci now won't run flake8 and mypy as this is already handled in a separate job by GitHub workflows.


dkrako:
we decided to run both in parallel

PR #140, opened 2023-02-14
Transforms: Fix padding bug in cut_into_subsequences

dkrako:
Thanks to the examples in #95 from @theDebbister we could find a bug for the padding in `cut_into_subsequences()`.

Fortunately I never used `keep_padded=True`, but I have added the respective tests to prevent regressions.

### DIFF
--- a/src/pymovements/transforms.py
+++ b/src/pymovements/transforms.py
@@ -285,22 +285,22 @@ def cut_into_subsequences(
     idx = 0
     for t in range(0, arr.shape[0]):
         for i in range(0, n):
-            # get piece of respective sequence
+            # get part of the sequence with window_size
             arr_tmp = np.expand_dims(arr[t, i * window_size: (i + 1) * window_size, :], axis=0)
 
-            # fill respective instance with piece
+            # fill row with respective part of the sequence
             arr_cut[idx, :, :] = arr_tmp
 
             idx = idx + 1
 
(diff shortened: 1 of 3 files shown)


src/pymovements/transforms.py
```diff
@@ -287,30 +285,23 @@ def cut_into_subsequences(
     idx = 0
     for t in range(0, arr.shape[0]):
         for i in range(0, n):
-            # cut out 1000 ms piece of trial t
+            # get piece of respective sequence
```
SiQube, line 288: if we have a comment here I think this should be more specific. e. g. get piece of sequence with shape (window_size, ...)

src/pymovements/transforms.py
```diff
@@ -287,30 +285,23 @@ def cut_into_subsequences(
...
             arr_tmp = np.expand_dims(arr[t, i * window_size: (i + 1) * window_size, :], axis=0)
 
-            # concatenate pieces
+            # fill respective instance with piece
```
SiQube, line 291: maybe part of the sequence?

src/pymovements/transforms.py
SiQube: or part of the sequence / subsequence

dkrako:
The function itself is implemented very ineffienctly, so I have created an issue for this #144

SiQube (approved):

Issue #65, opened 2022-11-08
use events DataFrame instead of list of events


dkrako:
## Description of the problem
Currently the event detection algorithms return a list of `Event`  objects.
For a usable processing of this result, an event dataframe should be constructed. 

## Description of a solution
Usually pandas is used for working with dataframes, but polars is a much more promising candidate:
https://www.analyticsvidhya.com/blog/2021/06/polars-the-fastest-dataframe-library-youve-never-heard-of/

That's why I'm much in favor of using polars instead of pandas for pymovements.
Unfortunately no one of us is experienced in polars yet, so maybe it makes sense to use pandas as an intermediate step.

## Minimum acceptance criteria
We should return a DataFrame with event data in all event detection methods.
If it is a pandas or polars dataframe doesn't matter for this first iteration.

PR #141, opened 2023-02-15
Events: Return dataframe in detection methods.

dkrako:
## Description
In the version before a list of Event objects was returned, which would be very inefficient to process.
I did this originally because I had to focus on other things at it seemed to be the most trivial way.

Now as I got a bit more experience with polars I changed this and we now have nice polars DataFrames as an output.
What I'm not sure about, should we now just delete these Event classes?

Fixes issue #65

## Implemented changes

- [x] Return DataFrame in `events.engbert.microsaccades()`
- [x] Return DataFrame in `events.idt.idt()`
- [x] Return DataFrame in `events.ivt.ivt()`
- [x] Refactor `events.idt.idt()` such that `transforms.consecutive()` is being used

## Type of change

## How Has This Been Tested?

I have tested this by using the step function.

- [x] IDT/IVT: 
  - [x] constant position sequence -> single fixation
  - [x] three steps (one short in the middle) -> two fixations
- [x] microsaccades:
  - [x] two steps -> one saccade
  - [x] four steps -> two saccades

## Checklist:

### DIFF
--- a/src/pymovements/events/engbert.py
+++ b/src/pymovements/events/engbert.py
@@ -6,8 +6,8 @@ from __future__ import annotations
 from collections.abc import Sized
 
 import numpy as np
+import polars as pl
 
-from pymovements.events import Saccade
 from pymovements.transforms import consecutive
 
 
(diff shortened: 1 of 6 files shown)


dkrako:
One more thing: I have identified a bug in IDT and added an issue for this: #142
The current tests are passing because I use the wrong values as expected values.
The correct ones are in the comment after and need to be fixed in the respective issue

SiQube (approved):
missing examples -- need issue for it

Issue #28, opened 2022-10-14
Create binocular event detection filter


dkrako:
## Description of the problem
The implemented eye tracking algorithms usually use a single eye for eye movement detection.
However, some eye movement events happen on both eyes simultaneously, so we need a method that filters out eye movements that occur only on one eye.

## Description of a solution
- [ ] implement binocular event filter
- [ ] create test cases and make them pass

## Minimum acceptance criteria
- [ ] a binocular event filter with passing tests

dkrako:
research algorithms if there are any

PR #152, opened 2023-02-16
Events: Implement common event detection interface

dkrako:
All event detection methods now have both positions and velocities parameters, independent of their actual usage.

This change also includes adding a new check that can be used across modules.

This change is the prerequisite for two things:
* a more straight-forward implementation in the coming `detect_events()` method in the `Dataset` class
* a greater range of properties to be included in the event dataframe

### DIFF
--- a/src/pymovements/events/engbert.py
+++ b/src/pymovements/events/engbert.py
@@ -28,10 +28,12 @@ import numpy as np
 import polars as pl
 
 from pymovements.transforms import consecutive
+from pymovements.utils.checks import check_shapes_positions_velocities
 
 
 def microsaccades(
-    velocities: np.ndarray,
+    positions: list[list[float]] | np.ndarray,
+    velocities: list[list[float]] | np.ndarray,
     threshold: np.ndarray | tuple[float] | str = 'engbert2015',
     threshold_factor: float = 6,
     minimum_duration: int = 6,
@@ -48,6 +50,8 @@ def microsaccades(
 
     Parameters
     ----------
+    positions : np.ndarray, shape (N, 2)
+        x and y positions of N samples in chronological order
     velocities : np.ndarray, shape (N, 2)
         x and y velocities of N samples in chronological order
     threshold : np.ndarray, tuple[float, float] or str
@@ -73,6 +77,11 @@ def microsaccades(
         If `threshold` value is below `min_threshold` value.
         If passed `threshold` is either not two-dimensional or not a supported method.
     """
+    positions = np.array(positions)
+    velocities = np.array(velocities)
+
+    check_shapes_positions_velocities(positions=positions, velocities=velocities)
+
     if isinstance(threshold, str):
         threshold = compute_threshold(velocities, method=threshold)
     else:
--- a/src/pymovements/events/engbert.py
+++ b/src/pymovements/events/engbert.py
@@ -32,8 +32,8 @@ from pymovements.utils.checks import check_shapes_positions_velocities
 
 
 def microsaccades(
-    positions: list[list[float]] | np.ndarray,
-    velocities: list[list[float]] | np.ndarray,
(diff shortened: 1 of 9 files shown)


src/pymovements/events/engbert.py
```diff
@@ -73,6 +77,12 @@ def microsaccades(
         If `threshold` value is below `min_threshold` value.
         If passed `threshold` is either not two-dimensional or not a supported method.
     """
+    positions = np.array(positions)
+    velocities = np.array(velocities)
+
+    # Check correct shapes of positions and velocities.
```
SiQube, line 83: I don't think this comment is necessary

src/pymovements/events/idt.py
```diff
@@ -84,24 +89,19 @@ def idt(
...
+    velocities = np.array(velocities)
 
-    # make sure positions is 2d
-    if positions.ndim != 2:
-        raise ValueError('positions needs to have shape (N, 2)')
-
-    # make sure positions has shape (N, 2)
-    if positions.shape[1] != 2:
-        raise ValueError('positions needs to have shape (N, 2)')
+    # Check correct shapes of positions and velocities.
```
SiQube, line 95: I think this comment is unnecessary

src/pymovements/events/idt.py
```diff
@@ -84,24 +89,19 @@ def idt(
...
+    check_shapes_positions_velocities(positions=positions, velocities=velocities)
 
     # Check if dispersion_threshold is greater 0
     if dispersion_threshold <= 0:
         raise ValueError('dispersion threshold must be greater than 0')
 
-    # Check if duration_threshold is greater 0
+    # Check if minimum_duration is greater 0
```
SiQube, line 102: unnecessary comment

src/pymovements/events/ivt.py
```diff
@@ -74,29 +75,12 @@ def ivt(
...
-    if velocities.shape[1] != 2:
-        raise ValueError('velocities need to have shape (N, 2)')
-
-    # Check matching shape for positions and velocities
-    if positions.shape != velocities.shape:
-        raise ValueError(
-            f"shape of positions {positions.shape} doesn't match"
-            f'shape of velocities {velocities.shape}',
-        )
+    # Check correct shapes of positions and velocities.
```
SiQube, line 78: delete?

src/pymovements/events/engbert.py
```diff
@@ -28,10 +28,12 @@
 import polars as pl
 
 from pymovements.transforms import consecutive
+from pymovements.utils.checks import check_shapes_positions_velocities
 
 
 def microsaccades(
-    velocities: np.ndarray,
+    positions: list[list[float]] | np.ndarray,
+    velocities: list[list[float]] | np.ndarray,
```
SiQube, line 36: should we here do list[list[float]] or list[tuple[float]]?

src/pymovements/events/engbert.py
```diff
@@ -48,6 +50,8 @@ def microsaccades(
 
     Parameters
     ----------
+    positions : np.ndarray, shape (N, 2)
+        x and y positions of N samples in chronological order
```
SiQube, line 54: so we only target one eye?

SiQube (approved):

src/pymovements/events/engbert.py
dkrako: I think we had this discussion already a few times, but lets keep it in #28 in the future, where we can decide on how to support binocular event detection one day.

just a basic summary of our past discussion outcomes:
* all our implemented event detection algorithms (and also the ones which we plan to implement) are only defined for a single eye
* we want to implement this functionality in a binocular filter in the preprocessing pipeline #28
* there we can also support different kinds of methods for merging the two eyes
* until then, the event detection will be purely monocular

PR #173, opened 2023-02-20
Remove velocities argument from IDT function

jakobchwastek:
## Description

Removed `velocities` argument from idt function since the velocities are not nesessary for the idt algorithm.

## Implemented changes

- [x] Remove `velocities` from function header and all code lines checking the value of this argument
- [ ] Remove the velocity calculation in the idt test functions

## Type of change 

- 
## Checklist:

### DIFF
--- a/src/pymovements/events/idt.py
+++ b/src/pymovements/events/idt.py
@@ -25,8 +25,6 @@ from __future__ import annotations
 import numpy as np
 import polars as pl
 
-from pymovements.utils.checks import check_shapes_positions_velocities
-
 
 def dispersion(positions: list[list[float]] | np.ndarray) -> float:
     """
@@ -50,7 +48,6 @@ def dispersion(positions: list[list[float]] | np.ndarray) -> float:
 
 def idt(
         positions: list[list[float]] | list[tuple[float, float]] | np.ndarray,
-        velocities: list[list[float]] | list[tuple[float, float]] | np.ndarray,
         dispersion_threshold: float,
         minimum_duration: int,
 ) -> pl.DataFrame:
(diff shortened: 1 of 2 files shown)


dkrako:
This was implemented in #152 and is needed for #159 to work properly.

I know this solution was not 100% optimal but the easiest one to implement for now.

Please create an issue to discuss how we can do this in the future.

dkrako:
Or maybe it's better that I reopen this again, we can have the discussion here too.

jakobchwastek:
Ahh okay, thanks for the update. I missed this. I was just starting to do an example for idt and tought I initialy just copy/pasted the function header from ivt with the additional argument.

dkrako:
No problem, it is actually a good idea to improve on this, and I already feared a bit that this will create usability issues.

But to fix the issue, we would now also need a more clever way of passing the arguments in this line:

https://github.com/aeye-lab/pymovements/blob/c833d5a74b6b3ee9e050db7619d6507cfecf2015/src/pymovements/datasets/dataset.py#L430

You could try to check the actual expected keyword arguments of the passed function via `inspect.getargspec()`

Moreover, I have defined an `EventDetectionCallable` protocol here:
https://github.com/aeye-lab/pymovements/blob/c833d5a74b6b3ee9e050db7619d6507cfecf2015/src/pymovements/events/events.py#L32-L56

If this protocol seems unpractical we can either get rid of it again or create two more protocols for velocity and position only detection methods.

dkrako (changes requested):
As described above, this will need further changes in the datasets module.

dkrako:
As this PR is now very behind the current main I created an separate issue for the broader case and close this PR

PR #197, opened 2023-02-23
Datasets: Use explicit rootpath in case of basic Event class

dkrako:
Ok so this might be a bit confusing and I need some feedback from the reviewers if this is a sensible approach.

During writing tests for the `Dataset` class I encountered a strange bug which was hard to find.

The `rootpath` property currently points to a directory named after the class which sits in the specified `root` during initilization.

Examples from the docstring:

```
class CustomDataset(Dataset):
    pass
dataset = CustomDataset(root='data')
dataset.rootpath
```

output:
```
Path('data/CustomDataset')
```

I would say this is a very convenient usage of the `root` init argument, because in case of all the classes inheriting from `PublicDataset`, you can then just specify your base data directory path and don't have to worry about anything else as subdirectories are created automatically.

But in case we just want to use the bare `Event` class we would actually like to point the `rootpath` attribute to the `root` specified during initialization and don't create any implicit subdirectories.

```
dataset = Dataset(root='/path/to/your/dataset')
dataset.rootpath
```

this is the current output:
```
Path('/path/to/your/dataset/Dataset')
```

Nobody wants this additional directory named `Dataset`. The expected output would rather be:
```
Path('/path/to/your/dataset')
```

So with this PR I propose to use the explicit path passed during initilization if the class is actually just `Dataset` and not a child class. We would have still the same output for `CustomDataset` but the alternative one in `Dataset`

But I am very unsure if this is a good approach as it can be confusing that the `rootpath` attribute is behaving differently for the two cases. I have already created an example in the docstring to make this more clear and obvious.

An alternative approach which would still cover the convenient way of passing the root data path to all the public datasets would be:

- move the current `rootpath()` definition to `PublicDatasaet`
- use the new `rootpath()` definition in `Dataset`

This would lead to 
 
```
class CustomDataset(Dataset):
    pass
dataset = CustomDataset(root='/path/to/your/dataset')
dataset.rootpath
```

output:
```
Path('/path/to/your/dataset')
```

but

```
class CustomPublicDataset(PublicDataset):
    pass
dataset = CustomPublicDataset(root='/path/to/your/dataset')
dataset.rootpath
```

output:
```
Path('/path/to/your/dataset/CustomPublicDataset')
```

What do you think about that?


dkrako:
There is also a more customizable approach possible, which could possibly be the best alternative:
Add the following arguments to `Dataset.__init__()`:

* `dataset_directory: str = '.'`
* `raw_directory: str = 'raw'`
* `preprocessed_directory: str = 'preprocessed'`
* `events_directory: str = 'events'`

Currently all the subdirectory name strings are hardcoded.

This would be the most flexible approach I think.
The PublicDataset subclasses would then use the class name as `dataset_directory`.

Moreover this introduces a lot of flexibility for local unpublished datasets, as these now also don't need this `raw` directory anymore in case they don't want to have that directory structure.

SiQube (approved):

PR #213, opened 2023-02-28
Datasets: Refactor attribute checks into own methods

dkrako:
Tests for new dataset check methods will follow soon in this PR.


dkrako:
Added all the tests now

SiQube (approved):

PR #219, opened 2023-03-01
Dataset: Add subset of fileinfo columns to events dataframe

dkrako:
This now behaves consistently with the gaze dataframe which also has some of the fileinfo columns available (like subject_id)


SiQube (approved):
sweet

PR #221, opened 2023-03-01
Datasets: Add ToyDataset

dkrako:


tests/datasets/test_dataset.py
```diff
@@ -369,13 +426,12 @@ def test_pos2vel(dataset_configuration):
...
-        **original_schema,
-        'x_left_vel': pl.Float64,
-        'y_left_vel': pl.Float64,
-        'x_right_vel': pl.Float64,
-        'y_right_vel': pl.Float64,
-    }
+    vel_schema = {}
+    for column_name in original_schema.keys():
+        if column_name.endswith('_pix'):
+            vel_column_name = column_name.replace('_pix', '_vel')
```
SiQube, line 432: we only test `_pix` => `_vel` and `_pix` => `_dva`, we might also want to add `_dva`=> `_vel`

SiQube (approved):
very cool PR!

tests/datasets/test_dataset.py
dkrako: From #220 we actually have `_dva` -> `_vel` but no `_pix` to `_vel` anymore.
This branch doesn't do any changes related to that anymore.

Issue #181, opened 2023-02-21
Add tests to events.engbert to reach 100%


dkrako:
## Current Behavior
Test coverage is currently at 95%:
https://app.codecov.io/gh/aeye-lab/pymovements/blob/main/src/pymovements/events/engbert.py

## Expected Behavior
Test coverage should be at 100%

## Minimum acceptance criteria
- [ ] 100% test coverage
- [ ] add tests for at all supported threshold method strings

Use pytest.parametrize for all tests, see the other tests in the test file

dkrako:
this was done in #227

PR #227, opened 2023-03-02
Tests: add 100% coverage for pymovements.events.engbert

SiQube:
basically a hotfix. add a test for the last code transition that was missing from 100% test coverage.


dkrako (approved):
Somehow I thought that you will need to use some noise for the threshold calculation but it works without, perfect!

Thanks a lot for your work!

dkrako:
This fixes #181

PR #228, opened 2023-03-02
Tests: add test for zero candidates

SiQube:


dkrako (approved):
Thank you so much for your work!

FYI @prassepaul : these are the missing tests I spoke about, so no more work needed

PR #241, opened 2023-03-03
Tests: add test coverage for pymovements.utils._decompress

SiQube:
the testcase is unreachable when extract_archive only is used

this coverage enables us to cover the case if _decompress is used

we advise that this method is not used as a standalone


dkrako (approved):
Thanks a lot!

PR #244, opened 2023-03-05
Codestyle: Unify codestyle across all files

SiQube:
unifying codestyle across files


SiQube:
we have to ignore one line for pylint -- irrespective of the rest of the changes -- see inline comment in pymovments.utils.downloads

dkrako (approved):
LGTM!

PR #254, opened 2023-03-06
Tests: Rename files with two underscores

SiQube:
my renaming script did not account for files with two underscores -- now fixed


dkrako (approved):
LGTM!

Issue #31, opened 2022-10-14
Implement heatmap plot


dkrako:
## Description of the problem
Eye tracking researchers often do a heatmap plot to see which positions where looked at most frequently.

## Description of a solution
- [x] implement plotting function
- [x] create an example notebook to visualize this plotting function

## Minimum acceptance criteria
- [ ] example notebook that demonstrates plotting function

dkrako:
in today's meeting we decided to use the gaze position samples as source for the heatmap, not the fixation data.

dkrako:
here's a relevant stackoverflow answer on how to plot your data:
https://stackoverflow.com/a/54088910

dkrako:
It would be nice if the function would have this signature:

```
heatmap(
    gaze: pl.DataFrame,
    position_columns: tuple[str, str] = ('x_pos', 'y_pos'),
    title: str | None = None,
    xlabel: str | None = None,
    ylabel: str | None = None,
    show: bool = True,
    savepath: str | None = None,
)
```

PR #276, opened 2023-03-13
Add Heatmap plotting function and notebook for demonstration (#31)

jakobchwastek:
## Description

I implemented a heatmap function using `histogram2d` and `imshow` and a very very basic notebook. I used imshow cause i thought  it has better interpolation capabilities than `pcolormesh`. Maybe you have to correct me. 
Please have a look and tell me what needs to be adjusted.

Fixes issue #31

## Implemented changes

- [x] Implement heatmap function
- [x] tutorial Notebook

## Type of change

## How Has This Been Tested?

This has just been tested by using the function in the tutorial notebook

## Checklist:


dkrako (changes requested):
This is very clean and has a clever way of using the experiment definition from the gaze dataframe.

I would just like to remove the `origin` argument from the plotting function.
this should be inferable from the `experiment.screen.origin` shouldn't it?

Is this related to the bug which you mentioned, where the origin should actually be `upper left` instead `lower left`?

src/pymovements/plotting/heatmap.py
```diff
@@ -0,0 +1,158 @@
...
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+"""
+Heatmap module.
+"""
+import matplotlib.pyplot as plt
```
dkrako, line 23: you need an additional import to get rid of all the test errors:

```
```

dkrako (changes requested):

jakobchwastek:
> I would just like to remove the `origin` argument from the plotting function.
> this should be inferable from the `experiment.screen.origin` shouldn't it?
> 
> Is this related to the bug which you mentioned, where the origin should actually be `upper left` instead `lower left`?

Yes, but i think when data is loaded with the correct origin in the first place it is not necessary to use origin here at all. How I understand all data should have the same coordinates system when choosing the correct origin in the Experiment/Screen class.

So should I remove argument or leave it for now?

dkrako:
> Looks like there is a problem because GazeDataFrame.experiment is optional..

Hmm.. okay then for this PR let's just raise an exception if experiment is None

dkrako:
If mypy still doesn't get it, just add a line with `assert gaze.experiment is not None`

we will fix this in a future PR

dkrako (changes requested):

src/pymovements/plotting/heatmap.py
```diff
@@ -0,0 +1,168 @@
...
+        ymin, ymax = screen.y_min_dva, screen.y_max_dva
+    else:
+        raise ValueError('Position columns must be in pixels or degrees')
+
+    # Define the grid and bin the gaze data
+    x_bins = np.linspace(xmin, xmax, num=gridsize[0])
+    y_bins = np.linspace(ymin, ymax, num=gridsize[1])
+
+    # Bin the gaze data
+    heatmap, x_edges, y_edges = np.histogram2d(x, y, bins=[x_bins, y_bins])
```
dkrako, line 128: naming the variable `heatmap` is problematic as it is also the function name.
our linters don't accept that.

name the variable `heatmap_values` or similar

dkrako (changes requested):

dkrako (changes requested):
for testing you can just copy the testing functions from here:
https://github.com/aeye-lab/pymovements/blob/main/tests/plotting/traceplot_test.py

then just adjust the function and parameter names accordingly to get 100% coverage.

dkrako (approved):
Thanks a lot! LGTM!

PR #286, opened 2023-03-15
Feature: add `pymovements.utils.asc2csv`

SiQube:
convert ascii file to csv

tests both created dataframes and sent messages


src/pymovements/utils/__init__.py
```diff
@@ -33,6 +33,7 @@
     pymovements.utils.paths
```
dkrako, line 33: please add the new module to the toctree here in the docstring, otherwise it won't show up in the documentation

src/pymovements/utils/asc2csv.py
dkrako, line 1: can you rename the file to `parsing.py`?
The naming is a bit more descriptive that way.
The function name can stay the same.

dkrako (changes requested):
Nice! This looks already great.
Just some minor corrections and it's good to go.

We will write a tutorial for this in the future.

src/pymovements/utils/__init__.py
SiQube: done

src/pymovements/utils/asc2csv.py
SiQube: done

dkrako (approved):
Thank you so much for your work!

PR #287, opened 2023-03-15
Git: Ignore notebook checkpoints

dkrako:


.gitignore
```diff
@@ -49,6 +49,7 @@ docs/build
 docs/doctree*
 docs/source/reference/pymovements.*.rst
 docs/source/tutorials/data
+docs/source/tutorials/.ipynb_checkpoints
```
SiQube, line 52: we could also just do .ipynb_checkpoints but does not really matter

PR #294, opened 2023-03-16
Dataset: don't download in init() but implicitly extract in download()

dkrako:
## Description

This PR removes the download and extract parameters from the initialization of the `PublicDataset` class.
Instead, the user is expected to call `PublicDataset.download()` explicitly. Extracting is done implicitly in this call.

At first this might seem odd as we might already be comfortable how public datasets are downloaded.

But the way it was implemented was rather bad practice, as a constructor is not supposed to do much more than just initializing your object.
That's why `download` was set to `False` per default in `PublicDataset.__init__()`.

Now we have a much more separated responsibilities in the methods:

`PublicDataset.__init__()` initializes your dataset object without any unexpected downloads.

`PublicDataset.download()` downloads and extracts your dataset.

Using `PublicDataset.download(extract=False)` you can disable this.

Using `PublicDataset.download(remove_finished=True` you can remove the original archive files.

And just for comparison to see that now it's also much cleaner from user perspective:

New:
```python
dataset = ToyDataset(root='data')
dataset.download(remove_finished=True)
```

Old:
```python
dataset = ToyDataset(
    root='data',
    download=True,
    remove_finished=True,
)
```

As `download()` now also returns a referrence to `self` we can even to that:
```python
dataset = ToyDataset(root='data').download(remove_finished=True)
```

## Implemented changes

- [x] Removed download and extract behavior from `init()`
- [x] add `extract` and `remove_finished` argument to `download()`
- [x] Updated examples accordingly
- [x] Updates all tutorials accordingly

## Type of change

## How Has This Been Tested?

Two tests were slightly adjusted:

- [x] `test_dataset_download` -> `test_dataset_download_no_extract` by setting `download(extract=False)`
- [x] `test_custom_dataset_download_extract` -> `test_custom_dataset_download_extract` by checking that `download()` also extracts by default

## Checklist:

### DIFF
--- a/docs/source/tutorials/R-tutorial.md
+++ b/docs/source/tutorials/R-tutorial.md
@@ -27,9 +27,15 @@ Access functions and data within python modules and classes via the `$` operator
 To test, you can proceed with the "pymovements in 10 minutes" tutorial,
 for example this is how you download the ToyDataset:
 ```r
-dataset = pm$datasets$ToyDataset(root='testdata', download=TRUE)
+dataset = pm$datasets$ToyDataset(root='data_directory')
+dataset$download()
 ```
 
+Now let's load in the dataset into R and display the found files:
+```r
+dataset$load()
+dataset$fileinfo
+```
 
 
 #### Related handy functions:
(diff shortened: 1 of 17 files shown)


SiQube (approved):
great PR

PR #293, opened 2023-03-16
Dataset: Return self instead of None

dkrako:
## Description

This allows for method chaining in the dataset class like so:

```python
dataset.pix2deg().pos2vel('smooth')
```

## Implemented changes

- [ ] Return self for all methods that returned None before
- [ ] Updated docstrings everywhere
- [ ] Updated function type signatures

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

- [ ] mypy static type checking

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] I have checked my code and corrected any misspellings

PR #295, opened 2023-03-17
CI: Check types more thoroughly with mypy (#295)

SiQube:
ass discussed in PR #293 we add 

```toml
[tools.mypy]
check_untyped_defs = true
disallow_any_generics = true
disallow_incomplete_defs = true
disallow_untyped_defs = true
warn_redundant_casts = true
warn_unused_ignores = true
```

to pyproject.toml. 

all code was adjusted accordingly. in some places I didn't figure out what exactly the type is and had to insert Any -- might change this later

### DIFF
--- a/docs/__init__.py
+++ b/docs/__init__.py
@@ -1,0 +1,19 @@
+# Copyright (c) 2023 The pymovements Project Authors
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
--- a/src/pymovements/events/events.py
+++ b/src/pymovements/events/events.py
@@ -125,7 +125,7 @@ class EventDataFrame:
     def __len__(self) -> int:
         return self.frame.__len__()
 
-    def __getitem__(self, *args, **kwargs) -> Any:
+    def __getitem__(self, *args: str, **kwargs: str) -> Any:
         return self.frame.__getitem__(*args, **kwargs)
 
     @property
@@ -133,7 +133,7 @@ class EventDataFrame:
         """List of column names."""
         return self.frame.columns
 
-    def _add_duration_property(self):
+    def _add_duration_property(self) -> None:
         """Adds duration property column to dataframe."""
         self.frame = self.frame.select([pl.all(), duration().alias('duration')])
 
@@ -179,7 +179,7 @@ class EventDetectionCallable(Protocol):
             velocities: list[list[float]] | list[tuple[float, float]] | np.ndarray,
             timesteps: list[int] | np.ndarray | None = None,
             minimum_duration: int = 0,
-            **kwargs,
+            **kwargs: int | str | float,
     ) -> EventDataFrame:
         """Minimal interface to be implemented by all event detection methods.
 
--- a/src/pymovements/events/events.py
+++ b/src/pymovements/events/events.py
@@ -88,7 +88,10 @@ class EventDataFrame:
 
             # Make sure lengths of onsets and offsets are equal.
             if onsets is not None:
+
+                # mypy does not get that offsets cannot be None (l. 87)
                 assert offsets is not None
+
                 checks.check_is_length_matching(onsets=onsets, offsets=offsets)
                 # In case name is given as a list, check that too.
                 if isinstance(name, Sequence) and not isinstance(name, str):
--- a/src/pymovements/events/events.py
+++ b/src/pymovements/events/events.py
@@ -125,7 +125,7 @@ class EventDataFrame:
     def __len__(self) -> int:
         return self.frame.__len__()
 
-    def __getitem__(self, *args: str, **kwargs: str) -> Any:
+    def __getitem__(self, *args: Any, **kwargs: Any) -> Any:
         return self.frame.__getitem__(*args, **kwargs)
 
     @property
--- a/src/pymovements/events/events.py
+++ b/src/pymovements/events/events.py
@@ -88,6 +88,7 @@ class EventDataFrame:
 
             # Make sure lengths of onsets and offsets are equal.
             if onsets is not None:
+                assert offsets is not None
                 checks.check_is_length_matching(onsets=onsets, offsets=offsets)
                 # In case name is given as a list, check that too.
                 if isinstance(name, Sequence) and not isinstance(name, str):
--- a/src/pymovements/utils/checks.py
+++ b/src/pymovements/utils/checks.py
@@ -23,9 +23,9 @@ This module holds basic checks which will be reused in other modules.
 from __future__ import annotations
 
 from typing import Any
+from typing import Sized
 
 import numpy as np
-import polars as pl
 
 
 def check_no_zeros(variable: Any, name: str = 'variable') -> None:
@@ -95,9 +95,7 @@ def check_shapes_positions_velocities(positions: np.ndarray, velocities: np.ndar
         )
 
 
-def check_two_kwargs(
-        **kwargs: Any,
-) -> None:
+def check_two_kwargs(**kwargs: Any) -> None:
     """Check if exactly two keyword arguments are given.
 
     Parameters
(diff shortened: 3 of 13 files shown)


docs/__init__.py
dkrako, line 1: is this file required?

src/pymovements/datasets/dataset.py
```diff
@@ -432,7 +432,9 @@ def pix2deg(self, verbose: bool = True) -> Dataset:
 
         return self
 
-    def pos2vel(self, method: str = 'smooth', verbose: bool = True, **kwargs) -> Dataset:
+    def pos2vel(
+            self, method: str = 'smooth', verbose: bool = True, **kwargs: int | float | str,
```
dkrako, line 436: If the function signature takes up more than a line I personally prefer to have each argument on a separate line.

src/pymovements/datasets/dataset.py
```diff
@@ -473,7 +475,7 @@ def detect_events(
             eye: str | None = 'auto',
             clear: bool = False,
             verbose: bool = True,
-            **kwargs,
+            **kwargs: int,
```
dkrako, line 478: can you use `float | int | str` here? then it's equal to `pymovements.events.EventDetectionCallable`

src/pymovements/events/events.py
```diff
@@ -125,15 +125,15 @@ def schema(self) -> pl.type_aliases.SchemaDict:
     def __len__(self) -> int:
         return self.frame.__len__()
 
-    def __getitem__(self, *args, **kwargs) -> Any:
+    def __getitem__(self, *args: str, **kwargs: str) -> Any:
```
dkrako, line 128: the `__getitem__()` method from `polars.DataFrame` is overloaded with a range of different types:
https://github.com/pola-rs/polars/blob/master/py-polars/polars/dataframe/frame.py#L1467-L1657

`polars` is actually a bit confusing when used with this traditional bracket indexing.
`df['a']` will get you the column 'a' as a series, `df[1]` will give you the first row as a dataframe and
`df[1, 'a']` will give you the value of the specific cell.

maybe we should use the following signature to be a bit more  consistent with the polars interface?
```
    def __getitem__(
        self,
        item: str | int | np.ndarray[Any, Any] | tuple[int, int] | tuple[int, str]),
    ) -> Any:
```

 or the full deal by supporting slices as indices:

```
MultiRowSelector: TypeAlias = slice | range | list[int] | pl.Series
MultiColSelector: TypeAlias = slice | range | list[int] | list[str] | list[bool] | pl.Series

def __getitem__(
        self,
        item: (
            str
            | int
            | np.ndarray[Any, Any]
            | MultiColSelector
            | tuple[int, MultiColSelector]
            | tuple[MultiRowSelector, MultiColSelector]
            | tuple[MultiRowSelector, int]
            | tuple[MultiRowSelector, str]
            | tuple[int, int]
            | tuple[int, str]
        ),
    ) -> DataFrame | Series:
```

src/pymovements/utils/checks.py
```diff
@@ -94,7 +95,9 @@ def check_shapes_positions_velocities(positions: np.ndarray, velocities: np.ndar
         )
 
 
-def check_two_kwargs(**kwargs) -> None:
+def check_two_kwargs(
```
dkrako, line 98: I think this can be a single line

src/pymovements/utils/checks.py
```diff
@@ -111,7 +114,9 @@ def check_two_kwargs(**kwargs) -> None:
         raise ValueError('there must be exactly two keyword arguments in kwargs')
 
 
-def check_is_mutual_exclusive(**kwargs) -> None:
+def check_is_mutual_exclusive(
```
dkrako, line 117: I would use `Any` here, as the type of the kwarg really does not matter

src/pymovements/utils/checks.py
```diff
@@ -137,7 +142,9 @@ def check_is_mutual_exclusive(**kwargs) -> None:
         )
 
 
-def check_is_none_is_mutual(**kwargs) -> None:
+def check_is_none_is_mutual(
+        **kwargs: None | list[int] | np.ndarray,
```
dkrako, line 146: I would use `Any` here, as the type of the kwarg really does not matter

src/pymovements/utils/checks.py
```diff
@@ -164,7 +171,10 @@ def check_is_none_is_mutual(**kwargs) -> None:
         )
 
 
-def check_is_length_matching(**kwargs) -> None:
+def check_is_length_matching(
+        **kwargs: np.ndarray | list[list[float]] | list[tuple[float, float]] |
```
dkrako, line 175: simply use `Sized` here

```
from typing import Sized
```

src/pymovements/utils/checks.py
```diff
@@ -183,5 +193,10 @@ def check_is_length_matching(**kwargs) -> None:
     value_1 = kwargs[key_1]
     value_2 = kwargs[key_2]
 
+    # mypy does not understand value_1 and value_2 can't be None
```
dkrako, line 196: will this still be needed if `None` is not one of the accepted types (is not `Sized`)?

src/pymovements/utils/decorators.py
```diff
@@ -25,16 +25,16 @@
 from typing import Any
 
 
-def auto_str(cls):
+def auto_str(cls: Any) -> str:
```
dkrako, line 28: this is actually a decorator and should return the same class as the input class.

you can achieve this by:
```
ClassT = TypeVar("ClassT") 

def auto_str(cls: ClassT) -> ClassT:
```
see here:
https://stackoverflow.com/a/65613529

In the example they also use the following argument:
`bound=Callable[..., Any]`. I don't know if this is applicable in our case, as this decorator should only be used for classes not functions.

src/pymovements/utils/decorators.py
```diff
@@ -25,16 +25,16 @@
...
     Automatically generate __str__() to include all arguments. Can be used as a decorator.
     """
-    def shorten(value: Any):
+    def shorten(value: Any) -> str:
         if isinstance(value, float):
             value = f'{value:.2f}'
         return value
 
-    def __str__(self):
+    def __str__(self: Any) -> str:
```
dkrako, line 37: it's a bit strange seing `self ` being typed but is necessary as `auto_str` is a function not a class, right?

dkrako (changes requested):
Thanks a lot for this work!
Seeing the commit history this seems to have been a tough fight with mypy and the external dependencies.

Some small details should be changed though.

Again, thanks a lot for the time you sunk into this.

docs/__init__.py
SiQube: unfortunately otherwise mypy doesn't pick up on it in the pyproject toml file (to ignore typed definitions), another solution would be to type the definitions in `docs/`

src/pymovements/datasets/dataset.py
SiQube: I then get:

```console
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

src/pymovements/datasets/dataset.py:554: error: Argument 4 to "__call__" of "EventDetectionCallable" has incompatible type "**Dict[str, Union[float, int, str]]"; expected "int"  [arg-type]
Found 1 error in 1 file (checked 1 source file)
```

docs/__init__.py
dkrako: alright, then lets keep it that way

src/pymovements/datasets/dataset.py
dkrako: that's very strange, but then lets keep it that way for now and fix it later when trouble comes up.

i think it's because minimum duration is not explicit in this method signature although it's part of the `EventDetectionCallable` signature

src/pymovements/datasets/dataset.py
SiQube: done

src/pymovements/events/events.py
SiQube: done

src/pymovements/utils/checks.py
SiQube: done

src/pymovements/utils/checks.py
SiQube: done

src/pymovements/utils/checks.py
SiQube: done

src/pymovements/utils/checks.py
SiQube: we then get

```console
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

src/pymovements/events/events.py:91: error: Argument "offsets" to "check_is_length_matching" has incompatible type "Union[List[int], ndarray[Any, Any], None]"; expected "Sized"  [arg-type]
Found 1 error in 1 file (checked 1 source file)
```

src/pymovements/utils/checks.py
SiQube: no, but see comment above (https://github.com/aeye-lab/pymovements/pull/295#discussion_r1165202105)

src/pymovements/utils/decorators.py
SiQube: yea :(

src/pymovements/utils/decorators.py
SiQube: done

src/pymovements/utils/checks.py
SiQube: I could of course use Any

SiQube:
@dkrako ready for review

src/pymovements/events/events.py
```diff
@@ -88,6 +88,7 @@ def __init__(
 
             # Make sure lengths of onsets and offsets are equal.
             if onsets is not None:
+                assert offsets is not None
```
dkrako, line 91: Please use `if offsets is None: raise ValueError()` here.

Asserts in production code are problematic and should be used exclusively for stuff that should never happen independent of user input.
If something is failing due to user input error or other dependent systems please use an exception instead.

See this article here: https://www.inspiredpython.com/article/assertions-and-exceptions-are-not-the-same

dkrako (changes requested):
Thanks a lot for this work!

One small change I would like to request: please replace this one assert with raising an exception. Then we're good to go!

src/pymovements/events/events.py
SiQube: this happens in [line 86-87](https://github.com/aeye-lab/pymovements/pull/295/files#diff-82cc6c96b216cd7ff8cefcac7928eb28601370e2180b3729e129a73e6c31fe7fR86-R87)

```python
# Make sure that if either onsets or offsets is None, the other one is None too.
checks.check_is_none_is_mutual(onsets=onsets, offsets=offsets)
```

SiQube: I'll pushed another commit with a comment explicitly pointing at line 87 that it is within mypy and in line 91 there can not be any none offsets

dkrako: ah sorry I didn't see that correctly, so everything's fine and I'll merge after the tests complete

Thanks a lot again!

dkrako (approved):

PR #298, opened 2023-03-23
Datasets: Use PublicDatasetDefinition to initialize a PublicDataset

dkrako:
## Description

This PR introduces the `PublicDatasetDefinition` class to initialize a `PublicDataset`.
The point with this is to clearly separate repsonsibilities:

The responsibility of the `PublicDataset` is now basically only the downloading and extracting of the dataset.
The information about the dataset itself has been completely moved to the new `PublicDatasetDefinition` class.

Each `PublicDatasetDefinition` is registered as a public dataset and can be referenced via string.
This makes it much more simple when using text configuration files, as you don't need to reference a classes anymore, but just use a string.

The `PublicDataset` initializer is now not crowded anymore which arguments. I expect that the number of attributes in defintions of public datasets will increase as we add new functionality like loading trial vars and a broad range of stimuli data.
Additionally, the `PublicDatasetDefinitions` are now made much more transparent as each attribute is public now.
All examples were adjusted for this.

Datasets can now no longer be initiliazed as so:
```python
dataset = pm.datasets.ToyDataset(root='data')
```

But are now initialized this way:
```python
dataset = pm.PublicDataset('ToyDataset', root='data')
```

You can also pass the class:
```python
dataset = pm.PublicDataset(pm.datasets.ToyDataset, root='data')
```

A `PublicDatasetDefinition` is a `dataclass` and can be registered using the  `@register_public_dataset` decorator:
```python
@dataclass
@register_public_dataset
class NewDataset(PublicDatasetDefinition):
    ...
```

As the definition is a `dataclass` you can also overwrite attributes via init:
```python
dataset = pm.PublicDataset(pm.datasets.ToyDataset(mirrors=['https://alternative.mirror.com']), root='data')
```

# Implemented changes

- [x] Move the datasets definitions to a separate directory
- [x] create a new decorator `@register_public_dataset` which registers datasets in a central dictionary
- [x] adjust `PublicDataset` class such that it now only accepts definitions.
- [x] adjusted examples and tutorials

## Type of change

## How Has This Been Tested?

- [x] The public dataset tests where adjusted.
- [x] The correct registering of public datasets was tested.

## Checklist:


SiQube (approved):
LGTM

PR #317, opened 2023-04-13
Gaze: Rewrite transforms as polars expressions

dkrako:
description coming soon


SiQube (approved):
very very very cool PR!

I personally like

```python
if a:
    ...
elif b:
    ...
else:
    ...
```

more than 

```python
if a:
    ...

elif b:
    ...

else:
    ...
```
but that is personal preferences. (I think I like it more because it is easier to distuingish between multiple ifs)

dkrako:
> I personally like
> 
> ```python
> if a:
>     ...
> elif b:
>     ...
> else:
>     ...
> ```
> 
> more than
> 
> ```python
> if a:
>     ...
> 
> elif b:
>     ...
> 
> else:
>     ...
> ```
> 
> but that is personal preferences. (I think I like it more because it is easier to distuingish between multiple ifs)

I see. I have adjusted the file with used the if-elif-else construct.

SiQube (dismissed):

SiQube:
we now have once `from typing import Callable` and once `from collections.abc import Callable`, is there a reason why?

dkrako:
> we now have once [`from typing import Callable`](src/pymovements/gaze/transforms_pl/transforms_library.py) and once [`from collections.abc import Callable`](src/pymovements/gaze/transforms_pl/savitzky_golay.py), is there a reason why?

pylint will raise an error when using `typing.Callable`:
`[R6002(consider-using-alias), ] 'typing.Callable' will be deprecated with PY39, consider using 'collections.abc.Callable' instead`

But the strange thing is, that python3.7 complains with `collections.abc.Callable` in  `transforms_library.py`.
Seems to have to do with the `...` in the specification of the `Callable` signature.
pylint also does not complain about that place so I think that's the best way to do for now.

SiQube (approved):

Issue #319, opened 2023-04-14
Curly braces format instead of regex expression for specifying dataset files


dkrako:
## Description of the problem
Currently regex expressions with named groups have to be used to specify the files of a dataset.
This is not very beginner-friendly to say the least. 

## Description of a solution
Instead it would be nice to specify dataset files by using the curly braces python formatting.

So instead of using something like this:
```python
r'(?P<subject_id>\d+)_(?P<session_id>\d+).csv'
```

it would be much nicer to support this:
```python
r'{subject_id}_{session_id}.csv'
```

## Minimum acceptance criteria
- [ ] implement a function `curly_to_regex(s: str) -> re.Pattern` in `utils.strings`
- [ ] replace `DatasetDefinition.filename_regex` with  `DatasetDefinition.filename_format`

PR #325, opened 2023-04-15
Dataset: Use curly braces format instead of pure regex

ketrab2003:
## Description

I added curly braces format for specifying dataset files.

I minimally changed specified functionality by adding need to specify type of value (by default it's string), I think this change is logical because strings are also default option in `filename_format_dtypes`.

So now instead of:
```
r'(?P<subject_id>\d+)_(?P<session_id>\d+).csv'
```
it can be written as:
```
r'{subject_id:d}_{session_id:d}.csv'
```
All other nuances are specified in function's docstring.

Also, old regex method should still work if needed, this change only adds functionality.

Fixes issue #319

## Implemented changes

- Implemented function `curly_to_regex(s: str) -> re.Pattern` in utils.strings;
- Added unit tests for `curly_to_regex` function (100% coverage);
- Replaced `DatasetDefinition.filename_regex` with `DatasetDefinition.filename_format`;
- Also replaced `DatasetDefinition.filename_regex_dtypes` with `DatasetDefinition.filename_format_dtypes`;
- Changed one docs tutorial accordingly, however I'm not shure if further documentation changes are not needed.

## Type of change

## How Has This Been Tested?

I added unit tests and ran `tox` command, all tests passed.

## Checklist:

### DIFF
--- a/docs/source/tutorials/local-dataset.ipynb
+++ b/docs/source/tutorials/local-dataset.ipynb
@@ -114,7 +114,7 @@
    "id": "cd6ab2fe",
    "metadata": {},
    "source": [
-    "We also define a `filename_regex` which is a regular expression used to match and extract values from filenames of data files in the dataset. For example, `r'trial_(?P<text_id>\\d+)_(?P<page_id>\\d+).csv'` will match filenames that follow the pattern `trial_{text_id}_{page_id}.csv` and extract the values of `text_id` and `page_id` for each file."
+    "We also define a `filename_format` which is a pattern expression used to match and extract values from filenames of data files in the dataset. For example, `r'trial_{text_id:d}_{page_id:d}.csv'` will match filenames that follow the pattern `trial_{text_id}_{page_id}.csv` and extract the values of `text_id` and `page_id` for each file."
    ]
   },
   {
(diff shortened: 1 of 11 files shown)


dkrako (approved):
Thank you so much for this high quality PR!

It's exactly what I thought what needs to be done, even with the extra of specifying character types with the expected number of characters.
It even supports escaping the curly braces by using `{{` and `}}`, nice!

There's one small issue left, but we can address this in a follow-up issue/PR as it's just related to possible user errors.

The issue is with unbalanced opening and closing braces like these two cases:

```python
>>> strings.curly_to_regex('{test}}')
re.compile('(?P<test>.+)}')
```

```python
>>> strings.curly_to_regex('{{test}')
re.compile('{test}')
```

These are clearly ill-formed format strings and when using f-strings these would raise errors:

```python
>>> f'{{test}'
  File "<stdin>", line 1
    f'{{test}'
              ^
SyntaxError: f-string: single '}' is not allowed
```
```python
>>> f'{test}}'
  File "<stdin>", line 1
    f'{test}}'
              ^
SyntaxError: f-string: single '}' is not allowed
```

f-strings also raise errors with this:

```python
>>> f'{test'
  File "<stdin>", line 1
    f'{test'
            ^
SyntaxError: f-string: expecting '}'
```

But your implementation also makes perfect sense:
```python
>>> strings.curly_to_regex('{test')
re.compile('{test')
```

As what I'm writing about is clearly a further enhancement, I will just create an issue out of this and merge this PR as it is.

Thanks again for this work!

PR #342, opened 2023-04-28
feat!: Add pattern argument to parse_eyelink() to be more flexible

dkrako:


src/pymovements/utils/parsing.py
```diff
@@ -52,96 +53,146 @@ def check_nan(sample_location: str) -> float:
...
+
+    return additional_columns
+
+
 def parse_eyelink(
         filepath: Path,
-        sync_msg_start_pattern: None | str = None,
-        sync_msg_stop_pattern: None | str = None,
+        patterns: list | None = None,
+        schema: dict | None = None,
```
SiQube, line 114: note that `list`/ `dict` is not a helpful type since we do not know what the list contains.

dkrako: thanks a lot for taking care of this!

PR #349, opened 2023-05-09
feat(dataset): Add verbose parameter to more dataset methods

dkrako:


SiQube (commented):
maybe we should either use int or bool? what's the reasoning behind using both?

dkrako:
The `int` only makes sense if there are different levels of verbosity implemented.
Otherwise it is misleading.

Moreover, the `bool` is used only for the method in `dataset_download.py` which is not part of the public API.

SiQube (approved):

PR #431, opened 2023-05-30
fix: Keep polars version below v0.18.0

dkrako:
there are still some small details to work out to migrate to polars v0.18

in the meantime we will stay safely below


dkrako:
closed in favor of #432

PR #440, opened 2023-06-06
feat(gaze): Add GazeDataFrame.transform()

dkrako:
This is still a work in progress!


dkrako:
I have ignored one line for pylint for this error: [cell-var-from-loop](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/cell-var-from-loop.html)

caused by these lines:
https://github.com/aeye-lab/pymovements/blob/40f85a63bf188e264fc52fff4515ff881e754d44/src/pymovements/gaze/transforms_pl/pix2deg.py#L79-L86

But now I have read a bit further and found this reddit thread:
https://www.reddit.com/r/learnpython/comments/kvct0d/pylint_cell_variable_val_defined_in_loop_why/

I toyed around a bit and pylint has actually exposed a bug here I think.
I will have to revisit this next week.

dkrako:
Wow pylint really exposed a bug here. I didn't know that was a potential problem in python.

If I understand correctly, the unexpected behavior comes from the delayed creation of the lambda function.
In the previous code snippet `component` would then always be the final value (`n_components - 1`).

As a very simple example, the output of this is crazy:

```python
>>> multipliers = [lambda x : i * x for i in range(4)]
>>> print([m(2) for m in multipliers])
[6, 6, 6, 6]
```

This is the fix for the problematic code of my last comment:

https://github.com/aeye-lab/pymovements/blob/a1dfad6ce83aac0226b97f71e864fbcc09716ce0/src/pymovements/gaze/transforms_pl/pix2deg.py#L81-L86

with

https://github.com/aeye-lab/pymovements/blob/a1dfad6ce83aac0226b97f71e864fbcc09716ce0/src/pymovements/gaze/transforms_pl/pix2deg.py#L91-L93

SiQube (approved):

PR #444, opened 2023-06-08
feat(gaze): Add column merge arguments to integration functions

dkrako:
This adds the new GazeDataFrame functionality to merge columns into position, velocity, etc, to the integration functions for numpy and pandas

### DIFF
--- a/src/pymovements/gaze/integration.py
+++ b/src/pymovements/gaze/integration.py
@@ -34,6 +34,10 @@ def from_numpy(
         schema: list[str],
         experiment: Experiment | None = None,
         orient: Literal['col', 'row'] = 'col',
+        pixel_columns: list[str] | None = None,
+        position_columns: list[str] | None = None,
+        velocity_columns: list[str] | None = None,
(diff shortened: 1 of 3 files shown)


SiQube (commented):
just to be sure: could you add tests with permutations of the x/y respective pixel, velocity etc.? e.g. `['y_pix', 'x_pix']` and also  `['y_pix', 'x_pix', 'x_vel', 'y_vel']`

should we maybe also add just providing `_pix` and then it uses `x_pix` and `y_pix`?

dkrako:
I personally would like to enforce the used order, so the cases you named are not supported and should raise an exception in the future. This issue is rather complicated and currently has low priority, so I would like to delegate adding these additional test cases.
Feel free to add a PR enforcing component order.

Regarding the second point, I think it makes more sense to handle this in another PR, as what you seem to be talking about  is extending the interface of `GazeDataFrame.__init__()` or `GazeDataFrame.merge_component_columns_into_tuple_column()`, which is out of scope of this PR.
Feel free to add a PR for your feature request.

PR #460, opened 2023-07-07
chore!: Drop support for python 3.7

### LINKED ISSUE #457: drop python 3.7 support
python 3.7 eol -- drop support? see [here.](https://endoflife.date/python)

we could have f-strings everywhere!

SiQube:
resolves #457


SiQube:
I additionally removed mandatory checks for 3.7

src/pymovements/utils/parsing.py
```diff
@@ -150,9 +150,8 @@ def parse_eyelink(
 
     for line in lines:
         for pattern_dict in compiled_patterns:
-            match = pattern_dict['pattern'].match(line)
 
-            if match:
+            if match := pattern_dict['pattern'].match(line):
```
dkrako, line 154: I think the walrus operator is handy for matching regexes (I always hated it to write two lines for this).

So keep this please.

src/pymovements/utils/paths.py
```diff
@@ -128,17 +128,15 @@ def match_filepaths(
...
-
-                filepath = childpath
-                if relative:
-                    # mypy is unaware that 'relative_anchor' can never be None (l.116)
-                    assert relative_anchor is not None
-                    filepath = filepath.relative_to(relative_anchor)
-
-                match_dict['filepath'] = str(filepath)
-                match_dicts.append(match_dict)
+        elif (match := regex.match(childpath.name)) is not None:
```
dkrako, line 131: Keep this please, but leave the `is not None` out

dkrako (changes requested):
I confirm that the walrus operator decreases readability in most cases.

I basically follow this line of thought:
https://therenegadecoder.com/code/the-controversy-behind-the-walrus-operator-in-python/

I think there are cases though, where the walrus operator indeed makes sense.
In this PR I would say that the regex match if clauses are perfectly fine with the walrus operator.

Please add `consider-using-assignment-expr` to the `pylintrc` under `disable`, maybe sort the entries alphabetically if you're at it.

Apart from that, thanks for all the work!

dkrako (changes requested):
I remembered one more thing which was necessary due to py37 support.

python 3.8 now has `typing.Literal` so we don't need `typing_extensions` anymore.

https://github.com/aeye-lab/pymovements/blob/e3ef6c91fd8671f6cba305cd645d7371c319f711/src/pymovements/gaze/integration.py#L26

Please also remove this dependency from the `pyproject.toml`:

https://github.com/aeye-lab/pymovements/blob/e3ef6c91fd8671f6cba305cd645d7371c319f711/pyproject.toml#L41

dkrako:
One more thing which I am unsure of: do we still need `from __future__ import annotations` everywhere?

We have this special pre-commit hook to add it everywhere:

https://github.com/aeye-lab/pymovements/blob/e3ef6c91fd8671f6cba305cd645d7371c319f711/.pre-commit-config.yaml#L21-L27

I'm not sure if this was due to python3.7 issues or if thisissue  will remain with python3.8?

SiQube:
> One more thing which I am unsure of: do we still need `from __future__ import annotations` everywhere?
> 
> We have this special pre-commit hook to add it everywhere:
> 
> https://github.com/aeye-lab/pymovements/blob/e3ef6c91fd8671f6cba305cd645d7371c319f711/.pre-commit-config.yaml#L21-L27
> 
> I'm not sure if this was due to python3.7 issues or if thisissue will remain with python3.8?

This will remain since there are changes from 3.8 => 3.9 and so on, e.g. [here are some 3.12 typing changes](https://docs.python.org/3.12/whatsnew/3.12.html). so I'd keep it!

dkrako (approved):

dkrako:
Thanks a lot for all the work!

PR #466, opened 2023-08-04
chore!: Restructure transforms modules for polars and numpy

dkrako:
## Description

This is the necessary breaking change which makes clear that from coming version on polars will be used for gaze transformations.

The old numpy transformations will still be available, but now only via `pymovements.gaze.transforms_numpy`

## Implemented Changes

* `pymovements.gaze.transforms` -> `pymovements.gaze.transforms_numpy`
* `pymovements.gaze.transforms_pl.*` -> `pymovements.gaze.transforms`


SiQube (approved):
Looks good to me!

PR #468, opened 2023-08-08
docs: Update requirements.txt to reflect location of dependencies

siqube-adobe:
while working on pymovements I realized that the requirements.txt does not reflect the correct location of dependencies -- here a quick PR.


dkrako:
nice catch! thanks a lot!

PR #469, opened 2023-08-08
feat: Add __version__ attribute

### LINKED ISSUE #404: Add __version__ attribute to pymovements
## Description of the problem
The `___version__` attribute is generally used for packages to get the used version.

But when we try this

```python
import pymovements as pm
pm.__version__
```
we get the following error:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'pymovements' has no attribute '__version__'
```

This is how other packages behave:
```python
>>> import numpy as np
>>> np.__version__
'1.24.2'
```

```python
>>> import pandas as pd
>>> pd.__version__
'1.4.2'
```

```python
>>> import polars as pl
>>> pl.__version__
'0.17.9'
```

siqube-adobe:
resolves #404


pyproject.toml
```diff
@@ -37,7 +38,8 @@ dependencies = [
   "polars>=0.18.3,<0.19.0",
   "pyarrow>=11.0.0,<12",
   "scipy>=1.5.4,<2",
-  "tqdm>=4.0.0,<5"
+  "tqdm>=4.0.0,<5",
+  "versioneer>=0.29"
```
dkrako, line 42: in non-vendor mode versioneer is only needed as a build dependency

pyproject.toml
```diff
@@ -62,6 +64,7 @@ test = [
   "pytest-cov>=4.0.0",
   "types-tqdm"
 ]
+toml = ["tomli; python_version < '3.11'"]
```
dkrako, line 67: this is not needed for Build-time dependency mode

src/pymovements/__init__.py
```diff
@@ -50,3 +50,6 @@
     'synthetic',
     'utils',
 ]
+
+from . import _version
```
dkrako, line 54: can you put it into the top and change it to `from pymovements import _version`

versioneer.py
dkrako, line 1: This complete file is not needed in Build-time dependency mode

pyproject.toml
```diff
@@ -2,6 +2,7 @@
 requires = [
   "setuptools>=42",
   "setuptools-git-versioning<2",
+  "versioneer[toml]==0.29",
```
dkrako, line 5: Don't we want to write it as `"versioneer[toml]>=0.29,<1.0"`?

dkrako (changes requested):
Thank you so much for working on this!

I think you have installed versioneer in vendored mode, which creates this `versioneer.py` file in the root directory.

If you install it in [Build-time dependency mode](https://github.com/python-versioneer/python-versioneer#build-time-dependency-mode), then installing `versioneer` is only needed during build/packaging process, together with some fancy `git archive` magic.

I just tested it locally, and for now I definitely prefer build-time dependency mode over the vendored mode.

pyproject.toml
siqube-adobe: changed

versioneer.py
siqube-adobe: done

src/pymovements/__init__.py
siqube-adobe: done

pyproject.toml
siqube-adobe: done

pyproject.toml
siqube-adobe: done

siqube-adobe:
@dkrako thank you for the review -- I've adjusted my code accordingly

src/pymovements/__init__.py
```diff
@@ -50,3 +51,8 @@
     'synthetic',
     'utils',
 ]
+
+__version__ = _version.get_versions()['version']
+
+from . import _version
```
dkrako, line 57: it seems you have forgotten to remove these two lines somehow?

dkrako (changes requested):

src/pymovements/__init__.py
siqube-adobe: done -- must've been reattached during some kind of merging

dkrako (approved):

PR #471, opened 2023-08-09
Fix trying new mirror print

### LINKED ISSUE #470: Trying next mirror messsage is printed although there is no next mirror
The following line should only be printed if there is a next mirror.

https://github.com/aeye-lab/pymovements/blob/66278f4fe5b90330ff0ceef3ae17def6ba4e757c/src/pymovements/dataset/dataset_download.py#L100

siqube-adobe:
resolves #470


dkrako (approved):
Thanks a lot!

PR #472, opened 2023-08-10
chore!: Use transform() in pix2deg(), pos2vel() and pos2acc()

dkrako:


dkrako:
There are 3 test cases which I just cannot get to pass.
The failing functions are `pos2vel()`, `pos2acc()` and `savitzky_golay()` from `pymovements.gaze.transforms` with all three having an empty series as input and expecting and empty series as output. 

The reason why they fail is coming from the following change in this PR:

OLD:
https://github.com/aeye-lab/pymovements/blob/dd122fd9d07bfebb8972abdfd3f6e6f583574cf1/src/pymovements/gaze/transforms.py#L537-L548

NEW:
https://github.com/aeye-lab/pymovements/blob/1b9d5b58eff0c9c61c9e21a3a9ca98d07d24bf6f/src/pymovements/gaze/transforms.py#L537-L542

This was changed, because refactoring the `GazeDataFrame.pos2vel()` method to use `GazeDataFrame.transform()` resulted in changes that raised a `polars.ComutationError`.

This is how `GazeDataFrame.pos2vel()` was implemented before:
https://github.com/aeye-lab/pymovements/blob/dd122fd9d07bfebb8972abdfd3f6e6f583574cf1/src/pymovements/gaze/gaze_dataframe.py#L314-L319

And this is how it is implemented now in `GazeDataFrame.transform()`:
https://github.com/aeye-lab/pymovements/blob/1b9d5b58eff0c9c61c9e21a3a9ca98d07d24bf6f/src/pymovements/gaze/gaze_dataframe.py#L275

We cannot just instantiate a `polars.Series` anymore, but need to return a `polars.Expr`.
This is the error that was raised on calling `GazeDataFrame.pos2vel()` with an empty dataframe before the changes in `savitzky_golay()`:

```
exceptions.ComputeError: The name: 'velocity' passed to `LazyFrame.with_columns` is duplicate
```
I just can't wrap my head around the issue... :exploding_head: 

As the test cases aren't very practical for most users, I would prefer to merge despite the problem and create an issue out of this. This post here should be a good starting point on how to fix the problem.

I have commented out the test cases here:

https://github.com/aeye-lab/pymovements/blob/1b9d5b58eff0c9c61c9e21a3a9ca98d07d24bf6f/tests/gaze/transforms/savitzky_golay_test.py#L167-L175

https://github.com/aeye-lab/pymovements/blob/1b9d5b58eff0c9c61c9e21a3a9ca98d07d24bf6f/tests/gaze/transforms/pos2acc_test.py#L117-L122

https://github.com/aeye-lab/pymovements/blob/1b9d5b58eff0c9c61c9e21a3a9ca98d07d24bf6f/tests/gaze/transforms/pos2vel_test.py#L160-L168

dkrako:
I need to update the tutorials a bit, some are failing and some texts are not up to date anymore.

But apart from this, the PR is ready for review.

SiQube (approved):

PR #473, opened 2023-08-10
feat: Add trial_columns and trialize transform() method

dkrako:


SiQube (approved):
very cool, I think we need a tutorial for it.

PR #512, opened 2023-09-08
ci: Add pydoclint to pre-commit hooks

dkrako:


SiQube:
I'm extremely conflicted about this PR. it is a lot of overhead but would improve readability for sure.

src/pymovements/dataset/dataset_paths.py
```diff
@@ -24,7 +24,29 @@
...
+        the preprocessed / event data. Default: `raw`
+    events : str
+        Name of directory under dataset path that will be used to store event data. We advise
+        the user to keep the event data separate from the original raw data. Default: `events`
+    preprocessed : str
+        Name of directory under dataset path that will be used to store preprocessed data. We
+        advise the user to keep the preprocessed data separate from the original raw data.
+        Default: `preprocessed`
+    downloads : str
+        Name of directory to store downloaded data.Default: `downloads`
```
prassepaul, line 48: data.Default -> data. Default (add whitespace)

src/pymovements/events/detection/_idt.py
```diff
@@ -71,24 +71,26 @@ def idt(
 
     Parameters
     ----------
-    positions: array-like, shape (N, 2)
+    positions: list[list[float]] | list[tuple[float, float]] | np.ndarray
         Continuous 2D position time series
-    timesteps: array-like, shape (N, )
+        shape (N, 2)
+    timesteps: list[int] | np.ndarray | None
```
prassepaul, line 77: add default values, where possible?

src/pymovements/events/detection/_ivt.py
```diff
@@ -63,12 +65,12 @@ def ivt(
         velocity is below the threshold, the point is classified as a fixation.
     include_nan: bool
         Indicator, whether we want to split events on missing/corrupt value (np.nan)
-    name:
+    name: str
```
prassepaul, line 68: see comment above. add default values

src/pymovements/events/detection/_microsaccades.py
```diff
@@ -53,15 +53,17 @@ def microsaccades(
...
+        shape (N, 2)
+    timesteps: list[int] | np.ndarray | None
         Corresponding continuous 1D timestep time series. If None, sample based timesteps are
         assumed.
+        shape (N, )
     minimum_duration: int
         Minimum saccade duration. The duration is specified in the units used in ``timesteps``.
          If ``timesteps`` is None, then ``minimum_duration`` is specified in numbers of samples.
-    threshold : np.ndarray, tuple[float, float] or str
+    threshold : np.ndarray | tuple[float, float] | str
```
prassepaul, line 66: see above. add defaults

prassepaul (changes requested):
please be consistent. we need to decide, if we want to add the default values or not.

SiQube:
I'll delete all the default values because it is not checked by the linter and hence manual effort to control for future PRs, opinions? @prassepaul @dkrako

dkrako:
> I'll delete all the default values because it is not checked by the linter and hence manual effort to control for future PRs, opinions? @prassepaul @dkrako

I agree. As we don't lint for default values (yet?), you can delete all default values from the docstrings.

.pre-commit-config.yaml
```diff
@@ -110,7 +115,7 @@ repos:
         files: ^tests/
         args:
           [
-            '--ignore=D103,D213',
+            '--add-ignore=D103,D107,D213',
```
dkrako, line 118: are these needed here still if these are already defined in pyproject.toml

src/pymovements/dataset/dataset.py
```diff
@@ -139,13 +139,6 @@ def scan(self) -> Dataset:
         -------
         Dataset
             Returns self, useful for method cascading.
-
-        Raises
```
dkrako, line 143: I'm not sure if I agree to remove the raises section. is pydoclint complaining?

src/pymovements/dataset/dataset.py
```diff
@@ -332,15 +301,9 @@ def pos2vel(
             Computation method. See :func:`~transforms.pos2vel()` for details, default: smooth.
         verbose : bool
             If True, show progress of computation.
-        **kwargs
+        **kwargs: Any
```
dkrako, line 304: Is pydoclint checking the spaces here?
Sometimes we use `parameter : type` sometimes `parameter: type`

Maybe we should be consistent with that?

But if we don't have a linter to support that this would be out of scope of this PR i think.

src/pymovements/dataset/dataset_library.py
```diff
@@ -33,6 +33,7 @@ class DatasetLibrary:
     ----------
     definitions:
         Dictionary of :py:class:`~pymovements.DatasetDefinition`.
+
```
dkrako, line 36: i'm confused, is this needed? I thought some of the docstring linters should complain about empty lines at the end?

src/pymovements/events/processing.py
```diff
@@ -35,20 +35,18 @@
 class EventProcessor:
     """Processes event and gaze dataframes.
 
-    Attributes
+    Parameters
```
dkrako, line 38: Shouldn't parameters come here as an addition?

`event_properties` refers to both the Parameter for the initialization, and the Attribute of the instance.

src/pymovements/events/processing.py
```diff
@@ -97,24 +90,22 @@ def process(self, events: EventDataFrame) -> pl.DataFrame:
 class EventGazeProcessor:
     """Processes event and gaze dataframes.
 
-    Attributes
+    Parameters
```
dkrako, line 93: Shouldn't parameters come here as an addition?

event_properties refers to both the Parameter for the initialization, and the Attribute of the instance.

src/pymovements/events/properties.py
```diff
@@ -53,15 +53,15 @@ def amplitude(
...
     Parameters
     ----------
-    position_column
+    position_column: str
         The column name of the position tuples.
-    n_components:
+    n_components: int
         Number of positional components. Usually these are the two components yaw and pitch.
 
-    Raises
```
dkrako, line 61: We shouldn't remove this section

src/pymovements/events/properties.py
```diff
@@ -94,15 +94,15 @@ def dispersion(
...
```
dkrako, line 102: We shouldn't remove this section

src/pymovements/events/properties.py
```diff
@@ -130,16 +130,16 @@ def disposition(
...
```
dkrako, line 138: We shouldn't remove this section

src/pymovements/gaze/experiment.py
```diff
@@ -31,55 +31,47 @@
 class Experiment:
     """Experiment class for holding experiment properties.
 
-    Attributes
```
dkrako, line 34: is pydoclint complaining about this section? If we delete this, we won't have documented anymore that we can get the screen of the experiment by using Experiment.screen.

src/pymovements/gaze/gaze_dataframe.py
```diff
@@ -341,11 +344,6 @@ def pix2deg(self) -> None:
 
         After success, the gaze dataframe is extended by the resulting dva position columns.
 
-        Raises
```
dkrako, line 344: We shouldn't remove this.

src/pymovements/gaze/screen.py
```diff
@@ -47,15 +47,22 @@ class Screen:
         Eye-to-screen distance in centimeters
     origin : str
         Specifies the screen location of the origin of the pixel coordinate system.
-    x_max_dva : float
```
dkrako, line 50: Removing this is not an option.

src/pymovements/utils/archives.py
```diff
@@ -48,21 +48,21 @@ def extract_archive(
...
-    verbose:
+    verbose : int
         Verbosity levels: (1) Print messages for extracting each dataset resource without printing
         messages for recursive archives. (2) Print additional messages for each recursive archive
         extract.
 
     Returns
     -------
-    Path :
+    Path : Path
```
dkrako, line 65: is there a reason for naming this return value Path?

is pydoclint complaining about the line otherwise?

maybe without the `:`?

src/pymovements/utils/archives.py
```diff
@@ -252,14 +252,15 @@ def _decompress(
...
-    destination_path : Path
+    destination_path : Path | None
         Path to the decompressed file.
     remove_finished : bool
         If ``True``, remove the source file after decompression.
 
     Returns
     -------
-    Path : Path to the decompressed file.
+    Path : Path
```
dkrako, line 262: is there a reason for naming this return value Path?

is pydoclint complaining about the line otherwise?

maybe without the `:`?

dkrako (commented):
Thank you for working through this chore!

I'm unsure about some of the changes.

Especially in the case of deleting the Attributes and Raises sections existing documentation is just removed without replacement.

dkrako:
Are there any updates for this PR?

If removing the `Attributes` and `Raises` sections was due to some pydoclint error, I would very much prefer to put an ignore to the specific error code and fix the problems in separate PRs. Otherwise the scope of this PR will be to big to handle.

SiQube:
ah sorry forgot -- yes the raises and so on were from pydoclint, I'll look for an argument to ignore this.

.pre-commit-config.yaml
SiQube: will check

src/pymovements/dataset/dataset.py
SiQube: pydoclint complains that the function does not have a raise in it, it just stupidly checks for raises within scope of function :-( -- ignore?

dkrako: yes, let's ignore all errors from Raises and Attributes sections for this PR

I hope there's some code like DXXX for this

src/pymovements/dataset/dataset.py
SiQube: it only checks if there is a type given not the correct format, I can check whether there is a linter for spaces.

I also think if we have strict `mypy` typing we do not necessarily need types within docstrings.

src/pymovements/dataset/dataset_library.py
SiQube: hmm, don't know, actually it should -- I'll check

src/pymovements/events/processing.py
SiQube: yes it should -- will fix it

src/pymovements/events/processing.py
SiQube: yes it should -- will fix it

src/pymovements/events/properties.py
SiQube: pydoclint complains that the function does not have a raise in it, it just stupidly checks for raises within the scope of function :-( -- ignore the pydoclint error?

src/pymovements/events/properties.py
SiQube: pydoclint complains that the function does not have a raise in it, it just stupidly checks for raises within the scope of function :-( -- ignore the pydoclint error?

src/pymovements/events/properties.py
SiQube: pydoclint complains that the function does not have a raise in it, it just stupidly checks for raises within the scope of function :-( -- ignore the pydoclint error?

src/pymovements/gaze/gaze_dataframe.py
SiQube: pydoclint complains that the function does not have a raise in it, it just stupidly checks for raises within the scope of function :-( -- ignore the pydoclint error?

src/pymovements/gaze/experiment.py
SiQube: I'll double check

src/pymovements/dataset/dataset.py
dkrako: there's no error code, see my latest main comment

src/pymovements/events/properties.py
dkrako: have you identified the error code?

src/pymovements/gaze/screen.py
dkrako: #583 refactored `x_max_dva` etc. into a property method. The docstring here is then probably not needed anymore.

SiQube:
> and actually, if we think that through, pydoclint is technically in the right: if we look at the stack-trace, the exception is not thrown by the documented outer function, but by the inner check function. otherwise we would have to specify a lot of useless exceptions (e.g. basically every function can be aborted by KeyboardException) and we would need to include that in every function.

that's why I deleted them from the documentation -- what is your opinion, should I readd them?

> 
> In the next step we have to find out why the attributes section is failing. Do you have the particular error message from pydoclint?

I'll look into this when we decided to add pydoclint -- it is a lot of work especially with the big refactors (see conflicts above) but I'd be willing to do it :)

dkrako:
> > and actually, if we think that through, pydoclint is technically in the right: if we look at the stack-trace, the exception is not thrown by the documented outer function, but by the inner check function. otherwise we would have to specify a lot of useless exceptions (e.g. basically every function can be aborted by KeyboardException) and we would need to include that in every function.
> 
> that's why I deleted them from the documentation -- what is your opinion, should I readd them?

I think just completely removing this from our documentation is not that good. Did you find a way to ingore this?
The same also with the attributes.

> 
> > In the next step we have to find out why the attributes section is failing. Do you have the particular error message from pydoclint?
> 
> I'll look into this when we decided to add pydoclint -- it is a lot of work especially with the big refactors (see conflicts above) but I'd be willing to do it :)

I don't understand. Should we reevaluate if we will add this? I think it's totally worth it.

But I would advise to put errors that need a lot of work into the list of ignores. We can then tackle them one by one in separate PRs.

Focus on finishing this PR without doing much extra work. Otherwise you will just get new merge conflicts next time.

SiQube:
problem is: 

```python
def a():
    """
    Function just raises stuff.

    Raises
    ------
    Exception
        foo bar
    """
    raise Exception


def b():
    """
    Function does nothing.
    """
    a()


def c():
    """
    Function does nothing.

    Raises
    ------
    Exception
       foo bar
    """
    a()


def d():
    """
    Function just raises stuff.
    """
    raise Exception
```

the problem is now that we get an error, for function `c` we get DOC502: Function `c` has a "Raises" section in the docstring, but there are not "raise" statements in the body

for function `b` we lose the documentation of the exception


there are a few different options

1. ignore all "errors" with raises
pro: no problems if a raise can happen in an inhereting way (no error thrown for function `c`)
con: no error when the unit has no docstring documentation for the raise (no error thrown for function `d`)

2. try except approach as discussed [here](https://github.com/aeye-lab/pymovements/pull/512#issuecomment-1757164028)

```python
try:
   check_something(variable)
except ValueError as e:
    raise e
```

I can't think of any other option

dkrako:
Thank you for the update. Let's keep it simple first with as little code changes as possible, so let's ignore `DOC502` for this PR.

We can then create an issue for resolving the raise problems, as we need to have a more thorough discussion if our exception raising needs an overhaul.

PR #545, opened 2023-09-13
chore: Migrate to polars v0.19

dkrako:
let's see if it breaks

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ dependencies = [
   "matplotlib>=3.0.0,<4",
   "numpy>=1.10.0,<2",
   "pandas>1.0.0,<3",
-  "polars>=0.18.3,<0.20.0",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -35,7 +35,7 @@ dependencies = [
   "matplotlib>=3.0.0,<4",
   "numpy>=1.10.0,<2",
   "pandas>1.0.0,<3",
-  "polars>=0.18.3,<0.19.0",
+  "polars>=0.18.3,<0.20.0",
```
SiQube, line 38: add here `polars>=0.19.0` instead of 0.18.3

SiQube (commented):
thanks for having alles aufm schirm!

pyproject.toml
dkrako: done

PR #549, opened 2023-09-13
build: don't include scope in dependabot commits

dkrako:
## Description

I made an error when writing the config, the scope should not be included in the commit messages


SiQube (approved):

SiQube:
no clue about dependabot -- was so far only annoying. good catch

PR #553, opened 2023-09-14
feat: Add events to GazeDataFrame init

dkrako:
## Description

Add events keyword argument to GazeDataFrame initializer.
For a start I would like to keep this rather simple.

#552 must be merged first

## Implemented changes

- [x] If events is None: create empty EventDataFrame
- [x] Else: create copy of passed EventDataFrame

## Type of change

## How Has This Been Tested?

- [x] Tested attributes after initialization for a few dataframe combinations

## Checklist:

### DIFF
--- a/src/pymovements/gaze/gaze_dataframe.py
+++ b/src/pymovements/gaze/gaze_dataframe.py
@@ -221,7 +221,7 @@ class GazeDataFrame:
         if events is None:
             self.events = pm.EventDataFrame()
         else:
-            self.events = events.clone()
+            self.events = events.copy()
 
     def transform(
             self,
--- a/src/pymovements/gaze/gaze_dataframe.py
+++ b/src/pymovements/gaze/gaze_dataframe.py
@@ -27,7 +27,7 @@ from typing import Any
 
 import polars as pl
 
-import pymovements as pm
+import pymovements as pm  # pylint: disable=cyclic-import
 from pymovements.gaze import transforms
 from pymovements.gaze.experiment import Experiment
 from pymovements.utils import checks
--- a/src/pymovements/gaze/gaze_dataframe.py
+++ b/src/pymovements/gaze/gaze_dataframe.py
@@ -27,6 +27,7 @@ from typing import Any
 
 import polars as pl
 
+import pymovements as pm
 from pymovements.gaze import transforms
 from pymovements.gaze.experiment import Experiment
 from pymovements.utils import checks
@@ -79,6 +80,7 @@ class GazeDataFrame:
             self,
             data: pl.DataFrame | None = None,
             experiment: Experiment | None = None,
+            events: pm.EventDataFrame | None = None,
             *,
             trial_columns: str | list[str] | None = None,
             time_column: str | None = None,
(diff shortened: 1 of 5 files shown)


src/pymovements/gaze/gaze_dataframe.py
```diff
@@ -27,6 +27,7 @@
 
 import polars as pl
 
+import pymovements as pm  # pylint: disable=cyclic-import
```
dkrako, line 30: This PR includes an EventsDataFrame in a GazeDataFrame, which leads to circular dependencies when importing both.

Doing something like
```python
from pymovements.gaze.gaze_dataframe import GazeDataFrame
```
results in a hard python circular import error.

I have found a workaround by importing the pymovements top-level package, but then pylint is not happy with me again.

Right now I just can't find a way to untie this knot. As this is just a linting issue, I would like to create a follow-up issue on how to solve this problem and leave it like this for this PR.

dkrako, line 30: This PR includes an EventsDataFrame in a GazeDataFrame, which leads to circular dependencies when importing both.

Doing something like
```python
from pymovements.gaze.gaze_dataframe import GazeDataFrame
```
results in a hard python circular import error.

I have found a workaround by importing the pymovements top-level package, but then pylint is not happy with me again.

Right now I just can't find a way to untie this knot. As this is just a linting issue, I would like to create a follow-up issue on how to solve this problem and leave it like this for this PR.

SiQube: ok

src/pymovements/gaze/gaze_dataframe.py
```diff
@@ -214,6 +218,11 @@ def __init__(
         self.n_components = _infer_n_components(self.frame, column_specifiers)
         self.experiment = experiment
 
+        if events is None:
+            self.events = pm.EventDataFrame()
+        else:
+            self.events = events.copy()
```
SiQube, line 224: we should have this structure in the same manner as [`self.frame`](https://github.com/aeye-lab/pymovements/blob/01b8c0389107fa21bccf151ed32e3bff21481263/src/pymovements/gaze/gaze_dataframe.py#L183-L187). imo, the `self.events` way is more readable.


btw, I just noticed that we initialize with data and then rename it to frame. I see where we are coming from but it might lead to user confusion

SiQube (commented):
as usual great work --  the circular import is kinda ugly. please open an issue before merging it

src/pymovements/gaze/gaze_dataframe.py
prassepaul: k

prassepaul (approved):

src/pymovements/gaze/gaze_dataframe.py
dkrako: Yes currently using data is a bit confusing, but I also think both terms `frame` leads to similar confusion, because if we pass events, they are data/frame too, right? So what's the difference?

The change I proposed last week also included, that we will slowly migrate into a `Gaze` class, that has `signal` and `events` (and `stimulus` and `meta` and so on) attributes. Let's discuss again and solve the confusion when we start working on that task.

PR #561, opened 2023-09-15
fix matplotlib<3.8 until we fixed typing

SiQube:
resolves #560

PR #563, opened 2023-09-15
chore: Migrate typing to matplotlib v3.8.0

### LINKED ISSUE #562: upgrade matplotlib to 3.8
I found the root cause:

https://github.com/matplotlib/matplotlib/releases

matplotlib 3.8 was released which got rid of typestubs and now uses inherent typing (which is different), I'll fix it asap.

_Originally posted by @SiQube in https://github.com/aeye-lab/pymovements/issues/560#issuecomment-1721280603_
            

have to fix mypy typing for it, see #560

SiQube:
there are still a couple of TODOs, see https://github.com/pydata/xarray/pull/8030, but we can upgrade to `matplotlib==3.8.0`.

I'll leave it up to my fellow code owners @dkrako and @prassepaul to decide which way to merge. either this one or #561. this PR would resolve #562


SiQube:
would resolve #562

src/pymovements/plotting/traceplot.py
```diff
@@ -21,12 +21,17 @@
 from __future__ import annotations
 
 import sys
+from typing import Dict
+from typing import Literal
+from typing import Sequence
+from typing import Tuple
 
-import matplotlib
+import matplotlib.colors
```
SiQube, line 29: this change follows: https://github.com/matplotlib/matplotlib/issues/24950#issuecomment-1379944582

src/pymovements/plotting/traceplot.py
```diff
@@ -37,44 +42,59 @@
 if 'pytest' in sys.modules:  # pragma: no cover
     matplotlib.use('Agg')
 
-DEFAULT_SEGMENTDATA = {
+LinearSegmentedColormapType: TypeAlias = Dict[
```
dkrako, line 45: can you use `dict` instead of `Dict` here?

src/pymovements/plotting/traceplot.py
```diff
@@ -37,44 +42,59 @@
...
+    Literal['red', 'green', 'blue', 'alpha'], Sequence[Tuple[float, ...]],
```
dkrako, line 46: can you use `tuple` instead of `Tuple` here?

dkrako (changes requested):
Thank you so much for taking care of this.

Regarding importing of typehints I have two minor change requests though.

src/pymovements/plotting/traceplot.py
SiQube: no unfortunately only works for 3.9+ and we (still) support 3.8

src/pymovements/plotting/traceplot.py
SiQube: no unfortunately only works for 3.9+ and we (still) support 3.8

src/pymovements/plotting/traceplot.py
dkrako: are you sure? `from __future__ import annotations` should fix that.

we're using this kind of type annotations everywhere else in the package, for example here:

https://github.com/aeye-lab/pymovements/blob/5e0959fb9ed209850733645eab5fc8ae77d31c3f/src/pymovements/synthetic/step_function.py#L28-L34

dkrako: I couldn't resist and had to test it by myself and you're right, it doesn't work.

I just absolutely don't understand why that's the case, because the future annotations solve this issue everywhere else (even in the same file there is `tuple[int, int]`.

dkrako (approved):
Alright, let's merge then

PR #565, opened 2023-09-15
ci: remove mypy ignore imports for matplotlib

SiQube:
needs #563 or #561


prassepaul (approved):

dkrako (approved):
Thank you for checking!

dkrako (changes requested):
mypy raises the errors, as matplotlib is not the recent version there somehow:

```
matplotlib==3.7.3
```
(in line 13 in the error log)

- [ ] I think this PR should come with a requirement of matplotlib>=3.8.0

SiQube:
okay how about we just kill mypy within github actions -- we're checking it via pre-commit.ci anyways. matplotlib 3.8 has dropped python 3.8... otherwise I'd say we stay at 3.7.3 with the already merged fix

dkrako:
> okay how about we just kill mypy within github actions -- we're checking it via pre-commit.ci anyways. matplotlib 3.8 has dropped python 3.8... otherwise I'd say we stay at 3.7.3 with the already merged fix

The current numpy release is also python >=3.9: https://numpy.org/neps/nep-0029-deprecation_policy.html

I'm actually not that hesitant about dropping python 3.8 altogether (but then in a separate PR).

SiQube:
requires #572

ritchie46:
> fails due to polars not building wheels for windows, see [here](https://github.com/pola-rs/polars/issues/11355), and [here](https://github.com/pola-rs/polars/issues/11356).

Should be fixed now.

dkrako (approved):

PR #585, opened 2023-10-04
test: Add functional tests for preprocessing gaze files

### LINKED ISSUE #492: write fixtures for test versions of our public datasets
## Description of the problem

regardless of our 100% coverage, the bug in #483 happened because we don't download the dataset resources of the public datasets (doing this for every test run would be a total waste of resources).

I fixed it without writing tests because I didn't have that much time&energy and fixing the bug had high priority.

## Description of the solution

Instead I should have written a small toy version of all the datasets which we provide, with just a few subjects and sessions, but where all dataframes have the exact column names and types of the respective datasets (values can be zero as in the other dataset test fixtures).

dkrako:
resolves #492


prassepaul (approved):
good work

PR #586, opened 2023-10-04
tests: Reuse dataset definitions in from_csv tests

dkrako:
Reusing the data from the public DatasetDefinitions should be more robust for future changes.


prassepaul (approved):
really good idea. thx

Issue #517, opened 2023-09-09
pos2vel() broken with GazeBase


dkrako:
## Current Behavior
There's a type error in a polars computation.
Maybe there's an operation that gets accidentally a column string?
Or the nesting didn't work properly?
Or there's something strange with the default values?

```
(base) user3@ift-dgx:~/lab/pymovements                                                                                                                                                                          $ cat t.py
import pymovements as pm                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           dataset = pm.Dataset('GazeBase', path='data/GazeBase')                                                                                                                                                           dataset.download()                                                                                                                                                                                               dataset.load()                                                                                                                                                                                                   dataset.pos2vel()

(base) user3@ift-dgx:~/lab/pymovements                                                                                                                                                                          $ pip install --user --upgrade pymovements                                                                                                                                                                       Looking in indexes: https://pypi.org/simple, https://packagecloud.io/github/git-lfs/pypi/simple                                                                                                                  Requirement already satisfied: pymovements in /home/user3/.local/lib/python3.9/site-packages (0.16.1.post0+git.47e29b71.dirty) 

(base) user3@ift-dgx:~/lab/pymovements
$ python t.py
Downloading https://figshare.com/ndownloader/files/27039812 to data/GazeBase/downloads/GazeBase_v2_0.zip
GazeBase_v2_0.zip: 100%|█████████████████████████████████████████████| 6.25G/6.25G [03:20<00:00, 33.4MB/s]
Checking integrity of GazeBase_v2_0.zip
Extracting GazeBase_v2_0.zip to data/GazeBase/raw
100%|███████████████████████████████████████████████████████████████| 12334/12334 [08:12<00:00, 25.02it/s]  
0%|                                                                  | 22/12334 [00:01<12:45, 16.09it/s]
Traceback (most recent call last):
  File "/home/user3/lab/pymovements/t.py", line 7, in <module>
    dataset.pos2vel()
  File "/home/user3/lab/pymovements/src/pymovements/dataset/dataset.py", line 355, in pos2vel
    gaze_df.pos2vel(method=method, **kwargs)
  File "/home/user3/lab/pymovements/src/pymovements/gaze/gaze_dataframe.py", line 373, in pos2vel
    self.transform('pos2vel', method=method, **kwargs)
  File "/home/user3/lab/pymovements/src/pymovements/gaze/gaze_dataframe.py", line 298, in transform
    [
[...]

PR #589, opened 2023-10-05
tests: Refactor tests into  unit and functional

dkrako:
unit tests and functional tests should be clearly separated. unit tests test functionality of single specific parts of the software. functional tests test use cases that consist of several steps with defined outcomes.

Next step will be creating a directory named `integration` for integration tests. integration tests test the functionality in regard to 3rd party systems. these can be databases or servers, in our case we will start out with the usecase of downloading public datasets and doing basic preprocessing. This should then finally reproduce #517


prassepaul (approved):

SiQube:
Hi before this get merged I wanted to write my two cents. Since we're refactoring we should move all the logic that we have in the tests into a `testing` repository and test it as well. It might as well be that we have bugs in all the logic we introduce within each of the tests scripts. This might be a different PR but better practice is to remove any logic from tests.

dkrako:
> Hi before this get merged I wanted to write my two cents. Since we're refactoring we should move all the logic that we have in the tests into a `testing` repository and test it as well. It might as well be that we have bugs in all the logic we introduce within each of the tests scripts. This might be a different PR but better practice is to remove any logic from tests.

Yes, this makes sense, and can then be put into `tests/testing` right? I think most of the logic is very straight forward, but maybe we will create some fixtures in the future that we want to reuse across tests or so? but that's something for the future

SiQube:
I'd say it should be it's own directory, since we exclude tests from coverage, could be `testing` but in root.

dkrako:
can you identify what particular logic of our tests you would like to have tests for and create an issue with this?

I see that for stuff like `polars.testing.assert_frame_equal()` this makes sense, but I can't really identify such logic in pymovements yet.

Issue #508, opened 2023-09-07
load gaze files with pm.gaze.from_asc()


dkrako:
## Description of the problem
Currently we need the Dataset class to load GazeDataFrames.

It would be much nicer to have a function like `pymovements.gaze.from_asc()` which returns a GazeDataFrame loaded from a asc file.

## Description of a solution
Most functionality is included in both `dataset/dataset_files.py` and `utils/parsing.py`.
`utils.parsing.parse_eyelink()` should be used in `gaze.from_asc()`, logic in `dataset/dataset_files.py` should be refactored into `gaze/files.py` and reused in `dataset_files.py`.

## Minimum acceptance criteria
- [ ] `pm.gaze.from_asc()` in `gaze/files.py`
- [ ] use `pm.gaze.from_asc()` in `dataset/dataset_files.py`

PR #590, opened 2023-10-05
feat: Add gaze.from_asc()

dkrako:
## Description

Implement from_eyelink_asc() which returns a GazeDataFrame read from an eyelink asc file.

Fixes issue #508 

## Type of change

## How Has This Been Tested?

- [x] Created test asc file and check GazeDataFrame for correct shape

## Checklist:


tests/unit/gaze/io/eyelink_asc_test.py
```diff
@@ -0,0 +1,41 @@
...
+            },
+            (16, 3),
+            id='eyelink_asc_mono',
+        ),
+    ],
+)
+def test_shapes(kwargs, shape):
+    gaze_dataframe = pm.gaze.from_eyelink_asc(**kwargs)
+
+    assert gaze_dataframe.frame.shape == shape
```
SiQube, line 41: We're only testing shape, should we test the values extracted?

SiQube (commented):
Great work, only one comment regarding testing. We might want to test for values not only shape.

tests/unit/gaze/io/eyelink_asc_test.py
dkrako: hm I would need to copy-paste them then one by one, what about checking the correct schema and length?

dkrako: but I can do both actually, then we're on the safe side

dkrako: done

prassepaul (approved):
really good.

Issue #597, opened 2023-10-06
Create new EyeTracker class and use it in Experiment


dkrako:
## Description of the problem

When parsing eye tracker related metadata in #510, we will need to have a place for keeping the data.

## Description of a solution

The experiment class would be perfect for holding that information, as it already includes the sampling rate (which can also be parsed from eyelink files).

The best solution would include creating a new class `EyeTracker(sampling_rate, eyes, model, version, mount)`.

The sampling rate would then just be forwarded as `self.sampling_rate = self.eye_tracker.sampling_rate`.

## Minimum acceptance criteria
- [x] Create new class EyeTracker
- [x] Use class in Experiment
- [x] create property `Experiment.sampling_rate` for short-access to `Experiment.eye_tracker.sampling_rate`

PR #617, opened 2023-10-26
feat: add EyeTracker class

josephine-funken:
## Description

Added an EyeTracker class that can keep parsed eyetracker related data.

Fixes issue #597 

## Implemented changes

- [x] Create class EyeTracker
- [x] Use class in Experiment
- [x] Add short-access to `Experiment.eye_tracker.sampling_rate` from `Experiment`

## Type of change

## How Has This Been Tested?

- [x] Test class `tests.unit.gaze.eyetracker_test.py`

## Checklist:


pyproject.toml
```diff
@@ -135,7 +135,8 @@ filterwarnings = [
   "ignore:pkg_resources is deprecated.*:DeprecationWarning",
   "ignore:Deprecated call to `pkg_resources.declare_namespace.*:DeprecationWarning",
   "ignore:distutils Version classes are deprecated. Use packaging.version instead.:DeprecationWarning",
-  "ignore:FigureCanvasAgg is non-interactive, and thus cannot be shown:UserWarning"
+  "ignore:FigureCanvasAgg is non-interactive, and thus cannot be shown:UserWarning",
+  "ignore:`the argument periods` for `shift` is deprecated. It has been renamed to `n`:DeprecationWarning"
```
SiQube, line 139: #618 fixes this

SiQube:
#618 fixes the docs problems

src/pymovements/gaze/eyetracker.py
```diff
@@ -0,0 +1,99 @@
...
+# SOFTWARE.
+"""This module holds the EyeTracker class."""
+from __future__ import annotations
+
+from pymovements.utils import checks
+from pymovements.utils import decorators
+
+
+@decorators.auto_str
+class EyeTracker:
```
dkrako, line 28: can you use the `@dataclass` decorator here instead?
as this class won't have any methods it would be best to implement this class as a dataclass right from the start.

You won't need to define an `__init__()` method then, but you will need to move the checks to a `__post_init__()` method. The docstring example can be moved to the class docstring.

See this page with info about dataclasses:
https://docs.python.org/3/library/dataclasses.html

And a nice video on why dataclasses are awesome:
https://www.youtube.com/watch?v=CvQ7e6yUtnw

pyproject.toml
dkrako, line 139: this line should be removed again

src/pymovements/gaze/experiment.py
```diff
@@ -109,12 +115,15 @@ def __init__(
             origin=origin,
         )
 
-        checks.check_is_not_none(sampling_rate=sampling_rate)
-        assert sampling_rate is not None
+        if sampling_rate is not None:
+            self.sampling_rate = sampling_rate
+        if eyetracker is not None:
+            self.sampling_rate = eyetracker.sampling_rate
```
dkrako, line 121: I think we should raise an error that both `eyetracker` and `sampling_rate` is passed.
You can use `checks.check_is_mutual_exclusive()` for that

dkrako (changes requested):
Thank you so much for working an this feature.

I'm totally confused where the current error in the tutorial notebook is coming from.
I'll help you with the error, so you can ignore it in the meantime

src/pymovements/gaze/eyetracker.py
josephine-funken: I've made this change

pyproject.toml
josephine-funken: Done

src/pymovements/gaze/experiment.py
josephine-funken: Done

src/pymovements/gaze/experiment.py
```diff
@@ -109,12 +115,17 @@ def __init__(
             origin=origin,
         )
 
-        checks.check_is_not_none(sampling_rate=sampling_rate)
-        assert sampling_rate is not None
+        checks.check_is_mutual_exclusive(sampling_rate=sampling_rate, eyetracker=eyetracker)
+
+        if sampling_rate is not None:
```
dkrako, line 120: I reconsidered a bit how we should implement this short-access. I worry that this is a bit too unsafe, because people could do something strange like updating the sampling rate which would introduce an inconsistency:
```python
experiment = Experiment(..., experiment=EyeTracker(1000))
experiment.eyetracker.sampling_rate = 12345.0
print(experiment.sampling_rate)
1000
```

I think it would be best to replace both if-clauses just with this:
```python
self._sampling_rate = sampling_rate
```
and then create a property for the sampling_rate that chooses the correct one:
```python
@property
def sampling_rate(self): 
    if self._sampling_rate is not None:
        return self._sampling_rate

    return self.eyetracker.sampling_rate
```

By storing the explicitly passed `sampling_rate` in a private attribute we can infer the value dynamically.

If you want, you can also include a `@sampling_rate.setter` method to set the sampling rate, but on assignment we should update `eyetracker.sampling_rate` too.
It's also ok to have this attribute read-only. Do as you wish on that case.

src/pymovements/gaze/eyetracker.py
```diff
@@ -0,0 +1,74 @@
...
+    left : bool
+        Whether the left eye is tracked
+    right : bool
+        Whether the right eye is tracked
+    model : str
+        EyeLink tracker model (e.g. 'EyeLink II', 'EyeLink 1000')
+    version : str
+        EyeLink software version number
+    mount : str
+        The mounting setup of the EyeLink (e.g. 'Desktop / Monocular / Remote')
```
dkrako, line 45: here, too, just eye tracker

src/pymovements/gaze/eyetracker.py
```diff
@@ -0,0 +1,74 @@
...
+        ...     version = '1.5.3',
+        ...     mount = 'Arm Mount / Monocular / Remote',
+        ... )
+        >>> print(eyetracker)
+        EyeTracker(sampling_rate=1000.0, left=False, right=True, model='EyeLink 1000 Plus',
+        version='1.5.3', mount='Arm Mount / Monocular / Remote')
+    """
+
+    sampling_rate: float
+    left: bool
```
dkrako, line 63: Apart from `sampling_rate` I think all of the other attributes should be optional with a default value of `None`

src/pymovements/gaze/eyetracker.py
```diff
@@ -0,0 +1,74 @@
...
+    ----------
+    sampling_rate : float
+        Sample rate of recording (in Hz)
+    left : bool
+        Whether the left eye is tracked
+    right : bool
+        Whether the right eye is tracked
+    model : str
+        EyeLink tracker model (e.g. 'EyeLink II', 'EyeLink 1000')
+    version : str
```
dkrako, line 42: I just thought that adding an additional attribute `vendor` would be really helpful when we will start adding support for other eye tracker vendors. can you add it between `version` and `mount`?

dkrako (changes requested):
This already looks great! I still have some minor change requests, but after incorporating these we are good to go.

src/pymovements/gaze/experiment.py
dkrako: I just noticed that implementing it this way will remove the sampling rate from the `__str__` string generated by the `@auto_str` decorator. So when working on this, you need to implement a short `__str__` method of the `Experiment` class. 

https://github.com/aeye-lab/pymovements/blob/f9c43ddfbd0212edfd409e8a3ad94844af5023ec/src/pymovements/utils/decorators.py#L36-L39

I'd propose this a simple implementation:

```python
    def __str__(self: Any) -> str:
        attributes = ', '.join(f'{key}={shorten(value)}' for key, value in vars(self).items())
        return f'{type(self).__name__}(sampling_rate={self.sampling_rate}, {attributes})'
```

The `@auto_str` decorator must be removed from the class then.

src/pymovements/gaze/experiment.py
```diff
@@ -109,12 +120,27 @@ def __init__(
...
+    @property
+    def sampling_rate(self) -> float:
+        """Assign sampling rate to experiment."""
+        if self._sampling_rate is not None:
+            return self._sampling_rate
+
+        if self.eyetracker is not None:
+            return self.eyetracker.sampling_rate
+
+        raise TypeError("'self._sampling_rate' and 'self.eyetracker' must not both be None")
```
dkrako, line 143: I think this line can not be reached, as we already check in the constructor that the sampling rate is not None.

So you can just remove the raise and the last if-condition, and just `return self.eyetracker.sampling_rate`

dkrako (changes requested):
One tiny change left and we're ready to go!

src/pymovements/gaze/experiment.py
josephine-funken: After removing the raise, mypy gave me the following error:
src/pymovements/gaze/experiment.py:140: error: Item "None" of "EyeTracker | None" has no attribute "sampling_rate"  [union-attr]
So, I added `# type: ignore [union-attr]` in line 140. Is this a viable approach?

dkrako: change the signature to `def sampling_rate(self) -> float | None:`

src/pymovements/gaze/experiment.py
```diff
@@ -20,24 +20,27 @@
...
-@decorators.auto_str
 class Experiment:
     """Experiment class for holding experiment properties.
 
     Attributes
     ----------
     screen : Screen
         Screen object for experiment
-    sampling_rate : float
+    _sampling_rate : float
```
dkrako, line 40: this is a private attribute, we don't need to document it

src/pymovements/gaze/experiment.py
```diff
@@ -109,12 +120,24 @@ def __init__(
...
+        checks.check_is_not_none(sampling_rate=self.sampling_rate)
+        assert self.sampling_rate is not None
 
-        checks.check_is_greater_than_zero(sampling_rate=sampling_rate)
+        checks.check_is_greater_than_zero(sampling_rate=self.sampling_rate)
 
-        self.sampling_rate = sampling_rate
+    @property
+    def sampling_rate(self) -> float:
+        """Assign sampling rate to experiment."""
```
dkrako, line 136: This is just a read only property (we don't need a setting method here but it would need the decorator `@sampling_rate.setter` and the value as the parameter). Change the docstring to `Get sampling rate of experiment.`

src/pymovements/gaze/eyetracker.py
```diff
@@ -0,0 +1,78 @@
...
+        ...     version = '1.5.3',
+        ...     vendor = 'EyeLink',
+        ...     mount = 'Arm Mount / Monocular / Remote',
+        ... )
+        >>> print(eyetracker)
+        EyeTracker(sampling_rate=1000.0, left=False, right=True, model='EyeLink 1000 Plus',
+        version='1.5.3', vendor='EyeLink', mount='Arm Mount / Monocular / Remote')
+    """
+
+    sampling_rate: float
```
dkrako, line 65: sampling rate should be optional. change to `sampling_rate: float | None`

src/pymovements/gaze/eyetracker.py
```diff
@@ -0,0 +1,78 @@
...
+    left: bool | None = None
+    right: bool | None = None
+    model: str | None = None
+    version: str | None = None
+    vendor: str | None = None
+    mount: str | None = None
+
+    def __post_init__(self) -> None:
+        """Check that the sampling rate is a positive number."""
+        checks.check_is_not_none(sampling_rate=self.sampling_rate)
```
dkrako, line 75: sampling rate is optional. as the greater then zero check needs a scalar sampling rate, use an if clause:

```python
if self.sampling_rate is None:
        checks.check_is_greater_than_zero(sampling_rate=self.sampling_rate)
```

tests/unit/gaze/eyetracker_test.py
```diff
@@ -0,0 +1,55 @@
...
+    EyeTracker(
+        1000.0, False, True, 'EyeLink 1000 Plus',
+        '1.5.3', 'Arm Mount / Monocular / Remote',
+    )
+
+
+@pytest.mark.parametrize(
+    ('sampling_rate', 'exception', 'exception_msg'),
+    [
+        pytest.param(
```
dkrako, line 36: this case can be probably deleted again

src/pymovements/gaze/eyetracker.py
```diff
@@ -0,0 +1,78 @@
...
+
+@dataclass
+class EyeTracker:
+    """EyeTracker class for holding eyetracker properties.
+
+    Attributes
+    ----------
+    sampling_rate : float
+        Sample rate of recording (in Hz)
+    left : bool
```
dkrako, line 36: as all parameters are optional, add `| None` to all the documented types in this docstring

dkrako (changes requested):
Alright, we're getting to it! Minor change requests regarding optional parameters.

Thank you for your endurance!

src/pymovements/datasets/toy_dataset_eyelink.py
```diff
@@ -115,7 +116,13 @@ class ToyDatasetEyeLink(DatasetDefinition):
         screen_height_cm=30.2,
         distance_cm=68,
         origin='lower left',
-        sampling_rate=1000,
+        eyetracker=EyeTracker(
+            sampling_rate=1000.0,
+            left=True,
```
dkrako, line 121: The ToyDataset is only a monocular recording, if I remember it is recorded with the right eye.
Please ask Jakob for confirmation as he did recording.

Here is one example file from the dataset: https://github.com/aeye-lab/pymovements-toy-dataset/blob/main/data/trial_0_1.csv

src/pymovements/gaze/experiment.py
```diff
@@ -64,8 +65,8 @@ class Experiment:
     ...     sampling_rate=1000.0,
     ... )
     >>> print(experiment)
-    Experiment(screen=Screen(width_px=1280, height_px=1024, width_cm=38,
-    height_cm=30, distance_cm=68, origin=lower left), sampling_rate=1000.00)
+    Experiment(sampling_rate=1000.00, screen=Screen(width_px=1280, height_px=1024, width_cm=38,
+    height_cm=30, distance_cm=68, origin=lower left), eyetracker=None, _sampling_rate=1000.00)
```
dkrako, line 69: Please leave out all attributes with a leading underscore from the `__str__()` method

src/pymovements/gaze/experiment.py
```diff
@@ -155,6 +231,7 @@ def pos2vel(
         ...     distance_cm=68,
         ...     origin='lower left',
         ...     sampling_rate=1000.0,
+        ...     eyetracker=None,
```
dkrako, line 234: Please leave out this line from the example, as this is already None by default.

But if you like, you can add another example to the docstring which doesn't pass any `sampling_rate` but an `EyeTracker` instance. In that case use simple one line descriptions like:

- initialize experiment by passing sampling_rate
- initialize experiment by passing an eye tracker instance

src/pymovements/gaze/experiment.py
```diff
@@ -107,12 +166,29 @@ def __init__(
...
+    @property
+    def sampling_rate(self) -> float | None:
+        """Get sampling rate of experiment."""
+        if self._sampling_rate is not None:
+            return self._sampling_rate
+
+        return self.eyetracker.sampling_rate
+
+    @sampling_rate.setter
+    def sampling_rate(self, _sampling_rate: float | None = None) -> None:
```
dkrako, line 189: Please rename the argument here to `sampling_rate` without the leading underscore.

We will only use the leading underscore for marking attributes as private.

src/pymovements/gaze/experiment.py
```diff
@@ -171,3 +248,14 @@ def pos2vel(
...
+
+    def __str__(self: Any) -> str:
+        """Print experiment."""
+
+        def shorten(value: Any) -> str:
+            if isinstance(value, float):
+                value = f'{value:.2f}'
+            return value
+
+        attributes = ', '.join(f'{key}={shorten(value)}' for key, value in vars(self).items())
```
dkrako, line 260: this will probably need something like `if not key.startswith('_')` to not list any private attributes with a leading underscore

src/pymovements/gaze/eyetracker.py
```diff
@@ -0,0 +1,79 @@
...
+    right: bool | None = None
+    model: str | None = None
+    version: str | None = None
+    vendor: str | None = None
+    mount: str | None = None
+
+    def __post_init__(self) -> None:
+        """Check that the sampling rate is a positive number."""
+        checks.check_is_not_none(sampling_rate=self.sampling_rate)
+        assert self.sampling_rate is not None
```
dkrako, line 76: please remove the first two lines here

src/pymovements/gaze/eyetracker.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+@dataclass
+class EyeTracker:
+    """EyeTracker class for holding eyetracker properties.
+
+    Attributes
+    ----------
+    sampling_rate : float | None
+        Sample rate of recording (in Hz)
```
dkrako, line 35: Please add a period after each sentence and add ` (default: None)` at the end of all the attribute descriptions.

src/pymovements/gaze/experiment.py
```diff
@@ -97,7 +98,65 @@ def __init__(
             distance_cm: float | None = None,
             origin: str = 'lower left',
             sampling_rate: float | None = None,
+            eyetracker: EyeTracker | None = None,
     ):
+        """Initialize Experiment.
```
dkrako, line 103: Starting #638 we don't allow for `__init__()` method docstrings anymore.
Please remove this docstring entirely and merge it into the docstring above (most of the text has been moved there already).
The sentence `Initialize Experiment.` wont' be necessary anymore.

src/pymovements/gaze/experiment.py
```diff
@@ -97,7 +98,65 @@ def __init__(
...
+        distance_cm : float | None
+            Eye-to-screen distance in centimeters. If None, a `distance_column` must be provided
+            in the `DatasetDefinition` or `GazeDataFrame`, which contains the eye-to-screen
+            distance for each sample in millimeters.
+        origin : str
+            Specifies the screen location of the origin of the pixel coordinate system.
+        sampling_rate : float | None
+            Sampling rate in Hz
+        eyetracker : EyeTracker | None
+            EyeTracker object for experiment
```
dkrako, line 124: Please add `(default: None)` at the end of the description

src/pymovements/gaze/experiment.py
```diff
@@ -107,12 +166,29 @@ def __init__(
...
+        """Get sampling rate of experiment."""
+        if self._sampling_rate is not None:
+            return self._sampling_rate
+
+        return self.eyetracker.sampling_rate
+
+    @sampling_rate.setter
+    def sampling_rate(self, _sampling_rate: float | None = None) -> None:
+        """Set sampling rate of experiment."""
+        self._sampling_rate = _sampling_rate
```
dkrako, line 191: a simple test like this should be sufficient to cover this line:

```python
experiment = Experiment(1280, 1024, 38, 30, sampling_rate=1000.0)
assert sampling_rate == 1000.0
experiment.sampling_rate = 100.0
assert sampling_rate == 100.0
```

dkrako (changes requested):
Thank you for your work! As discussed in the meeting, here are my minor change requests.

dkrako (approved):
Great! Nice work, thanks a lot for your endurance!

PR #618, opened 2023-10-28
chore: migrate to polars v0.19.12

SiQube:
recent update of `polars` added deprecations to both `prefix` and `shift`, which are fixed here. additionally, `shift` is not working with `np.int{32,64}` -- return type of np.ceil -- hence the change from `.astype(int)` to `int()`.


SiQube:
I've bumped the minimum requirement for `polars` to prevent a depreciation that will come (soon)

SiQube:
on top docs had a warning that needed attending, namely `'navigation_with_keys'` which I set to the recommended `False`.

see here:

```console
WARNING: The default value for `navigation_with_keys` will change to `False` in the next release.
If you wish to preserve the old behavior for your site, set `navigation_with_keys=True` in the `html_theme_options` dict in your `conf.py` file.
Be aware that `navigation_with_keys = True` has negative accessibility implications:https://github.com/pydata/pydata-sphinx-theme/issues/1492
```

SiQube:
this patch should fix our persisting problem in all other PRs

src/pymovements/gaze/integration.py
```diff
@@ -234,26 +234,26 @@ def from_numpy(
...
     if pixel is not None:
-        df = pl.from_numpy(data=pixel, orient=orient).select(pl.all().prefix('pixel_'))
+        df = pl.from_numpy(data=pixel, orient=orient).select(pl.all().name.prefix('pixel_'))
         dfs.append(df)
         pixel_columns = df.columns
 
     position_columns = None
     if position is not None:
-        df = pl.from_numpy(data=position, orient=orient).select(pl.all().prefix('position_'))
+        df = pl.from_numpy(data=position, orient=orient).select(pl.all().name.prefix('osition_'))
```
SiQube, line 243: I've had this mistake while renaming -- why isn't this tested?

dkrako (approved):
Thank you so much for resolving the issue!

src/pymovements/gaze/integration.py
dkrako: this is actually hard to test because the column names are just internal (i.e. we would need to mock).
at the end of the function we initialize a GazeDataFrame with `position_columns=position_columns`, which has the effect that the columns are nested into `position` and the original columns are removed, i.e. the original column names don't matter much.

Issue #629, opened 2023-11-21
Missing gaze.io and gaze.integration Functions in API Documentation


jakobchwastek:
The functions in `gaze.io` and `gaze.integration` are not listed in the API Reference section of the pymovements documentation on [pymovements.readthedocs.io](pymovements.readthedocs.io).

PR #630, opened 2023-11-22
docs: add 'Integration and IO' rubric to api reference docs

jakobchwastek:
fix issue #629


dkrako (changes requested):
Thank you very much for your proactiveness! Some minor change requests.

- [ ] It think the public interface should rather be

```python
pymovements.gaze.from_numpy()
pymovements.gaze.from_asc()
```

- [ ] Separate `Integration` and `IO` into distinct subsections

- [ ] `IO` could then rather be named to `Input / Output`. We will put `GazeDataFrame.save()` there someday.

- [ ] move the integration section to the end

dkrako (approved):
I reordered and used the shorter namespaces. Thanks again for taking care!

Issue #626, opened 2023-11-17
Timestamp Unit Specification


jakobchwastek:
## Description of the problem
Currently `GazeDataFrame` and `DatabaseDefinition` lack explicit information about the unit of timestamps in the time column, which is crucial for some future functions. 

## Description of a solution
- Add a `time_column_unit` parameter to the constructor of `GazeDataFrame` defining the unit of `time_column`.
- Implement a check in `GazeDataFrame` to verify if the unit has been set when a time column is specified.
- Add `time_column_unit` attribute in the `DatasetDefinition` class.
- Expand functions  `gaze.from_csv`, `gaze.from_pandas`, `gaze.from_numpy` to include a `time_column_unit` parameter.

## Minmum Acceptance Criteria
- Successful implementation of the proposed solution.
- Add or adjust unit and integrations tests.
- Update Documentation

jakobchwastek:
Would it be beneficial to standardize all timestamps to milliseconds during the initialization of `GazeDataFrame`? This approach could simplify the implementation and testing of functions that handle timestamps by limiting them to a single unit.
In that case `time_column_unit` would only reflect the unit of timestamps in input data frame.

dkrako:
Yes I think that's the way to go! That way we have a standardized unit for all our subsequent transformations and only need to check in the init

dkrako:
How do we want to specify the unit? as a string, e.g. `ms`, `sec`?

jakobchwastek:
Yes, i would go with string. Do you think supporting seconds and milliseconds is sufficient for now?

jakobchwastek:
Should we call the parameter `time_column_unit` or just `time_unit`

dkrako:
> Yes, i would go with string. Do you think supporting seconds and milliseconds is sufficient for now?

As a fallback we would probably also need to have a `step` unit. The specific time values in milliseconds would then be calculated by using the sampling rate.

So all in all we should support: `seconds`, `milliseconds`, `step`

dkrako:
> Should we call the parameter `time_column_unit` or just `time_unit`

I like simply `time_unit`. We just need to document that all time values will be transformed to milliseconds.

PR #643, opened 2023-12-13
feat: timestamp unit specification

jakobchwastek:
## Description

Fixes issue #626 

## Implemented changes

- Introduced support for seconds, milliseconds and steps as units of timestamps in the `time_column`

- Added a `time_unit` parameter to the constructor of GazeDataFrame
- Add a `time_unit` attribute in the DatasetDefinition class
- Developed a private function in GazeDataFrame to convert input timestamps to milliseconds.- Expand functions  `gaze.from_csv`, `gaze.from_pandas`, `gaze.from_numpy` to include a `time_unit` parameter
- Incorporated `time_unit` into existing local dataset definitions
- Updated tests and documentation examples to reflect these changes.

A `time_unit` specification is mandatory when a `time_column` is included in the data. To  ensure compatibility with certain event detection methods (e.g. idt) I choosed to cast time values to integer after conversion to millis when possible. For example, float timestamps in seconds with millisecond precision (e.g., `[0.001, 0.002]`) are converted to integer milliseconds (`[1, 2]`).

## Type of change

## How Has This Been Tested?

- Enhanced existing tests: Updated tests that utilize time values in `GazeDataFrame` to include the specification of the `time_unit`.
- Added conversion tests to the GazeDataFrame init tests.

## Checklist:


src/pymovements/dataset/dataset_definition.py
```diff
@@ -133,6 +140,7 @@ class DatasetDefinition:
 
     trial_columns: list[str] | None = None
     time_column: str | None = None
+    time_unit: str | None = None
```
dkrako, line 143: I think it makes sense to have 'ms' as default.
Most of the datasets i encountered have ms timestamps so it would reduce a bit of boilerplate there

src/pymovements/gaze/gaze_dataframe.py
```diff
@@ -57,7 +57,12 @@ class GazeDataFrame:
     time_column: str | None
         The name of the timestamp column in the input data frame. This column will be renamed to
         ``time``. (default: None)
-    pixel_columns: list[str] | None
+    time_unit: str | None
+        The unit of the timestamps in the timestamp column in the input data frame. Supported
+        units are 's' for seconds, 'ms' for milliseconds and 'step' for steps. If the unit is
+        'step' the experiment definition must be specified. All timestamps will be converted to
```
dkrako, line 63: It think we should be more specific here, we need the sampling rate

src/pymovements/gaze/gaze_dataframe.py
```diff
@@ -895,3 +902,41 @@ def _fill_event_detection_kwargs(
...
+
+        if time_unit == 's':
+            self.frame = self.frame.with_columns(pl.col('time').mul(1000))
+
+        elif time_unit == 'step':
+            if self.experiment is None:
+                raise ValueError("experiment must be specified if time_unit is 'step'")
+
+            self.frame = self.frame.with_columns(
+                pl.col('time').mul(1000).truediv(self.experiment.sampling_rate),
```
dkrako, line 923: Sampling rate can be none, add a test for this

src/pymovements/gaze/io.py
```diff
@@ -36,6 +36,7 @@ def from_csv(
         *,
         trial_columns: list[str] | None = None,
         time_column: str | None = None,
+        time_unit: str | None = None,
```
dkrako, line 39: Here too i think it makes sense to have ms as a default

src/pymovements/gaze/integration.py
```diff
@@ -46,6 +46,7 @@ def from_numpy(
         schema: list[str] | None = None,
         orient: Literal['col', 'row'] = 'col',
         time_column: str | None = None,
+        time_unit: str | None = None,
```
dkrako, line 49: Here i think it makes sense to have ms as a default

dkrako (changes requested):
This is great! The only issue I have is 'ms' not being the default

Maybe also discuss with @prassepaul and @SiQube

src/pymovements/gaze/gaze_dataframe.py
```diff
@@ -239,23 +245,28 @@ def __init__(
...
+                    time=pl.arange(0, len(self.frame)),
+                )
 
-            self.frame = self.frame.with_columns(
-                time=pl.arange(0, len(self.frame)) * sampling_rate_factor,
-            )
+                time_column = 'time'
+                time_unit = 'step'
+            else:
+                warnings.warn(
```
jakobchwastek, line 262: if no time column is specified and it is not possible to generate timestamps based on sampling rate there is a warning. Now all tests not using a time column in GazeDataFrame are failing. Can you give me an advise on how to handle this?

dkrako: oh what a bummer.. do you know if it affects more than `gaze_transform_test.py`?

I think the only real solution is adding an experiment definition to test cases where we don't want any warnings.

In the `gaze_transform_test.py` this definition here is used a lot due to simple trigonometrics:

```python
 pm.Experiment(
    screen_width_px=100,
    screen_height_px=100,
    screen_width_cm=100,
    screen_height_cm=100,
    distance_cm=100,
    origin='center',
    sampling_rate=1000,
)
```

Maybe you could create a variable at the top of the file named `experiment_100` (there's an `experiment` fixture already, so don't use that name) and reuse the definition in all the test cases in that file.

Sorry for that annoying work.

dkrako: you will probably need `experiment_100_center?` and `experiment_100_lower_left` though.

jakobchwastek: Thanks for your response, i'll look into it!

jakobchwastek: Most tests `gaze_init_test.py` using a gaze dataframe without time column fail too. I don't really think just adding a time column or experiment in every test where it is missing is a good solution as the time has no effect on the data that is being tested. So i think it is just blowing up the test cases. 

And for example here we have a test that is a legit case but would trigger a warning:
`        pytest.param(
            {
                'data': pl.from_dict(
                    {'x': [1.23, 2.34, 3.45], 'y': [4.56, 5.67, 6.78]},
                    schema={'x': pl.Float64, 'y': pl.Float64},
                ),
                'position_columns': ['x', 'y'],
            },
            pl.from_dict(
                {'position': [[1.23, 4.56], [2.34, 5.67], [3.45, 6.78]]},
                schema={'position': pl.List(pl.Float64)},
            ),
            2,
            id='df_three_rows_two_position_columns_no_time_no_experiment',
`


Couldn't we do something like:

`
def test_init_gaze_dataframe_has_expected_attrs(init_kwargs, expected_frame, expected_n_components):
    with warnings.catch_warnings():
        warnings.simplefilter('ignore', UserWarning)

        gaze = pm.GazeDataFrame(**init_kwargs)
        assert_frame_equal(gaze.frame, expected_frame)
        assert gaze.n_components == expected_n_components
`

dkrako: I see, filtering the warnings is probably the best approach then.

Instead of using warnings.simplefilter, please use [warnings.filterwarnings](https://docs.python.org/3/library/warnings.html#warnings.filterwarnings) and set the expected message and module to not ignore more warnings than necessary.

src/pymovements/gaze/gaze_dataframe.py
jakobchwastek: Actually sampling rate can not be None :)

https://github.com/aeye-lab/pymovements/blob/6bfb14d028a5fb9f0e331c4f6e006523890e3d63/src/pymovements/gaze/experiment.py#L121

dkrako (approved):
Thank you so much for putting all this effort into this!

PR #645, opened 2023-12-19
test: fix expected values in parsing

dkrako:


SiQube (approved):
weird bug -- thanks for fixing!

PR #660, opened 2024-01-11
build: add support for python 3.12

dkrako:
new year new try


dkrako:
I don't get it, what's wrong with pandas?

also it seems the ignore for the DeprecationWarning didn't work

I give up for now, see you next month

dkrako:
alright, the pandas issue seems to be resolved, there's one *DeprecationWarning* left:

```
DeprecationWarning: Python 3.14 will, by default, filter extracted tar archives and reject files or modify their metadata. Use the filter argument to control this behavior.
```

Which is raised by this extractall:

https://github.com/aeye-lab/pymovements/blob/07f8c9793478c260cda3a3f3ad51c5d7110e2014/src/pymovements/utils/archives.py#L152-L153

Heres some supplementary documentation:
https://docs.python.org/3/library/tarfile.html
https://docs.python.org/3/library/tarfile.html#tarfile-extraction-filter

Basically, Python <3.12 had unsecure features for extracting tarfiles activated by default which will be changed in 3.14.

There are two options we have:

1. check if we run python 3.12 (or check if tarfile.extractall() has a `filter` keyword argument) and pass `filter='data'` in case
2. add an ignore for the deprecation warning and leave everything else as is.

How do we want to proceeed?

SiQube:
@prassepaul I think you should chime in since @dkrako and I worked on the PR

src/pymovements/utils/archives.py
```diff
@@ -150,7 +151,10 @@ def _extract_tar(
         Compression filename suffix.
     """
     with tarfile.open(source_path, f'r:{compression[1:]}' if compression else 'r') as archive:
-        archive.extractall(destination_path)
+        if sys.version_info < (3, 12):  # pragma: <3.12 cover
+            archive.extractall(destination_path)
+        else:  # pragma: >=3.12 cover
```
dkrako, line 156: great that this pragma directive works. can you point me to a reference to get more information on this?

dkrako (commented):
I can't approve unfortunately as I'm co-author of this PR

SiQube (approved):

Issue #510, opened 2023-09-07
parse metadata in utils.parsing.parse_eyelink


dkrako:
## Description of the problem
The header of an asc file has a lot of metadata which we can parse.

## Description of a solution
Instead of just returning a dataframe, the function then should return a tuple of a dataframe and a dictionary with parsed info

## Minimum acceptance criteria
- [ ] parsing recording datetime
- [ ] parse eye tracker model and version
- [ ] parse tracker mount info
- [ ] parse display resolution
- [ ] get pupil size data type
- [ ] parse sampling rate
- [ ] parse filter
- [ ] parse `TRACKING`
- [ ] parse `CR`
- [ ] parse `VEL`
- [ ] parse `RES`
- [ ] parse `HTARG`
- [ ] parse `INPUT`
- [ ] parse `BUTTONS`
- [ ] parse `LEFT`
- [ ] parse `RIGHT`
- [ ] infer if monocular

## Context
the eyelink_parser from  eyelinker is a good reference for implementation:
https://github.com/cran/eyelinker/blob/master/R/eyelink_parser.R

dkrako:
This is the current signature:
https://github.com/aeye-lab/pymovements/blob/9eb46c56230cbe5deebbc7af6797745a46af1613/src/pymovements/utils/parsing.py#L110-L114

It should be changed to:

```python
 def parse_eyelink( 
         filepath: Path | str, 
         patterns: list | None = None, 
         schema: dict | None = None, 
 ) -> tuple[pl.DataFrame, dict[str, Any]] : 
```

The keys of the dictionary should resemble the metadata.

- [ ] In `parsing.py` you can create a global dictionary `EYELINK_META_REGEXES: list[str]` where each string will be a simple string holding the regex.
The group name in the regex will then be used as a key in the output meta dictionary
(e.g. `SAMPLES\s+GAZE\s+LEFT\s+RATE\s+(?P<sampling_rate>[-]?\d*[.]\d*)\s+TRACKING\s+(?P<tracking>\S+)\s+FILTER\s+(?P<tracking>\d+)\s+	INPUT` would write to the keys `samping_rate`, `tracking`, and `filter`.

- [ ] Compile all the strings in `EYELINK_META_REGEXES` with `re.compile` in the `parse_eyelink()` method, not the global context, otherwise each pymovements import will slow down because of compiling all the regexes. (the way `EYE_TRACKING_SAMPLE` is defined is not optimal, compile shouldn't be done in the global context) 

[...]

PR #674, opened 2024-02-08
Feature/parse eyelink

theDebbister:
## Description

Partially fixes issue #510 

## Implemented changes

binocular is not yet supported

- [x] parses datetime, version_number, model, SR, tracked eye, filter and tracking from eyelink asc file
- [x] For the version and the model: if those cannot be parsed correctly, their value will be 'unknown'
- [x] raise Warning if no metadata at all is found

## Type of change

## Checklist:

### DIFF
--- a/src/pymovements/utils/parsing.py
+++ b/src/pymovements/utils/parsing.py
@@ -223,7 +223,7 @@ def parse_eyelink(
                     # each metadata pattern should only match once
                     compiled_metadata_patterns.remove(pattern)
 
-    # note sure not if this is the best way...
+    # note sure if this is the best way...
     if metadata:
         metadata['version_number'], metadata['model'] = _parse_full_version(
             metadata['version_1'], metadata['version_2'],
@@ -263,7 +263,7 @@ def _parse_full_version(version_str_1: str, version_str_2: str) -> tuple[str, st
     Returns
     -------
     tuple[str, str]
-        Version number and model as strings.
+        Version number and model as strings or unknown if it cannot be parsed.
     """
     if version_str_1 == 'EYELINK II 1' and version_str_2:
         version_pattern = re.compile(r'.*v(?P<version_number>[0-9]\.[0-9]+).*')
--- a/src/pymovements/utils/parsing.py
+++ b/src/pymovements/utils/parsing.py
@@ -21,6 +21,7 @@
 from __future__ import annotations
 
 import re
+from collections import defaultdict
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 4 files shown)


src/pymovements/utils/parsing.py
```diff
@@ -194,6 +214,24 @@ def parse_eyelink(
...
+        if compiled_metadata_patterns:
+            for pattern in compiled_metadata_patterns:
+                if match := pattern.match(line):
+                    for column, value in match.groupdict().items():
+                        metadata[column] = value
+
+                    # each metadata pattern should only match once
+                    compiled_metadata_patterns.remove(pattern)
+
+    # note sure if this is the best way...
```
dkrako, line 226: I'd propose to do it this way:

```python
if not metadata:
    raise Warning()

metadata['version_number'], metadata['model'] = _parse_full_version(
    metadata['version_1'], metadata['version_2'],
)
```

I originally thought that we would also need to check if `version_1` and `version_2` is in the parsed metadata, but as it's a `defaultdict` you solved it elegantly already.

But anyway, please leave a comment there that the key/values will be generated on the fly (empty string) if the keys are not available.

tests/unit/utils/parsing_test.py
```diff
@@ -130,3 +166,98 @@ def test_parse_eyelink_raises_value_error(tmp_path, patterns):
...
+            'unknown',
+            id='unknown_version_1',
+        ),
+        pytest.param(
+            '** VERSION: EYELINK II 1\n'
+            '** EYELINK II CL Feb  1 2018 (EyeLink Portable Duo)',
+            'unknown',
+            'unknown',
+            id='unknown_version_2',
+        ),
```
dkrako, line 227: Please add another test case where there is metadata but no version information, for example like this:

```
** DATE: Wed Mar  8 09:25:20 2023
** TYPE: EDF_FILE BINARY EVENT SAMPLE TAGGED
```

which should have both model and version parsed as `unknown`

dkrako (changes requested):
This is great! You solved it really well and the testing is nice. Also thank you for taking your time for researching the version string chaos. I (almost) don't have anything to nitpick! ... just these tiny things I commented on :nerd_face:

src/pymovements/utils/parsing.py
dkrako: Ah now I see, that there's difference in output between the two:
My proposal would create `version_number` and `model` regardless if metadata has been parsed.

What do you think, which way would make most sense to you? I think both are valid options.

But anyway, please leave a comment there that the key/values will be generated on the fly (empty string) if the keys are not available.

src/pymovements/utils/parsing.py
```diff
@@ -209,4 +247,50 @@ def parse_eyelink(
         schema_overrides=schema_overrides,
     )
 
-    return df
+    return df, metadata
```
dkrako, line 250: Please cast `metadata: defaultdict` to a simple `dict` when returning.

A defaultdict can result in unexpected behavior and the documentation says it returns a plain dictionary.

dkrako (changes requested):

src/pymovements/utils/parsing.py
theDebbister: I left it like it was for now, but we can still adapt it.

dkrako (approved):
This is great! Thanks a lot!

PR #676, opened 2024-02-09
feat: add text stimulus class

SiQube:


src/pymovements/stimuli/text_stimulus.py
```diff
@@ -0,0 +1,38 @@
...
+import polars as pl
+
+import pymovements as pm  # pylint: disable=cyclic-import
+from pymovements.stimuli.stimulus import StimulusDataFrame
+
+
+class TextStimulusDataFrame(StimulusDataFrame):
+    def __init__(
+            self,
+            data: pl.DataFrame | None = None,
```
dkrako, line 32: Let's have the following signature for now, ignoring image data and focusing solely on AOIs. Let's also ignore the None case for now.

Also for the start, let's suppose we have character level AOIs just as the example file that we have.

The minimal contents of the AOIs should be:
- character: the character in the text string
- pixel_x: pixel x-position of the top left corner
- pixel_y: pixel y-position of the top left corner
- width: pixel width of aoi box
- height: pixel height of aoi box

It can potentially have additional columns
- page: the page index
- line: the line on the page
- column: the character index in the line, the naming can be improved here
- word: the word this character/substring belongs to

The `TextStimulus` constructor should then work similar to the `GazeDataFrame` constructor, where we rename columns to have consistent access later on.

I would propose this signature for the constructor:
```python
def __init__(
    self,
    aois: pl.DataFrame,
    *,
    character_column: str,
    pixel_x_column: str,
    pixel_y_column: str,
    width_column: str,
    height_column: str,
    page_column: str,
):
```

src/pymovements/stimuli/stimulus_files.py
```diff
@@ -0,0 +1,46 @@
...
+# SOFTWARE.
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Any
+
+import polars as pl
+
+
+def load_stimulus_file(
```
dkrako, line 28: let's rework this to `pymovements.stimulus.text.from_file()`

with the following signature:

```python
def from_file(
    self,
    aoi: str | Path,
    *,
    character_column: str,
    pixel_x_column: str,
    pixel_y_column: str,
    width_column: str,
    height_column: str,
    page_column: str,
) -> TextStimulus:
```

The job of this function is to read in the csv and call the `TextStimulus` constructor.

Let's keep it simple first and suppose that none of these are optional.

dkrako (changes requested):
Thank you so much for working on this task! I think the current startout supposes too big of a scope of this PR. We should keep the scope very limited and work iteratively on adding new functionalities in distinct PRs. The `TextStimulus` class will be experimental for a while, but I think this strategy is better than having too big of a scope in this big project.

This PR should focus on the basics first and only, returning a `TextStimulus` class which has an `aoi` attribute when calling `pymovements.stimulus.text.from_file()`.

src/pymovements/stimulus/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+.. rubric:: Classes
+
+.. autosummary::
+   :toctree:
+   :template: class.rst
+
+    pymovements.stimulus.Stimulus
+    pymovements.stimulus.TextStimulus
+"""
+from pymovements.stimulus.stimulus import Stimulus  # noqa: F401
```
dkrako, line 31: you can resolve F401 by adding the two imports to `__all__`

src/pymovements/stimulus/stimulus.py
```diff
@@ -0,0 +1,37 @@
...
+
+    path : str | Path | DatasetPaths
+        Path to the dataset directory. You can set up a custom directory structure by passing a
+        :py:class:`~pymovements.DatasetPaths` instance.
+    """
+
+    def __init__(
+            self,
+    ) -> None:
+        self.stimulus: list[Stimulus] = []
```
dkrako, line 37: I'm not sure if I follow. Why would a stimulus be a list of stimuli?
I think at this stage it's hard to know what this class should actually do, as we only have one type of stimulus and don't know yet what common attributes and functionalities would be needed across stimulus types.

I would just remove the init for now and create an empty class.

src/pymovements/stimulus/text.py
```diff
@@ -0,0 +1,94 @@
...
+        pixel_x_column: str,
+        pixel_y_column: str,
+        width_column: str,
+        height_column: str,
+        page_column: str,
+        custom_read_kwargs: dict[str, Any] | None = None,
+) -> TextStimulus:
+    """ Load text stimulus from file"""
+    if isinstance(aoi_path, str):
+        aoi_path = Path(aoi_path)
```
dkrako, line 69: polars supports str as path type, so I don't think this if is necessary

src/pymovements/stimulus/text.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+class TextStimulus(Stimulus):
+    """TextStimulus class
+    """
+
+    def __init__(
+            self,
+            aois: pl.DataFrame,
+            character_column: str,
```
dkrako, line 38: let's have a `*` here between `aois` and the column specifiers to enforce keyword arguments for the columns.

It could be that some columns will be added in the future. With the `*` we will stay backwards compatible regardless of the order of arguments in the future.

src/pymovements/stimulus/text.py
```diff
@@ -0,0 +1,94 @@
...
+            pixel_x_column: str,
+            pixel_y_column: str,
+            width_column: str,
+            height_column: str,
+            page_column: str,
+    ) -> None:
+
+        self.frame = aois.select(
```
dkrako, line 46: Let's call the attribute simply `aois`.

The text attribute will have three attributes/properties: 
- `text`
- `aois`
- `image`

We will take care of the `text` property as the followup.

src/pymovements/stimulus/text.py
```diff
@@ -0,0 +1,94 @@
...
+            pl.col(character_column),
```
dkrako, line 47: I wouldn't drop the remaining columns, there could be some further information present that's necessary for the user.

tests/files/toy_text_1_1_aoi.csv
dkrako, line 1: I think we can limit the examples to a single file as they have the exact same columns and formatting.
We will add further example files when we stumble on signficant variation of columns/format.

Let's keep this one here (`tests/files/toy_text_1_1_aoi.csv`) and rename it to `stimulus_text_aoi_example.csv`

tests/files/toy_text_2_2_aoi.csv
dkrako, line 1: I think you can remove this file again

tests/files/toy_text_3_3_aoi.csv
dkrako, line 1: I think you can remove this file again

tests/unit/stimulus/text_test.py
```diff
@@ -0,0 +1,163 @@
...
+            id='toy_text_2_2_aoi',
+        ),
+        pytest.param(
+            'tests/files/toy_text_3_3_aoi.csv',
+            id='toy_text_3_3_aoi',
+        ),
+    ],
+)
+def test_str_aoi_path(aoi_file):
+    pm.stimulus.text.from_file(
```
dkrako, line 45: these tests currently do not check for expected results.

there are a few things you can check (in separate test cases or combined, as you wish).

- is the returned object a `TextStimulus`
- does the `TextStimulus.aois` dataframe have the resulting columns that are expected from the example file?
- does the `TextStimulus.aois` dataframe have the expected number of rows expected from the example file?

tests/unit/stimulus/text_test.py
```diff
@@ -0,0 +1,163 @@
...
+            Path('tests/files/toy_text_2_2_aoi.csv'),
+            id='toy_text_2_2_aoi',
+        ),
+        pytest.param(
+            Path('tests/files/toy_text_3_3_aoi.csv'),
+            id='toy_text_3_3_aoi',
+        ),
+    ],
+)
+def test_Path_aoi_path_kwargs(aoi_file):
```
dkrako, line 134: I think you can combine all these four test functions into one by using parametrize over the args/kwargs you passt to `from_file()`

src/pymovements/stimulus/text.py
```diff
@@ -0,0 +1,94 @@
...
+            self,
+            aois: pl.DataFrame,
+            character_column: str,
+            pixel_x_column: str,
+            pixel_y_column: str,
+            width_column: str,
+            height_column: str,
+            page_column: str,
+    ) -> None:
+
```
dkrako, line 45: the columns specifiers have to be stored somehow otherwise we don't which is which after initialization

dkrako (changes requested):
Thanks a lot, that's a good direction the PR is taking!
There is some more work needed around the edges but we're getting to it soon!

src/pymovements/stimulus/stimulus.py
dkrako: we could also remove the class completely for now, just as you like

src/pymovements/stimulus/text.py
SiQube: we need this to check for suffix -- I can also change it to split at dot and check for the suffix differently?

src/pymovements/stimulus/stimulus.py
SiQube: the idea was similar to the gazedataframe but I deleted it for now

src/pymovements/stimulus/__init__.py
SiQube: done

src/pymovements/stimulus/text.py
SiQube: done

tests/files/toy_text_2_2_aoi.csv
SiQube: done

src/pymovements/stimulus/text.py
SiQube: done

src/pymovements/stimulus/text.py
SiQube: done

tests/unit/stimulus/text_test.py
SiQube: done

dkrako (commented):
Great! Just two minor things and then there's the question about if we should clone a input dataframe.

On a more general note, I would like to discuss tomorrow if we want the naming to be

1. `Stimulus.aois`
2. or `Stimulus.areas`

I orignally proposed `aois`, but somehow I think `areas` would be a better fit.

src/pymovements/stimulus/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+# SOFTWARE.
+"""Provides access to stimulus classes.
+
+.. rubric:: Classes
+
+.. autosummary::
+   :toctree:
+   :template: class.rst
+
+    pymovements.stimulus.Stimulus
```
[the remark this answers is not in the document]
SiQube: good catch -- done

src/pymovements/stimulus/text.py
```diff
@@ -0,0 +1,90 @@
...
+import polars as pl
+
+
+class TextStimulus:
+    """TextStimulus class."""
+
+    def __init__(
+            self,
+            *,
```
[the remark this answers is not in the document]
SiQube: done

src/pymovements/stimulus/text.py
```diff
@@ -0,0 +1,90 @@
...
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Any
+
+import polars as pl
+
+
+class TextStimulus:
+    """TextStimulus class."""
```
dkrako, line 30: Please add a description of the parameters here.

I'm somewhat surprised that pydoclint etc. didn't complain.

src/pymovements/stimulus/text.py
```diff
@@ -0,0 +1,90 @@
...
+        *,
+        aoi_column: str,
+        pixel_x_column: str,
+        pixel_y_column: str,
+        width_column: str,
+        height_column: str,
+        page_column: str,
+        custom_read_kwargs: dict[str, Any] | None = None,
+) -> TextStimulus:
+    """Load text stimulus from file."""
```
dkrako, line 64: Please add a description of the parameters here.

src/pymovements/__init__.py
```diff
@@ -61,6 +62,8 @@
     'plotting',
     'synthetic',
     'utils',
+
```
dkrako, line 65: add `'stimulus'`

src/pymovements/__init__.py
```diff
@@ -37,6 +37,7 @@
 from pymovements.gaze import Experiment
 from pymovements.gaze import GazeDataFrame
 from pymovements.gaze import Screen
+from pymovements.stimulus import text
```
dkrako, line 40: add `from pymovements import stimulus`

dkrako (changes requested):
Excellent! There are some tiny bits in the init and the docstrings need parameter specifications but that's about it.

src/pymovements/__init__.py
SiQube: done

src/pymovements/__init__.py
SiQube: done

src/pymovements/stimulus/text.py
SiQube: done

dkrako (approved):
Excellent! Thank you so much for this work!

PR #691, opened 2024-02-22
fix: update checksum of PoTeC resource file

dkrako:
The checksum of the PoTeC resource file has changed and thus `tox -e integration` failed.

I have updated the checksum.

@theDebbister can you confirm that the zip file has changed?
I can see a revision at the beginning of the current month here: https://osf.io/tgd9q

### DIFF
--- a/docs/source/bibliography.bib
+++ b/docs/source/bibliography.bib
@@ -36,11 +36,11 @@
 }
 
 @misc{potec,
-    author = {Jäger, Lena A. and Haller, Patrick and Jakobi, Deborah Noemie},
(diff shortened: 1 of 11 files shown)


SiQube:
I'll approve but please wait for @theDebbister

SiQube (approved):

theDebbister (commented):
yes, it has changed. totally forgot to mention it. But the reference and abstract have also changed in the meantime, I will quickly adapt and push this today. After that it's good to merge from my side..

SiQube:
@theDebbister can you also adjust the reference etc. for pymovements

PR #698, opened 2024-02-23
refactor: use gaze.io for loading dataset files

prassepaul:
## Description

refactors load_files.


src/pymovements/dataset/dataset_files.py
```diff
@@ -210,94 +212,46 @@ def load_gaze_files(
...
-                gaze_data,
-                experiment=definition.experiment,
-                trial_columns=definition.trial_columns,
-                time_column=definition.time_column,
-                distance_column=definition.distance_column,
-                pixel_columns=definition.pixel_columns,
-                position_columns=definition.position_columns,
-                velocity_columns=definition.velocity_columns,
-                acceleration_columns=definition.acceleration_columns,
+            gaze_df = load_gaze_file(
```
dkrako, line 227: You can leave out the `if-else` here, as both calls are identical.

src/pymovements/gaze/gaze_dataframe.py
```diff
@@ -184,6 +184,7 @@ class GazeDataFrame:
```
dkrako, line 184: I think it's finally safe to remove `valid_pixel_position_columns`, `valid_position_columns`, `valid_velocity_columns`, `valid_acceleration_columns`.

They aren't documented anywhere and are only used by `load_gaze_file()` in `dataset_files.py` for files that we have saved after preprocessing.

src/pymovements/dataset/dataset_files.py
```diff
@@ -318,16 +272,75 @@ def load_gaze_file(
...
-            gaze_df = pl.read_csv(filepath)
+            gaze_df = from_csv(
+                filepath,
+                trial_columns=definition.trial_columns,
+                add_columns=add_columns,
+                column_dtypes=definition.filename_format_dtypes,
+            )
+
+            pixel_columns: list[str] = list(
+                set(GazeDataFrame.valid_pixel_position_columns) & set(gaze_df.frame.columns),
```
dkrako, line 290: let's drop the usage of  `GazeDataFrame.valid_pixel_position_columns` here, as the preprocessed csv files are unnested using `GazeDataFrame.unnest()` before, so we actually know all the suffixes that can appear.

I now also see that the previous code with computing the union of two sets of column names is a bit error-prone, as we currently have to take care of the right order when passing `pixel_columns`.

Instead, let's write it this way:

```python
# suffixes as ordered after using GazeDataFrame.unnest()
component_suffixes = ['x', 'y', 'xl', 'yl', 'xr', 'yr', 'xa', 'ya']

pixel_columns = ['pixel_' + suffix for suffix in component_suffixes]
pixel_columns = [c for c in pixel_columns if c in gaze_df.frame.columns]
```

The same can be done for `position_columns`, `velocity_columns` and `acceleration_columns`

src/pymovements/dataset/dataset_files.py
```diff
@@ -318,16 +272,75 @@ def load_gaze_file(
...
+            )
+            velocity_columns: list[str] = list(
+                set(GazeDataFrame.valid_velocity_columns) & set(gaze_df.frame.columns),
+            )
+            acceleration_columns: list[str] = list(
+                set(GazeDataFrame.valid_acceleration_columns) & set(gaze_df.frame.columns),
+            )
+
+            column_specifiers: list[list[str]] = []
+            if len(pixel_columns) > 0:
```
dkrako, line 303: I remember you telling me that this is quite hacky but let's leave it as is because I also don't know how to solve it more elegantly.

src/pymovements/dataset/dataset_files.py
```diff
@@ -318,16 +272,75 @@ def load_gaze_file(
...
+                add_columns=add_columns,
+                column_dtypes=definition.filename_format_dtypes,
+                **custom_read_kwargs,
+            )
     elif filepath.suffix == '.feather':
-        gaze_df = pl.read_ipc(filepath)
+        gaze_df = from_ipc(
+            filepath,
+            experiment=definition.experiment,
+            column_map=definition.column_map,
```
dkrako, line 338: The column_map shouldn't be necessary here, it's rather used when using the original dataset is loaded.

I think we currently expect `feather` files to be saved after preprocessing, but in the future we should probably have a similar separation as when loading preprocessed and original CSV files. But that's out of scope of the PR. Just removing the `column_map` here is enough.

src/pymovements/dataset/dataset_files.py
```diff
@@ -318,16 +272,75 @@ def load_gaze_file(
...
+            add_columns=add_columns,
+            column_dtypes=definition.filename_format_dtypes,
+        )
     elif filepath.suffix == '.asc':
-        gaze_df, _ = parse_eyelink(filepath, **custom_read_kwargs)
+        gaze_df = from_asc(filepath, **custom_read_kwargs)
```
dkrako, line 343: `add_columns` and `column_dtypes` should be probably passed to `from_asc` too

tests/unit/dataset/dataset_test.py
```diff
@@ -334,7 +334,10 @@ def test_load_correct_raw_gaze_dfs(dataset_configuration):
 
     expected_gaze_dfs = dataset_configuration['raw_gaze_dfs']
     for result_gaze_df, expected_gaze_df in zip(dataset.gaze, expected_gaze_dfs):
-        assert_frame_equal(result_gaze_df.frame, expected_gaze_df.frame)
+        assert_frame_equal(
```
dkrako, line 337: You can just use `check_column_order=False` here

tests/unit/dataset/dataset_test.py
```diff
@@ -351,7 +354,10 @@ def test_load_correct_preprocessed_gaze_dfs(dataset_configuration):
 
     expected_gaze_dfs = dataset_configuration['preprocessed_gaze_dfs']
     for result_gaze_df, expected_gaze_df in zip(dataset.gaze, expected_gaze_dfs):
-        assert_frame_equal(result_gaze_df.frame, expected_gaze_df.frame)
+        assert_frame_equal(
```
dkrako, line 357: You can just use `check_column_order=False` here

dkrako (changes requested):
Great! Thank you so much for putting all the thought into it!

There are just a few tiny details which need to be worked on and then we're set to go.

src/pymovements/dataset/dataset_files.py
prassepaul: done

src/pymovements/gaze/gaze_dataframe.py
prassepaul: included valid_... in dataset_files.py

src/pymovements/dataset/dataset_files.py
prassepaul: done

src/pymovements/dataset/dataset_files.py
prassepaul: k. no change

src/pymovements/dataset/dataset_files.py
prassepaul: done

src/pymovements/dataset/dataset_files.py
prassepaul: added the parameter in gaze.io and passed the params in dataset_files.py

tests/unit/dataset/dataset_test.py
prassepaul: done

tests/unit/dataset/dataset_test.py
prassepaul: done

dkrako (approved):
Excellent! Thank you so much for resolving this!

PR #702, opened 2024-03-01
feat: parse eyelink metadata

theDebbister:
## Description

partially fixes issue #510 

## Implemented changes

- [x] additionally parses resolution
- [x] parses all validations and calibrations including the result overview
- [x] new method to pre process the metadata (mainly converting types for now)

## Type of change

## How Has This Been Tested?

unit tests have been extend, but especially for the calibration and validation parsing more cases need to be added based on acutal data files in order to cover all cases

## Checklist:


src/pymovements/utils/parsing.py
```diff
@@ -253,10 +298,46 @@ def parse_eyelink(
...
+        metadata['sampling_rate'] = float(metadata['sampling_rate'])
+
+    if 'day' in metadata:
+        metadata['day'] = int(metadata['day'])
+
+    if 'year' in metadata:
+        metadata['year'] = int(metadata['year'])
+
+    if metadata['version_number'] != 'unknown':
+        metadata['version_number'] = float(metadata['version_number'])
```
dkrako, line 336: although the version number is separated by a decimal point, I wouldn't advise cast it to a float, as version 3.12 is greater than 3.2, which is not the case for a decimal number.

So not casting this field would be better I think.

src/pymovements/utils/parsing.py
```diff
@@ -253,10 +298,46 @@ def parse_eyelink(
...
+    )
+
+    if 'resolution' in metadata:
+        resolution = metadata['resolution'].split()
+        metadata['resolution'] = tuple(int(res) for res in resolution)
+
+    if 'sampling_rate' in metadata:
+        metadata['sampling_rate'] = float(metadata['sampling_rate'])
+
+    if 'day' in metadata:
```
dkrako, line 329: can you also create a full [datetime](https://docs.python.org/3/library/datetime.html#datetime-objects) object here and save it to `metadata['datetime']`?

tests/unit/utils/parsing_test.py
```diff
@@ -116,17 +117,20 @@
...
-    'version_number': '6.12',
-    'sampling_rate': '1000.00',
+    'version_number': 6.12,
+    'sampling_rate': 1000.00,
     'filter': '2',
     'tracking': 'CR',
     'tracked_eye': 'LEFT',
+    'calibrations': [],
+    'validations': [],
+    'resolution': (0, 0, 1279, 1023),
```
dkrako, line 133: I would expect the resolution tuple to be `(1280, 1024)`.

This 4-tuple here rather refers to the coordinates of two opposing screen edges.

As the pixel counting starts with zero in the EyeLink system, the actual resolution should be calculated by `(coords[3] - coords[1] + 1, coords[4] - coords[2] + 1)`

tests/unit/utils/parsing_test.py
```diff
@@ -251,50 +264,109 @@ def test_parse_eyelink_version(tmp_path, metadata, expected_version, expected_mo
 
     assert metadata['version_number'] == expected_version
     assert metadata['model'] == expected_model
+    assert metadata['time'] == time
```
dkrako, line 267: I think it makes sense to test the time field separately. keep the separate test function for time (maybe rename it to `test_parse_eyelink_datetime()` and check for `metadata['datetime']`

dkrako (changes requested):
Excellent, thank you so much for working on this!

There are a few details which need some improvement, but it's the last mile.

src/pymovements/utils/parsing.py
```diff
@@ -29,7 +29,7 @@
 import polars as pl
 
 EYE_TRACKING_SAMPLE = re.compile(
-    r'(?P<time>(\d{7,10}))\s+'
+    r'(?P<time>(\d{6,10}))\s+'
```
dkrako, line 32: is it possible to go even lower (or higher) here, like `\d{4, 20}` or just a `\d+`? I don't have the EyeLink documentation at hand, and I don't really know where these ranges came from.

theDebbister: I quickly checked the manual and it does not say anthing specific about the length. But I assume it won't be lower than 3 or 4 as this would mean that the first timestamp is within 1 second of the experiment start.. But the manual says that only sample lines begin with digits so we are probably safe if we use `\d+`

dkrako (approved):
Excellent! Thanks a lot for putting all the work in this!

PR #710, opened 2024-03-28
fix!: correct pixel origin string to upper left instead of lower left

dkrako:
## Description

The pixel origin `lower left` is actually `upper left`. Luckily all logic is correct, just the naming wrong.

I didn't create any deprecation as this is a serious misnomer a user should be informed about the change.
All dataset definitions are updated in this PR, so in case of the public datasets nothing has to be changed.
In case of local datasets the user will be notified to change the experiment definition to `upper left` this way.

I think this is the best solution, as adding the correct calculation for `lower left` (and potentially others) would result in silent errors for users which created dataset definitions with pymovements<v0.18.0. This way we have at least a few versions inbetween that notify the user of this bugfix.

## Implemented changes

- [x] replace the string "lower left" with "upper left" in all files
- [x] add helpful error message to update dataset definition

## Type of change

## How Has This Been Tested?

- [x] test helpful error message

## Checklist:


prassepaul (approved):
finally!

SiQube (approved):

PR #711, opened 2024-03-28
fix: forward experiment from dataset definition to from_asc

dkrako:
integration test did not pass for `ToyDatasetEyeLink`, I checked why, and the experiment definition was not passed correctly. easy fix. added test.


prassepaul (approved):
makes sense

PR #719, opened 2024-04-19
feat: autodetect column names

prassepaul:
## Description

Added auto_column_detection


dkrako (changes requested):
Great! Now all of this can be removed from `dataset_files.py` right?

https://github.com/aeye-lab/pymovements/blob/8edc901b229174a6dc40a2c1c454457d3bd7ef76/src/pymovements/dataset/dataset_files.py#L283-L313

dkrako (dismissed):
Ah and now I noticed that the `auto_column_detect` argument should also be available for all the functions in `gaze.integration` and `gaze.io`. Could you add it to these functions too, please?

prassepaul:
I did the changes

dkrako:
this should be the fix: #729
can you approve and merge the changes afterwards from the mainbranch into your PR?

SiQube (approved):

PR #747, opened 2024-07-08
map gaze data to aois

SiQube:
add feature to insert another column to the gaze dataframe which indicates whether an aoi was looked at.


src/pymovements/gaze/gaze_dataframe.py
```diff
@@ -792,6 +792,70 @@ def columns(self) -> list[str]:
...
+            if eye == 'left':
+                x_eye = 'pixel_xl'
+                y_eye = 'pixel_yl'
+            elif eye == 'right':
+                x_eye = 'pixel_xr'
+                y_eye = 'pixel_yr'
+            else:
+                # auto to right eye
+                x_eye = 'pixel_xr'
+                y_eye = 'pixel_yr'
```
prassepaul, line 821: here we should allow more namings as in PR #719 

example:
# suffixes as ordered after using GazeDataFrame.unnest() 
 component_suffixes = ['x', 'y', 'xl', 'yl', 'xr', 'yr', 'xa', 'ya'] 
  
 pixel_columns = ['pixel_' + suffix for suffix in component_suffixes] 
 pixel_columns = [c for c in pixel_columns if c in gaze_df.frame.columns]

src/pymovements/gaze/gaze_dataframe.py
```diff
@@ -792,6 +792,70 @@ def columns(self) -> list[str]:
...
         return self.frame.columns
 
+    def map_to_aois(
+        self,
+        aoi_dataframe: pm.stimulus.TextStimulus,
+        *,
+        eye: str = 'auto',
+    ) -> None:
+        """Map gaze data to aois.
+
```
prassepaul, line 802: maybe add the comment that we treat the aois as (<= >=); so give a comment on the boundaries

prassepaul (changes requested):
see comments. so in short: allow more names for the pixel-coordinates. add more details to the functionality.

saeub:
Just out of curiosity, what is the reasoning behind having all the comparisons use `<=`? If I want to have AOIs that are right next to each other without overlapping, wouldn't that mean that I would have to define them like this:

```csv
aoi,top_left_x,top_left_y,width,height
first,0,0,49.99999999999999,100
second,50,0,50,100
```

That seems much more cumbersome than having the comparisons involving width/height use `<`:

```csv
aoi,top_left_x,top_left_y,width,height
first,0,0,50,100
second,50,0,50,100
```

prassepaul:
> Just out of curiosity, what is the reasoning behind having comparisons like `aoi['top_left_x'] <= x <= aoi['top_left_x'] + width` (with `<=` on both sides)? If I want to have AOIs that are right next to each other without overlapping, wouldn't that mean that I would have to define them like this:
> 
> ```
> aoi,top_left_x,top_left_y,width,height
> first,0,0,49.99999999999999,100
> second,50,0,50,100
> ```
> 
> That seems much more cumbersome than having the comparisons like `aoi['top_left_x'] <= x < aoi['top_left_x'] + width`:
> 
> ```
> aoi,top_left_x,top_left_y,width,height
> first,0,0,50,100
> second,50,0,50,100
> ```

You are right. Maybe we should decide, how to handle this. I we have AOIs next to each other we need to decide, how we want to handle it. are there any conventions? If we define it as '>' and '<', we would not allow points to be added to an AOI when it is on the border. Maybe this is the best option?

saeub:
Personally I think that `aoi['top_left_x'] <= x < aoi['top_left_x'] + width` would be the most intuitive and useful, since it avoids both overlap and "AOI-free" zones with adjacent AOIs. I imagine it could be quite annoying if there are some fixations that are right between two words but treated as if they were not even on the text.

SiQube:
changed boundary conditions @saeub @prassepaul, I'll add another PR later which enables to include the entire boundary.

prassepaul (approved):

prassepaul (approved):

PR #752, opened 2024-07-10
fix: add subject ID to dataset trial columns

saeub:
## Description

Some dataset definitions (like SB-SAT) do not include the subject ID in `trial_columns`, while others (like PoTeC) do. There might be a reason for this that I'm unaware of, but it leads to this bug:

```py
sbsat = pm.Dataset("SBSAT", "sb-sat").download()
sbsat.pix2deg().pos2vel(method="savitzky_golay", window_length=11, degree=3)
sbsat.detect("ivt")
sbsat.detect("microsaccades")
print(sbsat.events)
```

In this example, `detect("ivt")` works just fine, but when running `detect("microsaccades")`, the added saccades will have a subject ID `null` in the event frame:

```
shape: (3_541, 7)
 ┌────────────┬───────────┬───────────┬──────────┬─────────┬─────────┬──────────┐
 │ subject_id ┆ book_name ┆ screen_id ┆ name     ┆ onset   ┆ offset  ┆ duration │
 │ ---        ┆ ---       ┆ ---       ┆ ---      ┆ ---     ┆ ---     ┆ ---      │
 │ i64        ┆ str       ┆ i64       ┆ str      ┆ i64     ┆ i64     ┆ i64      │
 ╞════════════╪═══════════╪═══════════╪══════════╪═════════╪═════════╪══════════╡
 │ 2          ┆ flytrap   ┆ 4         ┆ fixation ┆ 1754221 ┆ 1754321 ┆ 100      │
 │ 2          ┆ flytrap   ┆ 4         ┆ fixation ┆ 1758061 ┆ 1758201 ┆ 140      │
 │ 2          ┆ flytrap   ┆ 5         ┆ fixation ┆ 1802511 ┆ 1802615 ┆ 104      │
 │ null       ┆ flytrap   ┆ 1         ┆ saccade  ┆ 1612856 ┆ 1612865 ┆ 9        │
 │ …          ┆ …         ┆ …         ┆ …        ┆ …       ┆ …       ┆ …        │
 │ null       ┆ dickens   ┆ 5         ┆ saccade  ┆ 3173398 ┆ 3173472 ┆ 74       │
 │ null       ┆ dickens   ┆ 5         ┆ saccade  ┆ 3174133 ┆ 3174201 ┆ 68       │
 │ null       ┆ dickens   ┆ 5         ┆ saccade  ┆ 3174204 ┆ 3174226 ┆ 22       │
[...]

PR #754, opened 2024-07-15
fix automated documentation by pinning sphinx to be <7.4

SiQube:
temporary solution to get documentation to run, which failed PR #752


prassepaul (approved):
k

PR #763, opened 2024-07-19
fix: default traceplot cmaps

saeub:
## Description

Currently, calling `pm.plotting.traceplot(dataset.gaze[0])` in a notebook (even without providing any color values), raises `ValueError: data mapping points must start with x=0 and end with x=1`, because the default cmap definitions are invalid:
https://github.com/aeye-lab/pymovements/blob/5b195e18571aca39f575867aa7fc6549f15c933d/src/pymovements/plotting/traceplot.py#L63-L67

From the [matplotlib documentation](https://matplotlib.org/stable/api/_as_gen/matplotlib.colors.LinearSegmentedColormap.html):
> Each row in the table for a given color is a sequence of x, y0, y1 tuples. In each sequence, x must increase monotonically from 0 to 1.

So to keep alpha constant, it would have to look like this:
```py
 'alpha': [
     (0.0, 1.0, 1.0),
     (1.0, 1.0, 1.0),
 ],
```
but since alpha is constant at 1.0, it can simply be removed.

## Implemented changes

- [x] Remove invalid cmap definition for alpha channel

## Type of change

## How Has This Been Tested?

Tested manually. This error only occurs when showing a plot in a notebook, not during testing, so it is probably very hard to test automatically.

## Checklist:

### DIFF
--- a/src/pymovements/plotting/traceplot.py
+++ b/src/pymovements/plotting/traceplot.py
@@ -60,11 +60,6 @@ DEFAULT_SEGMENTDATA: LinearSegmentedColormapType = {
         (0.5, 0.0, 0.0),
         (1.0, 0.0, 0.0),
     ],
-    'alpha': [
-        (1.0, 1.0, 1.0),
-        (1.0, 1.0, 1.0),
-        (1.0, 1.0, 1.0),
-    ],
 }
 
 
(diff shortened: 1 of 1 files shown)


SiQube (approved):
as usual great work -- thank you so much!

PR #770, opened 2024-08-01
remove debug print statements in tests

SiQube:


saeub:
Are these possibly also debug print statements?

https://github.com/aeye-lab/pymovements/blob/bac89121d3a0e876cc5c344da8b683cfdac6cd34/src/pymovements/dataset/dataset_files.py#L70-L71

saeub (approved):
Otherwise, LGTM

PR #771, opened 2024-08-05
enable fixation only datasets

### LINKED ISSUE #716: add support for event-only datasets
## Description of the problem
There are a lot of datasets published that only include event data and no raw gaze samples.

It would be great if the dataset library could also support these.

## Description of a solution

The `DatasetDefinition` should probably extended with a new attribute that signifies the type of data to expect.

My first proposal would be `datatype: Literal['raw', 'events']`. Let's brainstorm if we can find a better name.

Additionally, the `Dataset.load()` method must be changed accordingly:

https://github.com/aeye-lab/pymovements/blob/cb9ef9571c5b24f7609928d18efc3ae2520c1d03/src/pymovements/dataset/dataset.py#L77-L135

Currently, it always runs `Dataset.load_gaze_files()`, but this should then be dependent on `DatasetDefinition.datatype`.
Also, events should be set to `events: bool | None = None` and later assume a boolean value dependent on `DatasetDefinition.datatype`

I propose the following signature:

```python
    def load(
            self,
            *,
            gaze: bool | None = None,
            events: bool | None = None,
            preprocessed: bool = False,
            subset: dict[str, float | int | str | list[float | int | str]] | None = None,
            events_dirname: str | None = None,
            preprocessed_dirname: str | None = None,
            extension: str = 'feather',
    ) -> Dataset:
```

This would be backwards compatible:
```python
if dataset.definition.datatype == 'raw' and gaze is None:
    gaze = True
else:
    gaze = False

if dataset.definition.datatype == 'events' and events is None:
    events = True
else:
    events = False
```

- [ ] add argument to `DatasetDefinition` to indicate type of dataset
- [ ] adjust `Dataset.load()` default values to decide loading gaze or events during runtime

SiQube:
eventually resolves #716


SiQube:
for pre-commit ci to pass need to merge #774 first

SiQube:
resolves #716

SiQube:
we have to think about [integration tests](https://github.com/aeye-lab/pymovements/blob/main/tests/integration/public_dataset_processing_test.py) when considering this feature.

we may have to add a line which checks whether `has_gaze_files` is true. if not true, skip test. 


secondly, we may have to add integration tests, for datasets which just have `has_precomputed_event_files`. this might be more difficult as we do not enforce any standardized notations.

src/pymovements/events/precomputed.py
```diff
@@ -0,0 +1,37 @@
...
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+"""Module for the TextDataFrame."""
```
prassepaul, line 20: TextDataFrame sounds weird. Would change it to 'Module for the precomputed events'

prassepaul (changes requested):
I would only change the small thing pointed out... really big PR...

prassepaul (approved):

PR #774, opened 2024-08-17
as per repository types-all is deprecated

SiQube:
types-all is deprecated and leads to pre-commit.ci failing, see e.g. #771


saeub (approved):
LGTM

Issue #379, opened 2023-05-11
Add tutorial on how to add public datasets to pymovements


dkrako:
## Description of the problem
There are some small details which need to be done for adding new datasets to our dataset library.

For example it needs to be specified which test cases need to be added.

## Description of a solution
This could be done either as a tutorial in the tutorial section or an alternative would be to put it in a more comprehensive contributing guide.

I already planned to extend the contributing guide as it is currently a bit too brief for beginners I think.
I think the matplotlib contributing guide is a very good example on how to do this: https://matplotlib.org/stable/devel/contributing.html

## Minimum acceptance criteria
- [ ] a howto guide on how to add new datasets, either as a tutorial or in a section of the contributing guide

SiQube:
additionally the contributing tutorial should contain the following way to test the dataset:

```console
tox -e integration -- tests/integration/public_dataset_processing_test.py::test_public_dataset_processing[PoTeC]
```

originally mentioned in #692

saeub:
When creating a new file through the GitHub web interface, the file title and content can be pre-filled: https://github.com/aeye-lab/pymovements/new/main/src/pymovements/datasets?filename=my_dataset.yaml&value=template

This might lower the threshold for people to start adding their dataset without having to clone the repository first.

PR #785, opened 2024-08-22
merge tests for public datasets in a single file

### LINKED ISSUE #381: Refactor specific dataset tests in a single file
## Description of the problem
Currently there is a lot of copy-pasting going on in `tests/datasets` where each single dataset gets an own file and the only thing that changes is the dataset name and the path.

## Description of a solution
All these files can be refactored into the already existing `tests/datasets/datasets_test.py` which currently only checks if each dataset is registered and if the attributes are the specified ones.

- [ ] get rid of the specific dataset files in `tests/datasets` and move them into a single file.

SiQube:
merged public datasets utilities into single file.

resolves #381 

I stumbled upon this issue when working on #379. I think this is user friendlier to add more datasets (just add one line to `public_dataset` parametrize in in the file instead of copying file etc.)


tests/unit/datasets/datasets_test.py
```diff
@@ -20,51 +20,126 @@
...
 import pytest
 
 import pymovements as pm
 
 
 @pytest.mark.parametrize(
-    ('definition_class', 'dataset_name'),
+    ('public_dataset', 'dataset_name'),
     [
-        pytest.param(pm.datasets.ToyDataset, 'ToyDataset', id='ToyDataset'),
```
SiQube, line 31: I moved this s.t. we have alphabetical ordering

prassepaul (approved):

PR #786, opened 2024-08-24
feat!: resume archive extraction by skipping existing files

### LINKED ISSUE #488: Don't extract dataset archives twice
## Description of the problem

Currently calling something like this at the top of your notebook
```python
dataset = pm.Dataset('JuDo1000', path='data/JuDo1000')
dataset.download()
```
will lead to extracting the archives everytime again.

This was implemented this way before as it was the simplest way without checking that all data was correctly extracted.

Of course calling
```python
dataset = pm.Dataset('JuDo1000', path='data/JuDo1000')
dataset.download(extract=False)
```
would handle this, but we would need to change the notebook for that.

## Description of a solution

It would be much nicer to have some kind of detection in place to check if the dataset archives were already extracted.

There we would need to check at least if all files were extracted.
I personally would feel safer if there would also be some checking if the contents were extracted correctly, because it could be that the extraction was aborted during the last file. In such a case we would have no chance in getting a waning without checking the checksums / contents of each file.

SiQube:
# :boom: Breaking change
`Dataset.extract()` now accepts only keyword arguments, no positional arguments.
Replace something like `Dataset.extract(True)` with `Dataset.extract(remove_finished=True)`

# Description

resolves #488 eventually

currently, tox does not work locally for whatever reason. 

- [x] time vs old extraction
- [x] coverage


SiQube:
using this script

```python
from pathlib import Path
import pymovements as pm

pm.utils.archives.extract_archive(Path('gazebasevr.zip'))
```

and predownloaded gazebase-vr -- I executed this command

```console
time python t.py
```


for this branch:
```console
Extracting gazebasevr.zip to .

real	6m53,025s
user	0m31,852s
sys	0m8,898s
```


for current main:
```console
Extracting gazebasevr.zip to .

real	7m12,829s
user	0m33,634s
sys	0m9,835s
```

different workloads might affect the performance but it should be not too much slower. (in the example above the loop variant was even faster (?!?))

src/pymovements/utils/archives.py
```diff
@@ -151,10 +151,16 @@ def _extract_tar(
         Compression filename suffix.
     """
     with tarfile.open(source_path, f'r:{compression[1:]}' if compression else 'r') as archive:
-        if sys.version_info < (3, 12):  # pragma: <3.12 cover
-            archive.extractall(destination_path)
-        else:  # pragma: >=3.12 cover
-            archive.extractall(destination_path, filter='tar')
+        for member in archive.getnames():
+            if (
+                os.path.exists(os.path.join(destination_path, member)) and
```
dkrako, line 156: only checking if the path exists can be an issue if the user aborted extraction. then there will be a partially written file on the system and this if condition won't notice that.

luckily the [`TarInfo`](https://docs.python.org/3/library/tarfile.html#tarfile.TarInfo) has fields for `size` and `chksum`.

Checking the size should be sufficient for most use cases.
You can create a private function in `archives.py` to make the if condition a bit more readable.

For paranoid users we could add checksum functionality after extraction in another PR.

dkrako (changes requested):
great, thanks a lot! that was what I had in mind when creating #488.

We should introduce a new argument to let the user decide if continuing is desired or extracting should be done for all members. Something like `continue: bool = True` would be already sufficient.

src/pymovements/utils/archives.py
dkrako: any updates on checking the file size?

SiQube: there is a problem when we have `remove_top_level`, let's discuss Wednesday

dkrako: Until then, can you elaborate more on the problem in a written comment?
For me it's easier to process and decide when it's written rather than spoken.

SiQube: basically, if you extract twice, and `remove_finished` or `remove_top_level` double extraction does not work -- since the original extraction is removed. 

but this is a user choice and one can expect if you delete the download that you can't extract it anymore

src/pymovements/utils/archives.py
```diff
@@ -149,12 +150,22 @@ def _extract_tar(
...
     """
     with tarfile.open(source_path, f'r:{compression[1:]}' if compression else 'r') as archive:
-        if sys.version_info < (3, 12):  # pragma: <3.12 cover
-            archive.extractall(destination_path)
-        else:  # pragma: >=3.12 cover
-            archive.extractall(destination_path, filter='tar')
+        for member in archive.getnames():
+            if (
+                    os.path.exists(os.path.join(destination_path, member)) and
+                    member[-4:] not in _ARCHIVE_EXTRACTORS and
```
dkrako, line 160: probably it would be nice to add a short note that archives are extracted regardless of the value of `skip`/ `resume`

tests/unit/utils/archives_test.py
```diff
@@ -495,3 +497,54 @@ def test_decompress_unknown_compression_suffix():
...
+                os.path.join('toplevel', 'recursive.zip'),
+                os.path.join('toplevel', 'recursive'),
+                os.path.join('toplevel', 'recursive', 'singlechild'),
+                os.path.join('toplevel', 'recursive', 'singlechild', 'test.file'),
+            ),
+            id='recursive_true_remove_finished_false',
+        ),
+    ],
+)
+def test_extract_archive_destination_path_not_None_no_remove_top_level_no_remove_finished_twice(
```
dkrako, line 526: I'm not sure of I understand what you are testing here. the test would already pass without this PR, right?

As I don't think that it's worth much effort to test this thoroughly, I'd probably just check the calls to `TarFile.extract()` for using the expected arguments via `unittest.mock` (for the second `extract_archive()` call).

dkrako (changes requested):
Great, thanks a lot for your work! There are some small issues left to work out, but we're getting there

tests/unit/utils/archives_test.py
SiQube: see adjustments via print

src/pymovements/utils/archives.py
```diff
@@ -175,7 +186,13 @@ def _extract_zip(
     """
     compression_id = _ZIP_COMPRESSION_MAP[compression] if compression else zipfile.ZIP_STORED
     with zipfile.ZipFile(source_path, 'r', compression=compression_id) as archive:
-        archive.extractall(destination_path)
+        for member in archive.namelist():
+            if (
+                os.path.exists(os.path.join(destination_path, member)) and
```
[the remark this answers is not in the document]
SiQube: done

src/pymovements/utils/archives.py
```diff
@@ -149,12 +150,22 @@ def _extract_tar(
...
+        Skip already extracted files. (default: True)
     """
     with tarfile.open(source_path, f'r:{compression[1:]}' if compression else 'r') as archive:
-        if sys.version_info < (3, 12):  # pragma: <3.12 cover
-            archive.extractall(destination_path)
-        else:  # pragma: >=3.12 cover
-            archive.extractall(destination_path, filter='tar')
+        for member in archive.getnames():
+            if (
+                    os.path.exists(os.path.join(destination_path, member)) and
```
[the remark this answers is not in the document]
SiQube: done

src/pymovements/utils/archives.py
```diff
@@ -149,12 +150,22 @@ def _extract_tar(
...
+                    member[-4:] not in _ARCHIVE_EXTRACTORS and
+                    tarfile.TarInfo(os.path.join(destination_path, member)).size > 0 and
+                    skip
```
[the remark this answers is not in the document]
SiQube: I think this is done by python on the fly

src/pymovements/utils/archives.py
```diff
@@ -149,12 +150,22 @@ def _extract_tar(
...
     with tarfile.open(source_path, f'r:{compression[1:]}' if compression else 'r') as archive:
-        if sys.version_info < (3, 12):  # pragma: <3.12 cover
-            archive.extractall(destination_path)
-        else:  # pragma: >=3.12 cover
-            archive.extractall(destination_path, filter='tar')
+        for member in archive.getnames():
+            if (
+                    os.path.exists(os.path.join(destination_path, member)) and
+                    member[-4:] not in _ARCHIVE_EXTRACTORS and
+                    tarfile.TarInfo(os.path.join(destination_path, member)).size > 0 and
```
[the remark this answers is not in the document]
SiQube: done

src/pymovements/utils/archives.py
```diff
@@ -138,6 +138,7 @@ def _extract_tar(
         source_path: Path,
         destination_path: Path,
         compression: str | None,
+        skip: bool = True,
```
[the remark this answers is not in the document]
SiQube: done -- `resume`

src/pymovements/utils/archives.py
```diff
@@ -149,18 +157,37 @@ def _extract_tar(
         Path to the directory the file will be extracted to.
     compression: str | None
         Compression filename suffix.
+    resume: bool
+        Resume if archive was already previous extracted.
+    verbose: int
```
dkrako, line 162: as it's basically just a boolean switch I would use `bool` instead of `int`. otherwise there's potential confusion about the effect of the verbosity level.

we can still "upgrade" to `int` in case we have different verbosity levels, as positive `int` is "backwards compatible" to `bool`.

src/pymovements/utils/archives.py
```diff
@@ -149,18 +157,37 @@ def _extract_tar(
...
+        Print messages for resuming each dataset resource.
     """
     with tarfile.open(source_path, f'r:{compression[1:]}' if compression else 'r') as archive:
-        if sys.version_info < (3, 12):  # pragma: <3.12 cover
-            archive.extractall(destination_path)
-        else:  # pragma: >=3.12 cover
-            archive.extractall(destination_path, filter='tar')
+        for member in tqdm(archive.getmembers()):
+            member_name = member.name
```
dkrako, line 167: I doubt these two lines increase readability. there are probably just a leftover which should be refactored away

src/pymovements/utils/archives.py
```diff
@@ -172,13 +199,32 @@ def _extract_zip(
...
+        Print messages for resuming each dataset resource.
     """
     compression_id = _ZIP_COMPRESSION_MAP[compression] if compression else zipfile.ZIP_STORED
     with zipfile.ZipFile(source_path, 'r', compression=compression_id) as archive:
-        archive.extractall(destination_path)
-
-
-_ARCHIVE_EXTRACTORS: dict[str, Callable[[Path, Path, str | None], None]] = {
+        for member in tqdm(archive.filelist):
+            member_filename = member.filename
```
dkrako, line 210: same as above

src/pymovements/utils/archives.py
```diff
@@ -172,13 +199,32 @@ def _extract_zip(
         Path to the directory the file will be extracted to.
     compression: str | None
         Compression filename suffix.
+    resume: bool
+        Resume if archive was already previous extracted.
+    verbose: int
```
dkrako, line 204: same as above

src/pymovements/utils/archives.py
```diff
@@ -172,13 +199,32 @@ def _extract_zip(
...
+                if (
+                    os.path.exists(member_dest_path) and
+                    member_filename[-4:] in _ARCHIVE_EXTRACTORS and
+                    member_size == os.path.getsize(member_dest_path)
+                ):
+                    if verbose:
+                        print(f'Skipping {member_filename} due to previous extraction')
+                    continue
+                archive.extract(member_filename, destination_path)
+            else:
```
dkrako, line 223: in contrast to the tar implementation, there is no need for an additional if-else

src/pymovements/utils/archives.py
```diff
@@ -172,13 +199,32 @@ def _extract_zip(
...
+                member_size = member.file_size
+                if (
+                    os.path.exists(member_dest_path) and
+                    member_filename[-4:] in _ARCHIVE_EXTRACTORS and
+                    member_size == os.path.getsize(member_dest_path)
+                ):
+                    if verbose:
+                        print(f'Skipping {member_filename} due to previous extraction')
+                    continue
+                archive.extract(member_filename, destination_path)
```
dkrako, line 222: this line should be outside the if clause

src/pymovements/utils/archives.py
```diff
@@ -172,13 +199,32 @@ def _extract_zip(
...
     """
     compression_id = _ZIP_COMPRESSION_MAP[compression] if compression else zipfile.ZIP_STORED
     with zipfile.ZipFile(source_path, 'r', compression=compression_id) as archive:
-        archive.extractall(destination_path)
-
-
-_ARCHIVE_EXTRACTORS: dict[str, Callable[[Path, Path, str | None], None]] = {
+        for member in tqdm(archive.filelist):
+            member_filename = member.filename
+            member_dest_path = os.path.join(destination_path, member_filename)
```
dkrako, line 211: same as above

src/pymovements/utils/archives.py
```diff
@@ -149,18 +157,37 @@ def _extract_tar(
...
     """
     with tarfile.open(source_path, f'r:{compression[1:]}' if compression else 'r') as archive:
-        if sys.version_info < (3, 12):  # pragma: <3.12 cover
-            archive.extractall(destination_path)
-        else:  # pragma: >=3.12 cover
-            archive.extractall(destination_path, filter='tar')
+        for member in tqdm(archive.getmembers()):
+            member_name = member.name
+            member_size = member.size
+            member_dest_path = os.path.join(destination_path, member_name)
```
dkrako, line 169: while you're at it, you can probably move this line into the following if-clause

src/pymovements/utils/archives.py
```diff
@@ -42,6 +44,7 @@ def extract_archive(
         remove_finished: bool = False,
         remove_top_level: bool = True,
         verbose: int = 1,
+        resume: bool = False,
```
dkrako, line 47: verbose should come after resume

dkrako: which means this change would be a breaking change as `verbose` can be passed as a positional argument.

I strongly prefer `verbose` as the last argument. We should probably but a `*, ` after the recursive parameter and enforce passing `remove_finished`, `remove_top_level`, `resume` and `verbose` as keyword arguments.

We can either introduce the breaking in this PR or in a separate one. What do you think?

Also we should keep this issue in mind when enhancing signatures in the future. To be future-proof and backwards compatible it is necessary to keep the number of positional arguments to a comfortable minimum.

tests/unit/utils/archives_test.py
dkrako: alright, that should be good enough let's leave it at that

dkrako (changes requested):
Great! A few cosmetic creases to iron out and we're good to go

dkrako (changes requested):
One additional thing I forgot:

Please include this functionality also in `Dataset.extract()` and `Dataset.download()`

src/pymovements/utils/archives.py
```diff
@@ -65,6 +68,8 @@ def extract_archive(
         Verbosity levels: (1) Print messages for extracting each dataset resource without printing
         messages for recursive archives. (2) Print additional messages for each recursive archive
         extract. (default: 1)
+    resume: bool
+        Resume previous extraction. (default: True)
```
dkrako, line 72: Please add a few more info like this:
```
Resume previous extraction by skipping existing files. Checks for correct size of existing files but not integrity. (default: True)
```

dkrako (changes requested):

src/pymovements/utils/archives.py
SiQube: done

src/pymovements/utils/archives.py
SiQube: done

src/pymovements/utils/archives.py
SiQube: good catch -- done

src/pymovements/utils/archives.py
SiQube: done

src/pymovements/utils/archives.py
SiQube: we could bool(verbose) it I guess

SiQube:
@dkrako

pyproject.toml
dkrako, line 1: I guess this dependency change is some kind of leftover from your branch? It seems unrelated to the PR

src/pymovements/utils/archives.py
```diff
@@ -41,6 +43,7 @@ def extract_archive(
         recursive: bool = True,
```
dkrako, line 43: can you add a `*,` before `recursive`? This is PR a breaking change either way, so we can make it more future proof.

src/pymovements/dataset/dataset.py
```diff
@@ -1005,6 +1010,7 @@ def extract(
             self,
             remove_finished: bool = False,
```
dkrako, line 1011: can you add a `*,` before `remove_finished`? This is PR a breaking change either way, so we can make it more future proof.

src/pymovements/dataset/dataset_download.py
```diff
@@ -34,6 +34,7 @@ def download_dataset(
         paths: DatasetPaths,
         extract: bool = True,
```
dkrako, line 35: can you add a `*,` before `extract`? This is PR a breaking change either way, so we can make it more future proof.

src/pymovements/dataset/dataset_download.py
```diff
@@ -218,6 +223,7 @@ def extract_dataset(
         paths: DatasetPaths,
         remove_finished: bool = False,
```
dkrako, line 224: can you add a `*,` before `remove_finished`? This is PR a breaking change either way, so we can make it more future proof.

src/pymovements/utils/archives.py
```diff
@@ -138,6 +145,8 @@ def _extract_tar(
         source_path: Path,
         destination_path: Path,
         compression: str | None,
+        resume: bool,
```
dkrako, line 148: can you add a `*,` before `resume`? This is PR a breaking change either way, so we can make it more future proof.

src/pymovements/utils/archives.py
```diff
@@ -149,18 +158,35 @@ def _extract_tar(
...
+                archive.extract(member.name, destination_path)
+            else:  # pragma: >=3.12 cover
+                archive.extract(member.name, destination_path, filter='tar')
 
 
 def _extract_zip(
         source_path: Path,
         destination_path: Path,
         compression: str | None,
+        resume: bool,
```
dkrako, line 188: can you add a `*,` before `resume`? This is PR a breaking change either way, so we can make it more future proof.

dkrako (changes requested):
Great, there's just the minor issue with the change of order in the positional arguments. This PR technically introduces a breaking change to `Dataset.extract()`.

What was treated as `verbose` before would be interpreted as `resume` now. As the types are compatible this could introduce silent errors (although not that critical as the arguments won't change any critical logic).

It's just a tiny change but will make the interface more future proof.

pyproject.toml
SiQube: good catch, virtualmachine problems....

dkrako (approved):
Great work, sorry for all the nitpicking but sometimes the devil's in the details

PR #789, opened 2024-08-29
fix part of matplotlib lines are type ignored

SiQube:
fixes matplotlib typing which was commented in the deleted lines


saeub (approved):
LGTM

PR #794, opened 2024-09-05
refactor: exchange seperate arguments for dicts in dataset definition

SiQube:
as discussed here comes the big refactor -- instead of using e.g. `has_gaze_files=True` and `has_precomputed_event_files=False` I refactored everything into dictionaries. for the example above:

`has_files={'gaze': True, 'precomputed_events': False}`

this makes it way easier to extend it to stimuli, reading measures etc.


src/pymovements/dataset/dataset_files.py
```diff
@@ -61,33 +61,45 @@ def scan_dataset(definition: DatasetDefinition, paths: DatasetPaths) -> pl.DataF
...
             path=paths.raw,
-            regex=curly_to_regex(definition.filename_format),
+            regex=curly_to_regex(definition.filename_format['gaze']),
             relative=True,
         )
-    if definition.has_precomputed_event_files:
+        if not fileinfo_dicts:
+            raise RuntimeError(f'no matching files found in {paths.raw}')
+
+        fileinfo_df = pl.from_dicts(data=fileinfo_dicts, infer_schema_length=1)
```
prassepaul, line 74: are we sure, that a length of 1 is enough to infer the schema?

src/pymovements/dataset/dataset_files.py
```diff
@@ -61,33 +61,45 @@ def scan_dataset(definition: DatasetDefinition, paths: DatasetPaths) -> pl.DataF
...
+    if definition.has_files['precomputed_events']:
         fileinfo_dicts = match_filepaths(
             path=paths.precomputed_events,
-            regex=curly_to_regex(definition.filename_format),
+            regex=curly_to_regex(definition.filename_format['precomputed_events']),
             relative=True,
         )
+        if not fileinfo_dicts:
+            raise RuntimeError(f'no matching files found in {paths.precomputed_events}')
+        fileinfo_df = pl.from_dicts(data=fileinfo_dicts, infer_schema_length=1)
```
prassepaul, line 92: see comment above

prassepaul (changes requested):
see comments. but I think the PR looks good

src/pymovements/dataset/dataset_files.py
SiQube: this is not actually refactored and I'd leave it currently as is

prassepaul (approved):

PR #799, opened 2024-09-12
add stimulus to plotting

SiQube:


SiQube:
scanpath plotting does not work yet, 

you can use the following dummy script to run it:

```python
import pymovements as pm


dataset = pm.Dataset('ToyDataset', path='data/ToyDataset')
dataset.download()
dataset.load()
dataset.pix2deg()
dataset.pos2vel()
dataset.detect('ivt')
print(dataset.gaze[5].frame)
dataset.compute_properties(("location", {'position_column':"pixel"}))

print(dataset.events[5].frame)

pm.plotting.scanpathplot(
    events=dataset.events[5],
    position_column='location',
    title='Scanpathplot', 
    add_stimulus=True,
    path_to_image_stimulus='tests/files/pexels-zoorg-1000498.jpg',
    stimulus_origin='upper',
    alpha=.5,
)
```

to test scanpath plotting it might be better to use our own image instead of pexels

to run the entire pipeline, checkout this branch `add-stimulus-to-plotting-as-background` and then run the script below:


```console
pip install -e .
python plot_t.py
```

prassepaul (approved):

PR #801, opened 2024-09-16
add didec dataset and tests

SiQube:


tests/files/didec_example.txt
prassepaul, line 1: are we allowed to add this file the examples? copyright?

prassepaul (changes requested):
see comment about copyright

prassepaul (approved):

PR #851, opened 2024-10-17
feat: aoi mapping extended to end coordinate instead of width/height

SiQube:
previously we were only able to map fixations to aoi via `width` and `height`. this PR introduces also `end_coordinate_column` to map fixaitons to aois.


src/pymovements/stimulus/text.py
```diff
@@ -35,48 +35,62 @@ class TextStimulus:
...
+        Name of the column which contains the y coordinate's start position of the
+        areas of interest.
+    width_column: str | None
+        Name of the column which contains the width of the area of interest. (default: None)
+    height_column: str | None
+        Name of the column which contains the height of the area of interest. (default: None)
+    end_x_column: str | None
+        Name of the column which contains the x coordinate's end position of the areas of interest.
+        (default: None)
+    end_y_column: str | None
```
prassepaul, line 51: default is missing

prassepaul (changes requested):
see comment

SiQube:
good catch -- thank you -- changed!

prassepaul (approved):

PR #852, opened 2024-10-18
feat: entire  row of aoi dataframe instead of just ia string

SiQube:
sometimes we need more than just the string within the area of interest, but also the entire information contained for this area of interest. one example for this could be the calculation of landing position within a word. this feature enhancement allows for this.


tests/unit/gaze/gaze_aoi_mapping_test.py
```diff
@@ -478,7 +637,10 @@ def test_gaze_to_aoi_mapping_char_width_height(eye, aoi_column, gaze_type):
         )
 
     gaze_df.map_to_aois(aoi_df, eye=eye, gaze_type=gaze_type)
-    assert_frame_equal(gaze_df.frame, EXPECTED_DF[f'{aoi_column}_{eye}_{gaze_type}'])
+    print(f'{aoi_column}_{eye}_{gaze_type}')
```
dkrako, line 640: this is probably a leftover from debugging, right? the assert is commented out and the prints are added.

the best way would be adding these to the assert message, this way you will have a more helpful message in case of a failing test and won't need to add the prints again. but I fully understand that the time to implement this is better spent elsewhere

dkrako (changes requested):
I'm a bit unsure about the code as it's too late in the evening for me to think about the details.

I just noticed the commented out assert in the test function, this should probably change, right?

dkrako (approved):
Approved as discussed during our meeting last week

PR #857, opened 2024-10-22
dataset: beijing sentence corpus

SiQube:


tests/unit/datasets/datasets_test.py
```diff
@@ -31,6 +31,7 @@
     ('public_dataset', 'dataset_name'),
     # XXX: add public dataset in alphabetical order
```
dkrako, line 32: this comment here should probably removed, right? It was introduced by https://github.com/aeye-lab/pymovements/commit/054d567fe5c5e9b8f42c2ee193eb5a8b61f3c993

tests/unit/datasets/datasets_test.py
dkrako, line 1: There are still tests missing in `tests/functional` right?

tests/unit/datasets/datasets_test.py
SiQube: I think we can leave it but maybe remove the `XXX`?

tests/unit/datasets/datasets_test.py
SiQube: we don't have too much functionality for pre-computed events hence no tests (yet) added

tests/unit/datasets/datasets_test.py
dkrako: # please add datasets in alphabetical order

tests/unit/datasets/datasets_test.py
dkrako: ah right, didn't see that it's an event-only dataset

dkrako (approved):
Great! Thanks a lot for increasing the size of our dataset library!

PR #858, opened 2024-10-23
feat: add support for .ias files in stimulus.text.from_file()

SiQube:
default eyelink aoi files end in `.ias`, hence I add the file type to `from_file` for text stimulus


dkrako (approved):
Great, thanks for clearing things up!

PR #862, opened 2024-10-23
dataset: add InteRead dataset

SiQube:
add [InteRead](https://aclanthology.org/2024.lrec-main.802/) dataset


dkrako (approved):
Great! Thanks a lot!

PR #873, opened 2024-10-23
hotfix: CopCo dataset precomputed events loading

SiQube:
there was a problem due to multi quotation marks


dkrako (approved):
Great! Thanks a lot for catching the bug!

PR #879, opened 2024-10-24
feat: split `TextStimulus` by column values

izaskr:
## Description

The change  introduces a splitting function for the aoi dataframe in the TextStimulus class. This allows for the aoi file to be split according to some criteria, e.g. pages. 

## Implemented changes

- [ ] Add a method to split the aoi dataframe by some criteria into a list of dataframes.

## Type of change

## How Has This Been Tested?

- [ ] Test `test_text_stimulus_splitting` checks if the number of resulting dataframes equals the expected length.

## Checklist:


tests/unit/stimulus/text_test.py
```diff
@@ -226,3 +226,34 @@ def test_text_stimulus_unsupported_format():
...
+    [
+        pytest.param(
+            'tests/files/toy_text_1_1_aoi.csv',
+            None,
+            id='toy_text_1_1_aoi',
+        ),
+        pytest.param(
+            Path('tests/files/toy_text_1_1_aoi.csv'),
+            {'separator': ','},
+            id='toy_text_1_1_aoi',
```
dkrako, line 242: avoid having the same id for different test parametrizations.

you could use something like: `toy_text_1_1_aoi_separator`

dkrako (changes requested):
I'm not sure if the `Textstimulus.aois` attribute should change types.

Before calling `split_aois_by()` the type is `pl.DataFrame` afterwards it's a list of dataframes. It would be nice to have just one type for the `aois` field.

Also, when you just split the data, you won't know the splitting criteria anymore, i.e. you won't know which dataframe is which belonging to which page.

Can you include a short code example where and how you expect the method to be used? It is probably part of a greater plan.

src/pymovements/stimulus/text.py
```diff
@@ -80,6 +80,24 @@ def __init__(
         self.end_y_column = end_y_column
         self.page_column = page_column
 
+    def split_aois_by(
```
SiQube, line 83: I think rename this to `.split(self, by: str)`

tests/unit/stimulus/text_test.py
```diff
@@ -226,3 +226,34 @@ def test_text_stimulus_unsupported_format():
...
+        start_x_column='top_left_x',
+        start_y_column='top_left_y',
+        width_column='width',
+        height_column='height',
+        page_column='page',
+        custom_read_kwargs=custom_read_kwargs,
+    )
+
+    aois_df = aois_df.split_aois_by(by='line_idx')
+    assert len(aois_df) == 2
```
dkrako, line 259: it would be nice to add two more tests:

- [ ] check that the values of the 'line_idx' column is unique for every stimulus
- [ ] check that the values of the  'line_idx' column are the same for every row per stimulus

dkrako (changes requested):
just a few things about the testing and we're ready to go!

tests/unit/stimulus/text_test.py
```diff
@@ -226,3 +226,97 @@ def test_text_stimulus_unsupported_format():
...
+        aoi_column='char',
+        start_x_column='top_left_x',
+        start_y_column='top_left_y',
+        width_column='width',
+        height_column='height',
+        page_column='page',
+        custom_read_kwargs=custom_read_kwargs,
+    )
+
+    n_unique = aois_df.aois.n_unique(subset=['line_idx'])
```
dkrako, line 320: Let's rewrite this to explicitly check the difference between the splits:

```python
aois_df = aois_df.split(by='line_idx')

unique_values = []
for df in aois_df:
    unique_values.extend(df.aois.unique(subset=['line_idx']))

assert len(unique_values) == len(set(unique_values))
```

dkrako (changes requested):
thanks a lot! just some more details on the tests and we're ready to merge

tests/unit/stimulus/text_test.py
izaskr: df.aois.unique(subset=['line_idx']) doesn't return a list, so I changed this to:

src/pymovements/stimulus/text.py
```diff
@@ -80,6 +80,24 @@ def __init__(
...
+        ----------
+        by: str
+            Splitting criteria.
+
+        Returns
+        -------
+        list[TextStimulus]
+            A list of TextStimulus objects.
+        """
+        return self.aois.partition_by(by=by, as_dict=False)
```
dkrako, line 99: I must have missed this in my review two weeks ago.

this line returns a list of `polars.DataFrame` not a list of `TextStimulus` objects.

You should probably do something like this:

```python
return [
    TextStimulus(
        aois=df,
        aoi_column=self.aoi_column,
        width_column=self.width_column,
        ...,  # remaining fields
        page_column=self.page_column,
    )
    for df in self.aois.partition_by(by=by, as_dict=False)
]
```

dkrako: @SiQube I am confused why mypy didn't complain here, as I would expect a type mismatch. I am clueless. Do you have an explanation? I am a bit worried that there's something wrong in our CI.

dkrako (changes requested):
Thank you @izaskr for your work!

Unfortunately I have missed a crucial line during my review two weeks ago as the `split() ` method doesn't return `list[TextStimulus]`.

I have posted a small code snippet how to change the logic accordingly, shouldn't be much work.

This change would also need an additional test: please check that all fields (e.g. `aoi_column`, `width_column`, etc.) in the returned `TextStimulus` splits are equal to the original `TextStimulus`.

src/pymovements/stimulus/text.py
SiQube: I actually don't know, I'll try to figure out.

dkrako (approved):
Great, finally we got there! Thank you so much @izaskr and @SiQube for your work!

PR #883, opened 2024-10-24
Updated plotting function!

ivanvajs1996:
The plotting function has been updated with the linear fit line and the R2 residual display.

### DIFF
--- a/src/pymovements/plotting/main_sequence_plot.py
+++ b/src/pymovements/plotting/main_sequence_plot.py
@@ -26,16 +26,21 @@ from matplotlib.collections import Collection
 
 from pymovements.events import EventDataFrame
 
+from sklearn.metrics import r2_score
+import numpy as np
+
+
 
 def main_sequence_plot(
-        event_df: EventDataFrame,
+ event_df: EventDataFrame,
         marker_size: float = 25,
         color: str = 'purple',
+        color_line: str = 'red',
         alpha: float = 0.5,
         marker: str = 'o',
         figsize: tuple[int, int] = (15, 5),
-        title: str | None = None,
-        savepath: str | None = None,
+        title: str = None,
+        savepath: str = None,
         show: bool = True,
         **kwargs: Collection,
 ) -> None:
@@ -49,6 +54,8 @@ def main_sequence_plot(
         Size of the marker symbol. (default: 25)
     color: str
         Color of the marker symbol. (default: 'purple')
+    color_line: str
+        Color of the linear fit line (default: 'red')
     alpha: float
         Alpha value (=transparency) of the marker symbol. Between 0 and 1. (default: 0.5)
     marker: str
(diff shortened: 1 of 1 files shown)


src/pymovements/plotting/main_sequence_plot.py
```diff
@@ -21,21 +21,24 @@
...
 import matplotlib.pyplot as plt
+import numpy as np
 import polars as pl
 from matplotlib.collections import Collection
+from sklearn.metrics import r2_score
 
 from pymovements.events import EventDataFrame
 
 
 def main_sequence_plot(
```
dkrako, line 32: add one argument for turning the fit and the R2 on and off.

something like this:
```python
fit: bool = True,
measure: bool = True,
```

Although we could also call it line, we should keep in mind that we might want to do non-linear fits somewhen.
Having the more general name would enable us to use the argument to specify the method in the future like `fit='linear'` or `fit='exponential'`, where `fit=True` would be equivalent to `fit='linear'`.

The same holds true for the R2 measure. In contrast to what I said yesterday, R2 is not a valid measure for nonlinear fits, so here too, we could specify measures in the future, e.g. standard error of regression S.

These two blog posts are probably good inspiration for making this PR future proof for follow ups:
https://blog.minitab.com/en/adventures-in-statistics-2/why-is-there-no-r-squared-for-nonlinear-regression
https://blog.minitab.com/en/adventures-in-statistics-2/regression-analysis-how-to-interpret-s-the-standard-error-of-the-regression

src/pymovements/plotting/main_sequence_plot.py
```diff
@@ -21,21 +21,24 @@
...
-        event_df: EventDataFrame,
+    event_df: EventDataFrame,
         marker_size: float = 25,
         color: str = 'purple',
+        color_line: str = 'red',
```
dkrako, line 36: let's call the argument `fit_color`

src/pymovements/plotting/main_sequence_plot.py
```diff
@@ -21,21 +21,24 @@
...
         alpha: float = 0.5,
         marker: str = 'o',
         figsize: tuple[int, int] = (15, 5),
-        title: str | None = None,
-        savepath: str | None = None,
+        title: str = None,
```
dkrako, line 40: revert these two changes to allow for `None`

src/pymovements/plotting/main_sequence_plot.py
```diff
@@ -100,8 +105,29 @@ def main_sequence_plot(
...
 
+    a, b = np.polyfit(amplitudes, peak_velocities, 1)
+
+    # line plotting estimation
+    min_ampl = min(amplitudes)
+    max_ampl = max(amplitudes)
+    line_x = [min_ampl, max_ampl]
+    line_y = [a * min_ampl + b, a * max_ampl + b]
+
+    # residual calcualtion
```
dkrako, line 116: typo

src/pymovements/plotting/main_sequence_plot.py
```diff
@@ -21,21 +21,24 @@
 from __future__ import annotations
 
 import matplotlib.pyplot as plt
+import numpy as np
 import polars as pl
 from matplotlib.collections import Collection
+from sklearn.metrics import r2_score
```
dkrako, line 27: you need to add sklearn to the dependencies in https://github.com/aeye-lab/pymovements/blob/main/pyproject.toml

dkrako (changes requested):
Great, thanks a lot! I have just some minor change requests, mostly relating to the function signature.

The plot testing is currently very rudimentary in pymovements. You don't need to check the plots for correctness, but you'll need to cover all branches such that we at least test that there's no unexpected runtime errors.

dkrako:
Hi @ivanvajs1996 just wanted to ask if you're planning to finish this PR or if someone else should take over.

There requested changes are rather small, so it won't be much work.

Issue #875, opened 2024-10-24
Metadata records the tracked eye only once


saphjra:
The function https://github.com/aeye-lab/pymovements/blob/49802e4429fea6299f1ffbb8382af30ecc53e6d7/src/pymovements/gaze/io.py#L271 
returns a dataframe and a dict where metadata is stored.
This metadata dict has a key 'tracked_eye' and can have the values ‘L’, ‘R’ or ‘LR’

which are extracted by the following pattern: 

https://github.com/aeye-lab/pymovements/blob/502dc4410f15b38361980def59fe8ff98734c94f/src/pymovements/utils/parsing.py#L51-L56
and executed by the following code line: https://github.com/aeye-lab/pymovements/blob/502dc4410f15b38361980def59fe8ff98734c94f/src/pymovements/utils/parsing.py#L350

However, this implicitly assumes that the tracked Eye does not change during a Session. 
It would be better to have a record of the tracked eye for each trial during the Session. 
The implementation could also have effect on issue #868  and https://github.com/aeye-lab/pymovements/issues/686#:~:text=binocular%20eyelink%20parsing-,%23686,-Open 

To achieve this we can perform the matching multiple times instead of only once, and save them for example as a list, where the tracked eye for each trial is stored. 

## Description of a solution
create a separated TRACKED_EYE_REGEX 
remove it from the EYELINK_META_REGEXES
adapt : https://github.com/aeye-lab/pymovements/blob/502dc4410f15b38361980def59fe8ff98734c94f/src/pymovements/utils/parsing.py#L178

The additional information extracted can then be used for example for quality checking (is the same eye tracked during the whole experiment)

dkrako:
Further enhancements discussed in #1072

PR #767, opened 2024-07-23
feat!: Custom patterns for parsing logged metadata in ASC files

saeub:
## Description

Fixes #758

## Implemented changes

- [x] **BREAKING:** Return metadata dictionary from `from_asc()`
- [x] Add argument `matadata_patterns` to `from_asc()` and `parse_eyelink()`
- [x] Default value of `None` for metadata keys specified in `metadata_patterns` that are not present in the ASC file

## TODO

- [x] Make pylint happy
- [x] Fix test coverage

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change is or requires a documentation update

## How Has This Been Tested?

- [x] Metadata returned
- [x] Test default `None` values

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

PR #884, opened 2024-10-25
feat: Store metadata from ASC in experiment metadata

### LINKED ISSUE #868: Metadata Documentation
## Description of the problem
Currently, the metadata structure we receive after recording is poorly organized, making it difficult to interpret the various pieces of information.

## Description of a solution
We aim to document this information in a clear, structured format to ensure it is easy to read and understand 

Include steps of the proposed solution.
 -Create a class that will organize the metadata in a more structured and meticulous manner

What are the expected benefits/improvements from working on this new feature?
This feature will centralize metadata information, making it easier to access and manage.

## Sample Code

```py
>>> gazeframe, metadata = from_asc(...)
>>> metadata.tracked_eye
'right'
>>> metadata.calibrations
[...]
```

The metadata could also be part of the gazeframe (gazeframe.metadata).

saeub:
## Description

Put the metadata parsed in `parse_eyelink()` into the `Experiment`, `Screen`, and `EyeTracker` classes, and check for mismatches compared to user-provided metadata. Create additional attributes for metadata that is not exposed yet (e.g., calibrations and validations).

Fixes #868.

Related to #875.

## Implemented changes

- [x] Store metadata from ASC file in corresponding attributes of `Experiment` object
- [x] Raise error if there are conflicts with existing values in `Experiment`
- [x] Store remaining metadata in private attribute `GazeDataFrame._metadata` to revert breaking change introduced in #767

## Type of change

## How Has This Been Tested?

- [x] `test_from_asc_fills_in_experiment_metadata`
- [x] `test_from_asc_detects_mismatches_in_experiment_metadata`

## Checklist:

### DIFF
--- a/src/pymovements/datasets/toy_dataset_eyelink.py
+++ b/src/pymovements/datasets/toy_dataset_eyelink.py
@@ -166,8 +166,8 @@ class ToyDatasetEyeLink(DatasetDefinition):
         origin='upper left',
         eyetracker=EyeTracker(
             sampling_rate=1000.0,
-            left=False,
-            right=True,
+            left=True,
+            right=False,
             model='EyeLink Portable Duo',
             vendor='EyeLink',
         ),
--- a/src/pymovements/gaze/gaze_dataframe.py
+++ b/src/pymovements/gaze/gaze_dataframe.py
@@ -283,7 +283,8 @@ class GazeDataFrame:
         else:
             self.events = events.copy()
 
-        self._metadata: dict | None = None
+        # Remove this attribute once #893 is fixed
+        self._metadata: dict[str, Any] | None = None
 
     def apply(
             self,
--- a/src/pymovements/gaze/gaze_dataframe.py
+++ b/src/pymovements/gaze/gaze_dataframe.py
@@ -283,6 +283,8 @@ class GazeDataFrame:
         else:
             self.events = events.copy()
 
+        self._metadata: dict | None = None
+
     def apply(
             self,
             function: str,
--- a/src/pymovements/gaze/io.py
+++ b/src/pymovements/gaze/io.py
@@ -277,7 +277,7 @@ def from_asc(
         experiment: Experiment | None = None,
         add_columns: dict[str, str] | None = None,
         column_schema_overrides: dict[str, Any] | None = None,
-) -> tuple[GazeDataFrame, dict[str, Any]]:
+) -> GazeDataFrame:
     """Initialize a :py:class:`pymovements.gaze.gaze_dataframe.GazeDataFrame`.
 
     Parameters
@@ -303,8 +303,8 @@ def from_asc(
 
     Returns
     -------
-    tuple[GazeDataFrame, dict[str, Any]]
-        The gaze data frame and a metadata dictionary read from the asc file.
+    GazeDataFrame
+        The gaze data frame read from the asc file.
 
     Examples
     --------
@@ -312,7 +312,7 @@ def from_asc(
     We can then load the data into a ``GazeDataFrame``:
 
     >>> from pymovements.gaze.io import from_asc
-    >>> gaze, metadata = from_asc(file='tests/files/eyelink_monocular_example.asc')
+    >>> gaze = from_asc(file='tests/files/eyelink_monocular_example.asc')
     >>> gaze.frame
     shape: (16, 3)
     ┌─────────┬───────┬────────────────┐
@@ -332,7 +332,7 @@ def from_asc(
     │ 2339290 ┆ 618.0 ┆ [637.6, 531.4] │
     │ 2339291 ┆ 618.0 ┆ [637.3, 531.2] │
     └─────────┴───────┴────────────────┘
-    >>> metadata['sampling_rate']
+    >>> gaze.experiment.eye_tracker.sampling_rate
     1000.0
     """
     if isinstance(patterns, str):
@@ -343,7 +343,7 @@ def from_asc(
             raise ValueError(f"unknown pattern key '{patterns}'. Supported keys are: eyelink")
 
     # Read data.
-    gaze_data, metadata = parse_eyelink(
+    gaze_data, _ = parse_eyelink(
         file, patterns=patterns, schema=schema, metadata_patterns=metadata_patterns,
     )
 
@@ -368,7 +368,7 @@ def from_asc(
         time_unit='ms',
         pixel_columns=['x_pix', 'y_pix'],
     )
-    return gaze_df, metadata
+    return gaze_df
 
 
 def from_ipc(
--- a/src/pymovements/gaze/io.py
+++ b/src/pymovements/gaze/io.py
@@ -374,81 +374,69 @@ def from_asc(
     if experiment.eyetracker is None:
         experiment.eyetracker = EyeTracker()
 
-    # resolution compariosn with ascii file and experiment
+   # Resolution comparison with ASCII file and experiment
     if metadata['resolution'] == (experiment.screen.height_px, experiment.screen.width_px):
         pass
-    elif (experiment.screen.height_px, experiment.screen.width_px) is None:
+    if (experiment.screen.height_px, experiment.screen.width_px) is None:
         experiment.screen.height_px, experiment.screen.width_px = metadata['resolution']
 
-    else:
+    if metadata['resolution'] != (experiment.screen.height_px, experiment.screen.width_px):
         raise ValueError(
-            f"Ascii file says resolution should be this: {
-                metadata['resolution']
-            }. But resolution provided this: {
-                experiment.screen.height_px,
-                experiment.screen.width_px
-            }",
+            f"Ascii file says resolution should be this: {metadata['resolution']}. "
+            f'But resolution provided is this: '
+            f'({experiment.screen.height_px}, {experiment.screen.width_px})',
         )
 
-    # sample rate comparion between metadata and eyetracker metadata
+    # Sample rate comparison between metadata and eyetracker metadata
     if metadata['sampling_rate'] != experiment.eyetracker.sampling_rate:
         raise ValueError(
-            f"Ascii file says sampling rate should be this: {
-                metadata['sampling_rate']
-            }. But sampling rate provided this: {
-                experiment.eyetracker.sampling_rate
-            }",
+            f"Ascii file says sampling rate should be this: {metadata['sampling_rate']}. "
+            f'But sampling rate provided is this: {experiment.eyetracker.sampling_rate}',
         )
-    elif experiment.eyetracker.sampling_rate is None:
+
+    if experiment.eyetracker.sampling_rate is None:
         experiment.eyetracker.sampling_rate = metadata['sampling_rate']
 
-    else:
-        pass
+# Left Eye or Right Eye
+
     # left Eye or right Eye
     if metadata['traked_eye'] == 'R' and not experiment.eyetracker.right:
         raise ValueError(
-            f"Ascii file syas its is:{
-                metadata['traked_eye']
-            }. But eye was used by experiment is :{
-                experiment.eyetracker.right
-            }",
+            f"Ascii file says it is: {metadata['tracked_eye']}. "
+            f'But the eye used by the experiment is: {experiment.eyetracker.right}',
         )
-    elif experiment.eyetracker.right is None:
+
+    if experiment.eyetracker.right is None:
         experiment.eyetracker.right = metadata['traked_eye']
 
     if metadata['traked_eye'] == 'L' and not experiment.eyetracker.left:
         raise ValueError(
-            f"Ascii file syas its is:{
-                metadata['traked_eye']
-            }. But eye was used by experiment is :{
-                experiment.eyetracker.left
-            }",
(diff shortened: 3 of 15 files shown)


src/pymovements/gaze/io.py
```diff
@@ -360,6 +362,98 @@ def from_asc(
             for fileinfo_key, fileinfo_dtype in column_schema_overrides.items()
         ])
 
+    if experiment is None:
+        experiment = Experiment(sampling_rate=metadata['sampling_rate'])
+    if experiment.screen is None:
```
dkrako, line 367: Actually, `Experiment.screen` can never be `None` until someone tampered with it. That's why this line isn't covered and you would need to create a bit of an artificial test where you tamper with the `screen` value.

`Experiment.screen` is set right at the start of `Experiment.__init__(...)`:

https://github.com/aeye-lab/pymovements/blob/4042879f4577b68b08c197095ae1a413c332311c/src/pymovements/gaze/experiment.py#L98-L105

Instead I think it's probably best to remove this if clause completely and handle the edge case in a follow-up PR: the `Experiment` class can be improved by implementing the `screen` field as a property. This way we can guarantee its integrity (being of type `Screen`).

src/pymovements/gaze/io.py
```diff
@@ -360,6 +362,98 @@ def from_asc(
...
+        experiment.screen = Screen()
+    if experiment.eyetracker is None:
```
dkrako, line 369: Good catch! You actually exposed a type ambivalence in the `Experiment.eyetracker` field here. In contrast to the `screen` field, the `eyetracker` field can be `None`. In this case you should keep the if-clause in this PR, as the current implementation in `Experiment.eyetracker` is indeed type-ambivalent.

In a follow-up the `eyetracker` field could be improved in a similar way as I suggested with the `screen` field. Feel free to create a short initial issue for this where we can develop that idea further.

src/pymovements/gaze/io.py
```diff
@@ -360,6 +362,98 @@ def from_asc(
...
+        experiment.eyetracker = EyeTracker()
+
+    # Resolution comparison with ASCII file and experiment
+
+    if experiment.screen.width_px is None or experiment.screen.height_px is None:
```
dkrako, line 374: leave a comment that you assume that width and height are always missing mutually.

src/pymovements/gaze/io.py
```diff
@@ -360,6 +362,98 @@ def from_asc(
...
+        experiment.screen.width_px, experiment.screen.height_px = metadata['resolution']
+
+    if metadata['resolution'] != (experiment.screen.width_px, experiment.screen.height_px):
+        raise ValueError(
```
dkrako, line 378: Great that you check for consistency here.

I think it's ok to be pretty restrictive in this PR, but maybe it's a good idea to raise a warning instead in the future.

As there's quite a lot copy-paste in the following lines it would be nice if you could create a simple private helper function like this one:

```python
def _raise_inconsistent_values(value1: Any, value2: Any, name: str) -> None:
```
The error string can be then parametrized by the input arguments.

src/pymovements/gaze/io.py
```diff
@@ -360,6 +362,98 @@ def from_asc(
...
+        experiment.eyetracker.mount = metadata['mount_configuration']['mount_type']
+
+    if metadata['mount_configuration']['mount_type'] != experiment.eyetracker.mount:
+        raise ValueError(
+            f'Ascii file says mount config should be this: '
+            f"{metadata['mount_configuration']['mount_type']}. "
+            f'But mount config provided is: {experiment.eyetracker.mount}',
+        )
+
+    # model check git
```
dkrako, line 426: a wild `git` appeared!

tests/unit/gaze/io/asc_test.py
```diff
@@ -192,3 +156,16 @@ def test_from_asc_raises_exception(kwargs, exception, message):
 
     msg, = excinfo.value.args
     assert msg == message
+
+
+def test_from_asc_fills_in_experiment_metadata():
+    gaze = pm.gaze.from_asc('tests/files/eyelink_monocular_example.asc', experiment=None)
```
dkrako, line 162: can you parametrize this and check also for `eyelink_monocular_2khz_example.asc`?
at least the sampling_rate would differ.

dkrako (dismissed):
Great, thank you so much for your work! I have some minor change requests, but apart from these the PR is in very good shape.

I would like to include this PR in our upcoming release. This way we avoid the breaking change in the `from_asc()` interface.
@SiQube @prassepaul What do you think of further delaying the release until this is merged?

dkrako:
Oh, my review did not take account of the changes you just pushed an hour ago. I hope this is not too confusing.

src/pymovements/gaze/io.py
```diff
@@ -360,6 +362,91 @@ def from_asc(
...
+        issues.append(
+            f"Eye tracker software version: {
+                experiment.eyetracker.version
+            } vs. {
+                metadata['version_number']
+            }",
+        )
+
+    if issues:
+        raise ValueError(
```
dkrako, line 445: that's also a good way to get rid of the copy-paste code. You can ignore my previous comment on the private helper function then.

saeub:
@dkrako Thanks! Working on the changes now.

Originally, we also wanted to store the calibrations and validations in the `GazeDataFrame` somehow (which is the main thing still missing if `metadata` is not returned from `from_asc()`). Should I create another PR for that, so we can merge this one sooner?

dkrako:
> @dkrako Thanks! Yes, it seems like we were working in parallel :D Working on the remaining changes now.
> 
> Originally, we also wanted to store the calibrations and validations in the `GazeDataFrame` somehow (which is the main thing still missing if `metadata` is not returned from `from_asc()`). Should I create another PR for that, so we can merge this one sooner?

Yes let's store the calibrations and validations in a follow-up PR.

As the `metadata` is not returned anymore, you could add a simple line in the `from_asc()` method right before you return the `gaze_df`, where you store the metadata in a undocumented private field, for example like this:

```python
gaze_df._metadata = metadata
```

This way a (both very informed and impatient) user would still be able to use `from_asc()` and get the metadata.
It's a bit hacky, but that would only be a temporary solution and which would be improved in the follow up PR.
I would keep the field private and undocumented such that we can silently remove it in the next version.

src/pymovements/gaze/io.py
saeub: #891

src/pymovements/gaze/io.py
```diff
@@ -360,6 +361,75 @@ def from_asc(
...
+        experiment.eyetracker.mount = metadata['mount_configuration']['mount_type']
+    elif experiment.eyetracker.mount != metadata['mount_configuration']['mount_type']:
+        issues.append(f"Mount configuration: {experiment.eyetracker.mount} vs. "
+                      f"{metadata['mount_configuration']['mount_type']}")
+
+    # Eye tracker vendor
+    asc_vendor = 'EyeLink' if 'EyeLink' in metadata['model'] else None
+    if experiment.eyetracker.vendor is None:
+        experiment.eyetracker.vendor = asc_vendor
+    elif experiment.eyetracker.vendor != asc_vendor:
```
saeub, line 411: Not sure why all of these `elif` lines get flagged as "partial coverage" :thinking: If I change them to `if`, they are flagged as fully covered, but conceptually I think `elif` makes more sense?

saeub: @dkrako Any idea about why the coverage isn't passing here? Otherwise, from my side, I think the PR is ready.

src/pymovements/gaze/io.py
```diff
@@ -360,6 +361,75 @@ def from_asc(
...
+        experiment.eyetracker = EyeTracker()
+
+    # Compare metadata from experiment definition with metadata from ASC file.
+    # Fill in missing metadata in experiment definition and raise an error if there are conflicts
+    issues = []
+
+    # Screen resolution (assuming that width and height will always be missing or set together)
+    experiment_resolution = (experiment.screen.width_px, experiment.screen.height_px)
+    if experiment_resolution == (None, None):
+        experiment.screen.width_px, experiment.screen.height_px = metadata['resolution']
```
SiQube, line 376: is resolution always within ascii?

src/pymovements/gaze/io.py
```diff
@@ -360,6 +361,75 @@ def from_asc(
...
+    if experiment.eyetracker.sampling_rate is None:
+        experiment.eyetracker.sampling_rate = metadata['sampling_rate']
+    elif experiment.eyetracker.sampling_rate != metadata['sampling_rate']:
+        issues.append(
+            f"Sampling rate: {experiment.eyetracker.sampling_rate} vs. {metadata['sampling_rate']}",
+        )
+
+    # Tracked eye
+    asc_left_eye = 'L' in metadata['tracked_eye']
+    asc_right_eye = 'R' in metadata['tracked_eye']
```
SiQube, line 390: is it always `L` vs `R`? is `tracked_eye` always recorded?

src/pymovements/gaze/io.py
```diff
@@ -368,7 +438,8 @@ def from_asc(
         time_unit='ms',
         pixel_columns=['x_pix', 'y_pix'],
     )
-    return gaze_df, metadata
+    gaze_df._metadata = metadata  # pylint: disable=protected-access
```
SiQube, line 441: why are we calling it `_metadata` and not just `metadata`?CC @dkrako

tests/unit/dataset/dataset_files_test.py
```diff
@@ -206,7 +206,7 @@ def test_load_eyelink_file(tmp_path, read_kwargs):
         filepath,
         fileinfo_row={},
         definition=DatasetDefinition(
-            experiment=pm.Experiment(1024, 768, 38, 30, None, 'center', 100),
+            experiment=pm.Experiment(1280, 1024, 38, 30, None, 'center', 100),
```
SiQube, line 209: why did this change?

tests/unit/gaze/io/asc_test.py
```diff
@@ -131,64 +131,128 @@
...
+            id='unknown_pattern',
+        ),
+    ],
+)
+def test_from_asc_raises_exception(kwargs, exception, message):
+    with pytest.raises(exception) as excinfo:
+        pm.gaze.from_asc(**kwargs)
+
+    msg, = excinfo.value.args
+    assert msg == message
```
SiQube, line 158: this seems to be only one test so we do not need parametrize

SiQube (commented):
only some minor comments and questions

src/pymovements/datasets/toy_dataset_eyelink.py
```diff
@@ -166,8 +166,8 @@ class ToyDatasetEyeLink(DatasetDefinition):
         origin='upper left',
         eyetracker=EyeTracker(
             sampling_rate=1000.0,
-            left=False,
-            right=True,
+            left=True,
+            right=False,
```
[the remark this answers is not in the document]
saeub: It was always left-tracked in the ASC file, but since the mismatch wasn't tested until now, the tests never failed until now.

src/pymovements/gaze/gaze_dataframe.py
```diff
@@ -283,6 +283,8 @@ def __init__(
         else:
             self.events = events.copy()
 
+        self._metadata: dict | None = None
```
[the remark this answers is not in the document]
saeub: It's actually `dict[str, Any]` (there is a lot of arbitrarily nested data in there)

src/pymovements/gaze/io.py
saeub: I think it's either `L`, `R`, or `LR` (binocular). Recorded in the same way as sampling rate.

src/pymovements/gaze/io.py
saeub: We're planning on removing the private attribute as soon as we found a better way to store the remaining metadata (calibrations, validations, changing sampling rate / tracked eye). See https://github.com/aeye-lab/pymovements/pull/884#issuecomment-2465477675

tests/unit/dataset/dataset_files_test.py
saeub: Same as above. It's different from the resolution logged in the ASC file, but up to now it didn't matter because the mismatch wasn't checked.

tests/unit/gaze/io/asc_test.py
saeub: I did not touch this test :sweat_smile: (the diff is a bit messed up here...)

tests/unit/dataset/dataset_files_test.py
SiQube: perfect

SiQube (approved):

saeub:
@SiQube I don't understand how this could happen, but it seems like merging #897 broke the code coverage [here](https://app.codecov.io/gh/aeye-lab/pymovements/pull/884/blob/src/pymovements/utils/archives.py#L153)? Any ideas?

SiQube:
no its a stupid bug the 0.06 come from the <py3.12 >=py3.12 ...

saeub:
Looks like upgrading and downgrading the codecov action did the trick :sweat_smile:

PR #885, opened 2024-10-25
:mushroom: events moved to gaze df

senisioi:
## Description

- issue 871: https://github.com/aeye-lab/pymovements/issues/871 

Fixes #

## Implemented changes

- [] added the functionality to compute event properties at the gaze datafram level
- [] raised an error if the gaze object did not run detect events before
- [] added logging messages for `verbose` parameter to make sense 

## Type of change

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,7 +49,7 @@ dynamic = ["version"]
 docs = [
   "ipykernel>=6.13.0",
   "nbconvert>=7.0.0,<7.17",
-  "nbsphinx>=0.8.8,<0.9.6",
+  "nbsphinx>=0.8.8,<0.10",
   "pandoc",
   "pybtex",
   "pydata-sphinx-theme>=0.12",


src/pymovements/__init__.py
```diff
@@ -79,3 +81,7 @@
 ]
 
 __version__ = _version.get_versions()['version']
+
+LOG_LEVEL = os.environ.get('LOG_LEVEL', 'INFO')
+LOG_CONFIG = '[%(levelname)s] %(asctime)s %(name)s:%(lineno)d - %(message)s'
+logging.basicConfig(level=LOG_LEVEL, format=LOG_CONFIG)
```
dkrako, line 87: I needed a bit of time to think about this.

we probably shouldn't set this in the package, as a user might have a custom logging config defined. after importing pymovements their config would be overwritten.

this logic should actually be on the side of the user, so I would suggest to revert the changes in this file.

still, defining logging levels and other configuration aspects can form a small mini tutorial. this would be out of scope of this PR but feel free to create one in case you like.

src/pymovements/gaze/gaze_dataframe.py
```diff
@@ -883,6 +889,55 @@ def detect(
                 how='diagonal',
             )
 
+    def compute_event_properties(
+            self,
+            event_properties: str | tuple[str, dict[str, Any]]
+            | list[str | tuple[str, dict[str, Any]]],
+            name: str | None = None,
+            verbose: bool = True,
+    ) -> pm.EventDataFrame:
```
dkrako, line 898: there are inconsistencies between the return types of `Dataset` and `GazeDataFrame` methods.
`Dataset` inplace methods return `Dataset`, the one from `GazeDataFrame` return `None`.

We will resolve this inconsistency sometime in the future when we decide which inplace return type we want to favor.

For the time being, it would be best to stay as consistent as possible with the particular modules to not introduce any further confusion, i.e. I would prefer returning `None` here

src/pymovements/gaze/gaze_dataframe.py
```diff
@@ -883,6 +889,55 @@ def detect(
...
+        pm.EventDataFrame
+            Returns self, useful for method cascading.
+        """
+        if self.events is None:
+            raise RuntimeError(
+                'No events available to compute event properties. '
+                'Consider calling detect before.',
+            )
+
+        if verbose is True:
```
dkrako, line 929: To log the debug message one would have to set `verbose` to `True` and set the logging level to `DEBUG`.

So probably the verbose parameter isn't needed after all in this method and the if condition can be updated to
 ```python
if logger.isEnabledFor(logging.DEBUG):
```
to avoid creating the debug string if not needed

dkrako (changes requested):
Thank you so much! This PR is already in great shape!

There are some small quirks left to work out, especially regarding the logging/verbose parts.

src/pymovements/__init__.py
senisioi: indeed, will revert
although I'm not sure it's overwriting the user logging config, this should be only at the package level; but I'll have to check; on the long run probably a `logging.ini` would be more appropriate

src/pymovements/gaze/gaze_dataframe.py
senisioi: the verbose thing was there just to be consistent with the Dataset method, but maybe we can remove it together with the logging code above and keep it simple;
the message is not very informative either as it shows the same thing over and over again when the method is called from the Dataset

src/pymovements/gaze/gaze_dataframe.py
senisioi: haha, okay, I didn't notice that; yes let's keep inconsistencies consistent

Issue #453, opened 2023-06-13
Refactor list columns into struct columns


dkrako:
## Description of the problem
In the current state, we always have to now a magic number (the index) to get a specific component.

This then introduces some issues:

* the requirement of a correct order of components
* exploding needs to pass the component types
* when specifying a specific eye, we need to to map eye to components
* we need to carry a special `n_components` variable from expressions to expression

## Description of a solution
In addition to the `pl.List` dtype, polars also provides the `pl.Struct` dtype, where we have access to named fields:

https://pola-rs.github.io/polars/py-polars/html/reference/expressions/api/polars.struct.html#polars.struct
https://pola-rs.github.io/polars/py-polars/html/reference/series/struct.html

This could also be used to get rid of the `n_components` variable we are currently carrying around

## Minimum acceptance criteria
- [ ] `pixel`, `position`, `velocity`, and `acceleration` are all changed to 'pl.Struct`

dkrako:
Having components as named components in a `pl.Struct` would simplify writing a polars expression like `eye()`, that returns a `pl.String` (`L` left, `R` right, `M` monocular unknown eye, `B` or `LR` binocular, `C` cyclopean, `LCR` left right cyclopean?)

the expression could be simply applied to a column via `gaze.frame.position.eyes()` or `gaze.frame.position.gaze.eyes()`

dkrako:
This could also allow for much more intuitive and flexible component specification:

```
pixel_columns:
    x: x_pix
    y: y_pix
```

or

```
pixel_columns:
   xr: x_right
   yr: y_right
   xl: x_left
   yl: y_left
```

PR #887, opened 2024-10-25
feat: add support for multiple recording specs per file to gaze.from_asc()

saphjra:
## Description

Record all the tracked eye sides in the metadata to fix the issue  #875

## Implemented changes

I modified the code in parsing.py to record every time it can be matched in the message line of the asci files

- Added Regex for Tracking Eye Information in the asci file
- added a condition to match the pattern multiple times and add it to a list
- created an additional  pre_processed_metadata key : tracked_eyes, where the list is stored

## Type of change

## How Has This Been Tested?
 I ran it on the to ascii converted ch1hr007.edf of the pilot-hr-1-zh data to verify the output, but no further testing has been done so far

## Checklist:


src/pymovements/utils/parsing.py
```diff
@@ -93,6 +92,11 @@
     r'END\s+(?P<timestamp>(\d+[.]?\d*))\s+\s+(?P<types>.*)\s+RES\s+'
     r'(?P<xres>[\d\.]*)\s+(?P<yres>[\d\.]*)\s*',
 )
+TRACKED_EYE_REGEX = re.compile(
```
dkrako, line 95: It actually looks like the overall tracking configuration can change, not just the tracked eye.

So instead of just reparsing `tracked_eye`, also reparse the remaining fields (`tracking_mode`, `sampling_rate`, `file_sample_filter`, `link_sample_filter`)

Then you should probably rename the regex to something like `TRACKING_CONFIG_REGEX`.

src/pymovements/utils/parsing.py
```diff
@@ -247,7 +251,7 @@ def parse_eyelink(
     calibrations = []
     blinks = []
     invalid_samples = []
-
+    tracked_eyes = []
```
dkrako, line 254: We should then probably call this `tracking_configs`.

src/pymovements/utils/parsing.py
```diff
@@ -371,6 +378,7 @@ def parse_eyelink(
     pre_processed_metadata['data_loss_ratio'] = data_loss_ratio
     pre_processed_metadata['data_loss_ratio_blinks'] = data_loss_ratio_blinks
     pre_processed_metadata['total_recording_duration_ms'] = total_recording_duration
+    pre_processed_metadata['tracked_eyes'] = tracked_eyes
```
dkrako, line 381: The dictionary field should then also be renamed to `tracking_configs`.

Also make sure that the first match from `EYELINK_META_REGEXES` is included too. Otherwise the entry will be empty if there were no changes in tracking configuration.

dkrako (changes requested):
Great, thanks a lot for the submission!

I would like to increase the scope of this PR to not only account for changes of tracked eyes but also changes of the overall recording configuration. They are all included in the same line and can be matched with a single regex.

Apart from this, please put tests into https://github.com/aeye-lab/pymovements/blob/feature/metadata/tests/unit/utils/parsing_test.py

You can test it two ways: either create a new test file with changing configs or just specify a string like it is done for `ASC_TEXT` to create a file on the fly.

tests/unit/utils/parsing_test.py
```diff
@@ -584,9 +627,10 @@ def test_parse_eyelink_blinks(tmp_path, metadata, expected_blinks):
             'END	10000024 	SAMPLES	EVENTS	RES	  38.54	  31.12\n',
             0.25,
             0.75,
-            id='missing_timestamps_lost_samples',
+            id='missing_timestamps_lost_samples4',
         ),
         pytest.param(
+            '** DATE: Wed Mar  8 09:25:20 2023\n'
```
dkrako, line 633: these added date strings aren't really necessary, right?
It really doesn't matter in this case and you don't need to revert them, but usually I would advise to avoid changes to existing test logic, e.g. changing test values.

Your other changes here, like adding documentation or changing test ids, are of course the spirit that we need! :1st_place_medal:

dkrako (changes requested):

dkrako (changes requested):
Great, thanks a lot for your update!

Apart from the comments I left, there is the issue that #884 is now conflicting with your PR. We will probably merge #884 first, as it's close to finished.

We will then need to work out how to integrate this PR here into the new functionality from that PR.

I would probably suggest to do something similar as it's done with the screen distance, which can be statically defined in `Experiment`, but can also be dynamically defined as a column in `GazeDataFrame`.

Let's think about how to work this issue out and discuss that next week.
Until then, the work on the other comments should be straight forward.

tests/unit/utils/parsing_test.py
saphjra: Thanks for all your Feedback 😄 

The thing is, without this additional line, the metadata will be empty, since the MSG line gets parsed into the recording_config and not metadata anymore. However if metadata is empty/ none the code will raise a warning and all the tests, where I added the Date line,  will fail due to that. So I figured, I add a line that should be present in any dataset, which gets parsed by the metadata. 

Is there a better way, to solve this?

dkrako:
Hi @saphjra I just noticed that you're probably waiting on some further feedback on this PR right?
I won't be able to review this week but I try to have a go at it next week so we can finish this up in the next two weeks. Thank you for your patience.

tests/unit/utils/parsing_test.py
dkrako: You can leave the date strings in the tests. I have created the issue #944 such that parsing continues after printing (instead of raising) a warning.

src/pymovements/utils/parsing.py
```diff
@@ -433,12 +433,41 @@ def _pre_process_metadata(metadata: defaultdict[str, Any]) -> dict[str, Any]:
...
+        prints a warning if no recording configuration is found or
+        if the sampling rate is inconsistent.
+    """
+    if not recording_config:
+        sampling_rate = None
+        print('Warning: No recording configuration found. Cannot calculate data loss.')
+    else:
+        sampling_rates = {d.get('sampling_rate') for d in recording_config}
+        if len(sampling_rates) != 1:
+            print(
```
dkrako, line 457: instead of using `print()` pleasue use `warnings.warn()`

src/pymovements/utils/parsing.py
```diff
@@ -349,19 +352,20 @@ def parse_eyelink(
                     # each metadata pattern should only match once
                     compiled_metadata_patterns.remove(pattern_dict)
 
-    if not metadata:
-        raise Warning('No metadata found. Please check the file for errors.')
-
     # if the sampling rate is not found, we cannot calculate the data loss
     actual_number_of_samples = len(samples['time'])
+    # if we don't have any recording config, we cannot calculate the data loss
+    sampling_rate = _check_sampling_rate(recording_config)
```
dkrako, line 358: please assign this to `metadata['sampling_rate']` for backwards-compatibility

src/pymovements/gaze/io.py
```diff
@@ -332,7 +332,7 @@ def from_asc(
     │ 2339290 ┆ 618.0 ┆ [637.6, 531.4] │
     │ 2339291 ┆ 618.0 ┆ [637.3, 531.2] │
     └─────────┴───────┴────────────────┘
-    >>> metadata['sampling_rate']
+    >>> float(metadata["recording_config"][0]['sampling_rate'])
```
dkrako: this line should be the same as the one in the main branch:
https://github.com/aeye-lab/pymovements/blob/fd01f61a46793ebf9090042060be8d0ba0316b33/src/pymovements/gaze/io.py#L342

dkrako (changes requested):
Thanks again for your work and patience!

I resolved the merge conflicts on your branch.

One thing that remains is how to handle the changes in tracked eye in `gaze.from_asc()`. 

Currently, it is passed to the `Experiment` object:

https://github.com/aeye-lab/pymovements/blob/fd01f61a46793ebf9090042060be8d0ba0316b33/src/pymovements/gaze/io.py#L394-L404

this needs to be adjusted for the case of changing the tracked eye during the experiment. I would propose to do it in a similar way as it's done for `distance`.

If the tracked eye changes, then set `experiment.eye.tracker.left` and `experiment.eye.tracker.right` to `None`. Additionally, create a column in the dataframe named `eye` which holds either `L`, `R` or `LR` (or `None`?) for each timestep.

src/pymovements/utils/parsing.py
dkrako: I changed my mind, backwards compatibility isn't really needed for that field as its now passed to `Experiment` in `from_asc()`.

src/pymovements/gaze/io.py
dkrako: Ah I see this was already resolved by my merge commit

saeub:
@dkrako I resolved the merge conflicts and am now continuing this (with @saphjra's consent :slightly_smiling_face:)

Some updates:
- My recent changes also fix #944.
- I tried to remove unnecessary metadata (like dates) from parsing unit tests and instead used `@pytest.mark.filterwarnings` to ignore the warnings about missing metadata.
- You mentioned at some point that the tracked eye could be added as a column in the gaze dataframe, just like the distance column. However, it looks like there is already a way of specifying the tracked eye using [nested (multi-component) columns](https://github.com/aeye-lab/pymovements/blob/09b1ee6184e346dda52be485dd4922277bab1d58/src/pymovements/gaze/gaze_dataframe.py#L1320), so it seems unnecessary to also add a separate column for the tracked eye. Or am I missing something?

saeub:
> Just so that we don't forget, can you create an issue with your ideas before merging?

I created #1072. I agree that #453 would also make this much more usable.

Just out of curiosity, I was wondering if you also considered the following option when implementing the multi-components feature: Instead of allowing multiple components in the same column, adding a column `eye` in `GazeDataFrame`; in the binocular case, we would have two rows for each timestamp, one for the left and one for the right eye, like this:

| time | pixel | eye |
| --- | --- | --- |
| 1 | [123, 456] | L |
| 1 | [123, 456] | R |
| 2 | [123, 456] | L |
| 3 | [123, 456] | R |

This would make filtering for a specific eye or applying event detection to a specific eye very easy. (Although it might make it difficult to notice whether the tracked eye changed between trials.)

dkrako:
> > Just so that we don't forget, can you create an issue with your ideas before merging?
> 
> I created #1072. I agree that #453 would also make this much more usable.
> 
> Just out of curiosity, I was wondering if you also considered the following option when implementing the multi-components feature: Instead of allowing multiple components in the same column, adding a column `eye` in `GazeDataFrame`; in the binocular case, we would have two rows for each timestamp, one for the left and one for the right eye, like this:
> time 	pixel 	eye
> 1 	[123, 456] 	L
> 1 	[123, 456] 	R
> 2 	[123, 456] 	L
> 3 	[123, 456] 	R
> 
> This would make filtering for a specific eye or applying event detection to a specific eye very easy. (Although it might make it difficult to notice whether the tracked eye changed between trials.)

A dedicated column that encodes the eye was my original thought for this PR as it's the easiest to implement.
I'm ok with this PR not fully integrating changes of recording specs by writing them into columns. That can be implemented later on.

But generally, instead of having a dedicated column, it would be much nicer to have an expression. A column would need to be checked or recalculated to maintain consistency. An expression that returns the tracked eye from a specific column would be dynamic and computed on-demand. // see https://github.com/aeye-lab/pymovements/issues/453#issuecomment-2787339168

dkrako (changes requested):
Great! Just a small change request for the warning message, otherwise we're good to go

We will improve further via these issues:

- #893
- #907
- #1072
- #1082

dkrako (approved):
Great! Thank you both for taking care of this!

PR #892, opened 2024-11-09
Autofix: Avoid type ambivalence in Experiment.screen and Experiment.eyetracker

lattaai5:
Implemented read-only properties for Experiment.screen and Experiment.eyetracker to ensure they are always of the correct type. Removed unnecessary None check in the __init__ method. 
> [!CAUTION]
> Disclaimer: This fix was created by Latta AI and you should never merge before you check the correctness of generated code!
>
> The fix provided by Latta AI might not be complete and it can serve as an inspiration.

---
This bug was fixed for free by Latta AI - https://latta.ai/ourmission

If you no longer want Latta AI to attempt fixing issues on your repository, you can block this account.

### DIFF
--- a/src/pymovements/gaze/experiment.py
+++ b/src/pymovements/gaze/experiment.py
@@ -95,7 +95,7 @@ class Experiment:
             sampling_rate: float | None = None,
             eyetracker: EyeTracker | None = None,
     ):
-        self.screen = Screen(
+        self._screen = Screen(
             width_px=screen_width_px,
             height_px=screen_height_px,
             width_cm=screen_width_cm,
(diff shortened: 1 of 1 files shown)


saeub:
THE ROBOTS ARE TRYING TO STEAL OUR JOBS!!1!1!!

dkrako:
wow! how did that happen?

actually I don't think that we need to be that restrictive.
I guess the property should be writable, we just need to assure that it's of correct type, and if it's `None`, then create a default empty instance of the type.

@saeub what do you think?

src/pymovements/gaze/experiment.py
```diff
@@ -106,15 +106,25 @@ def __init__(
...
         checks.check_is_mutual_exclusive(sampling_rate=sampling_rate, eyetracker=eyetracker)
 
-        self.eyetracker = eyetracker
+        self._eyetracker = eyetracker
 
         self._sampling_rate = sampling_rate
 
         checks.check_is_not_none(sampling_rate=self.sampling_rate)
-        assert self.sampling_rate is not None
+        checks.check_is_not_none(sampling_rate=self.sampling_rate)
```
dkrako, line 114: @lattaai5 revert this change

dkrako (changes requested):

src/pymovements/gaze/experiment.py
dkrako: guess that didn't work 😅

src/pymovements/gaze/experiment.py
```diff
@@ -106,15 +106,25 @@ def __init__(
...
 
         checks.check_is_greater_than_zero(sampling_rate=self.sampling_rate)
 
+    @property
+    def screen(self) -> Screen:
+        """Get the Screen object for the experiment."""
+        return self._screen
+
+    @property
+    def eyetracker(self) -> EyeTracker | None:
```
dkrako, line 124: this is still type ambivalent, it should always return an `EyeTracker` object

dkrako (changes requested):

src/pymovements/gaze/experiment.py
```diff
@@ -106,15 +106,25 @@ def __init__(
 
         checks.check_is_mutual_exclusive(sampling_rate=sampling_rate, eyetracker=eyetracker)
 
-        self.eyetracker = eyetracker
+        self._eyetracker = eyetracker
```
dkrako, line 109: if `eyetracker` is `None` then create a default empty `EyeTracker` instance

dkrako (changes requested):

PR #894, opened 2024-11-10
fixed fakenews download link

aarbeikop:
Adjusted the download link for FakeNews & corrected polars exception


SiQube:
refactored in #897

PR #901, opened 2024-11-28
docs: add CITATION.cff

SiQube:
github enabled a feature where you can add bibtex files using `CITATION.cff`


dkrako (approved):
Nice catch! Thanks a lot! :)

PR #906, opened 2024-12-11
feat: add center_offset parameter to pix2deg and center_origin

### LINKED ISSUE #534: Add center offset argument to pix2deg()
## Description of the problem
pix2deg() currently assumes that the eye is projected orthogonal to the center of the screen.

This is not always the case and so an offset needs to be specified.

OmerShubi:
## Description

Fixes #534 - Adding center_offset to center_origin function and propagating it to pix2deg.

## Implemented changes

The offset is implemented in the center_origin() function in pymovements.gaze.transforms.

The offset should be given in centimeters and is expected to be a tuple of floats with a zero default (center_offset: tuple[float, float] = (0, 0)). A positive value moves the origin to the left and down respectively.

## Type of change

## How Has This Been Tested?

 - test offset in tests/unit/gaze/transforms/center_origin_test.py
 - test offset in tests/unit/gaze/transforms/pix2deg_test.py
 -  test offset in tests/unit/gaze/gaze_transform_test.py

In all cases two additional tests where created, one where the origin is top left and the other center. An offset of (10,10) is applied.

## Checklist:

### DIFF
--- a/src/pymovements/gaze/transforms.py
+++ b/src/pymovements/gaze/transforms.py
@@ -20,10 +20,7 @@
 """Module for py:func:`pymovements.gaze.transforms."""
 from __future__ import annotations
 
-try:
-    from collections.abc import Callable  # noqa
-except ImportError:
-    from collections import Callable  # noqa
+from collections.abc import Callable
 from functools import partial
 from typing import Any
 from typing import TypeVar
--- a/src/pymovements/gaze/transforms.py
+++ b/src/pymovements/gaze/transforms.py
@@ -114,6 +114,7 @@ def center_origin(
         n_components: int,
         pixel_column: str = 'pixel',
         output_column: str | None = None,
+        center_offset: tuple[float, float] = (0, 0),
 ) -> pl.Expr:
     """Center pixel data.
 
@@ -131,6 +132,8 @@ def center_origin(
         Name of the input column with pixel data. (default: 'pixel')
     output_column: str | None
         Name of the output column with centered pixel data. (default: None)
+    center_offset: tuple[float, float]
+        Offset of the eye from the center in centimeters. (default: (0, 0))
 
     Returns
     -------
@@ -157,7 +160,8 @@ def center_origin(
 
     centered_pixels = pl.concat_list(
         [
-            pl.col(pixel_column).list.get(component) - origin_offset[component % 2]
+            pl.col(pixel_column).list.get(component) -
+            origin_offset[component % 2] - center_offset[component % 2]
             for component in range(n_components)
         ],
     ).alias(output_column)
@@ -224,6 +228,7 @@ def pix2deg(
         n_components: int,
         pixel_column: str = 'pixel',
         position_column: str = 'position',
+        center_offset: tuple[float, float] = (0, 0),
 ) -> pl.Expr:
     """Convert pixel screen coordinates to degrees of visual angle.
 
@@ -246,6 +251,8 @@ def pix2deg(
         The input pixel column name. (default: 'pixel')
     position_column: str
         The output position column name. (default: 'position')
+    center_offset: tuple[float, float]
+        Offset of the eye from the center in centimeters. (default: (0, 0))
 
     Returns
     -------
@@ -260,6 +267,7 @@ def pix2deg(
         origin=origin,
         n_components=n_components,
         pixel_column=pixel_column,
+        center_offset=center_offset,
     )
 
     if isinstance(distance, (float, int)):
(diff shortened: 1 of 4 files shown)


tests/unit/gaze/gaze_transform_test.py
```diff
@@ -791,6 +791,82 @@ def fixture_experiment():
...
+                        'y_pix': [10.0],
+                        'x_dva': [0.0],
+                        'y_dva': [-26.3354],
+                    },
+                ),
+                pixel_columns=['x_pix', 'y_pix'],
+                position_columns=['x_dva', 'y_dva'],
+            ),
+            id='pix2deg_center_offset_upper_left_origin',
+        ),
```
dkrako, line 869: please add test cases for `center_origin` (one for lowerleft and center)

tests/unit/gaze/transforms/pix2deg_test.py
```diff
@@ -499,6 +499,37 @@ def test_pix2deg_raises_error(kwargs, series, exception, msg_substrings):
...
+        pytest.param(
+            {
+                'screen_resolution': (100, 100),
+                'screen_size': (100, 100),
+                'distance': 100,
+                'origin': 'center',
+                'pixel_column': 'pixel',
+                'position_column': 'position',
+                'n_components': 2,
+                'center_offset': (10, 10),
```
dkrako, line 511: please add at least two tests with different center_offset values

src/pymovements/gaze/transforms.py
```diff
@@ -131,6 +132,9 @@ def center_origin(
         Name of the input column with pixel data. (default: 'pixel')
     output_column: str | None
         Name of the output column with centered pixel data. (default: None)
+    center_offset: tuple[float, float]
+        Offset of the eye from the center in centimeters.
+        A positive value moves the origin to the left or down. (default: (0, 0))
```
dkrako, line 137: I would actually expect the origin to move right or down.

Maybe I misunderstand something, so I'll try to recap the logic of `center_origin`: in screen coordinates we implicitly assume that pixels are counted from left to right, and from top to bottom.

Let's start with the trivial case where the pixel origin is in the center of the screen and a pixel is a cm wide:

```python
Experiment(
	sampling_rate=1000,
	screen_width_px=100,
	screen_height_px=100,
	screen_width_cm=100,
	screen_height_cm=100,
	distance_cm=100,
	origin='center',
)
```

Specifically this means that the foot of the perpendicular from the eye to the screen plane is located exactly at the pixel origin.  The eye-center, left-top, right-top, left-bottom, right-bottom result in these pixel coordinates:
`(0, 0), (-50, 50), (50, 50), (-50, -50), (50, -50)`

Applying `center_origin` will not have any effect on these coordinates.

When we use `center_origin=(10, 0)` the foot of the perpendicular from the eye to the screen plane is located 10 cm right to the screen center. As a pixel is a centimeter wide it's easy to transform. Applying `center_origin` to the list of coordinates results in:
`(-10, 0), (-60, 50), (40, 50), (-60, -50), (40, -50)`

If we would use `origin='upper left'` without `center_origin` the eye-center and screen edge coordinates are:
`(49.5, 49.5), (0, 0), (100, 0), (0, 100), (100, 100)`

Applying `center_origin` will result in the pixel coordinates of the first case:
 `(0, 0), (-50, 50), (50, 50), (-50, -50), (50, -50)`

Here again, when we use `center_origin=(10, 0)` the foot of the perpendicular from the eye to the screen plane is located 10 cm right to the screen center. `center_origin` to the list of coordinates results in:
`(0, 0), (-60, 50), (40, 50), (-60, -50), (40, -50)`

I'm not sure if the documentation is off or the logic. Please add some more test cases to make sure I understand correctly. Judging from the center_origin test cases it seems that the logic is correct and it's the wording of the documentation that confuses me.

tests/unit/gaze/gaze_transform_test.py
```diff
@@ -791,6 +791,82 @@ def fixture_experiment():
...
+                    screen_height_px=100,
+                    screen_width_cm=100,
+                    screen_height_cm=100,
+                    distance_cm=None,
+                    origin='center',
+                ),
+                'pixel_columns': ['x_pix', 'y_pix'],
+                'distance_column': 'distance',
+            },
+            'pix2deg', {'center_offset': (10, 10)},
```
dkrako, line 816: here it would be nice if `center_offset` would be included in the `Screen` class in `Experiment`, just like it is with `origin`. This way one would only need to define the center offset once in the Definition and further computations will be taken care of.

Would you be willing to include this in this PR? Depending on your motivation this can also be worked on in a follow-up PR, as this will lead to small changes further down the road.

dkrako (changes requested):
Great work, thank you so much!

I have some misunderstandings about some of the calculations. Probably it's just the wording in the docstring, but I would like to make sure. Please add some tests to make sure.

PR #940, opened 2025-02-10
hotfix: make timestamp in msg_prefix for metadata regex compile optional

### LINKED ISSUE #939: Problem with additional metadata when parsing eyelink data
## Question

I've used EyeLink together with PsychoPy in a behavioural task. Therefore, in my asc file there is metadata with information about trial start/end. The lines look like that:

```
MSG	2034441 SyncPulseReceived
```
or
```
MSG	2055307 TRIALID TRIAL 0 MOVIESTART
```

I've tried defining metadata_patterns when using gaze.from_asc. However, it didn't work. Could you please provide me with the direct syntax?

## Sample Code

Note, this is not working:

```
data = pm.gaze.from_asc(
    file='/path/to/file/P001.asc',
    metadata_patterns=[
        {'key': 'sync_pulse',
         'pattern': r"(?P<sync_pulse>\d+)\s+SyncPulseReceived"}
    ]
)
```

SiQube:
as raised in #939 some people might want to extract the timestamp of a custom metadata type. I currently don't know if this would be possible. this PR provides a patch which fixes the problem raised in #939 while maintaining all other properties of `compile_patterns`.

I've additionally added a test which reproduces the problem. the code change of interest is in `src/pymovements/utils/parsing.py`.

feel free to reject this proposed change with a strategy to pass the new test (`test_extract_timestamp_for_metadata`) in `tests/unit/gaze/io/asc_test.py` and comment the solution on #939. additionally, we should double check documentation s.t. people can easily extract custom (meta)data from ascii files.

resolves #939

### DIFF
--- a/src/pymovements/utils/parsing.py
+++ b/src/pymovements/utils/parsing.py
@@ -128,7 +128,7 @@ def compile_patterns(patterns: list[dict[str, Any] | str]) -> list[dict[str, Any
     list[dict[str, Any]]
         Returns from string compiled regex patterns.
     """
-    msg_prefix = r'MSG\s+\d+[.]?\d*\s+'
+    msg_prefix = r'MSG\s+(?:\d+?[.]?\d*\s+)?'
 
     compiled_patterns = []
 
--- a/tests/unit/gaze/io/asc_test.py
+++ b/tests/unit/gaze/io/asc_test.py
@@ -368,3 +368,60 @@ def test_from_asc_detects_mismatches_in_experiment_metadata(experiment_kwargs, i
     expected_msg = 'Experiment metadata does not match the metadata in the ASC file:\n'
     expected_msg += '\n'.join(f'- {issue}' for issue in issues)
     assert msg == expected_msg
+
+
+# 939
+# only testing lines MOVIESTART and SyncPulseReceived
+FILE_ASCII = r"""\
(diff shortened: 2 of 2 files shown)


dkrako:
@SiQube Thanks for taking care of #939 !

The `metadata_patterns` argument is intended to be used with patterns that are only expected to be parsed once per file.

I guess that this part of our code renders your workaround unusable for use cases where the messages have multiple occurrences witch changing values:

https://github.com/aeye-lab/pymovements/blob/8d3230cbb9f0dc6d969bc7b44b46561dc6b46833/src/pymovements/utils/parsing.py#L340-L351

If we take something like the message `MSG	2055307 TRIALID TRIAL 0 MOVIESTART`, your test pattern `(?P<movie_start>\d+).*?(?P<movie_id>\d+)\s+MOVIESTART` would match multiple times and reassign the values to the metadata field, losing the prior values in the process.

So I guess this is change is not harmful and can be still useful as a workaround to get the timestamp of a metadata pattern. 

In case of multiple matches per file this would only contain the last matched values. I would rather propose an additional `message_patterns` argument to parse messages, or extend the `patterns` argument to unify `message` patterns, `metadata` patterns, and sample column patterns as written in https://github.com/aeye-lab/pymovements/issues/939#issuecomment-2652231398

SiQube:
I closed this, let's work on #939 instead

Issue #893, opened 2024-11-09
Store calibration and other metadata in well-documented form


saeub:
## Description of the problem

After merging #884, some metadata are still left in the metadata dictionary (stored as a private attribute `GazeDataFrame._metadata`). In particular:

- Calibrations
- Validations
- Blinks
- Changes in tracked eye or sampling rate (see #887)

There should be a well-documented public interface to access these metadata.

## Description of a solution

I see several options to approach a solution:

- Storing those metadata in custom data structures as attributes of `GazeDataFrame` (similar to the metadata stored in `Experiment`, `EyeTracker` etc.)
- Storing time-dependent meta-events (like calibrations, validations, blinks) in a DataFrame similar to fixations and saccades (but it's probably difficult to find a useful set of columns for all the different types of events)

We might need to think about what metadata is available for non-EyeLink devices too, and how they could all fit into the same data structure.

## Minimum acceptance criteria

- [ ] Public interface to access (ideally) all metadata returned from `parse_eyelink()` (and possibly other file formats)
- [ ] Documentation for every type of metadata

dkrako:
Probably the easiest next step will be to pass the parsed blinks as an `EventDataFrame` to the `GazeDataFrame.__init__()`.
The dataframe will need the three columns: `name`, `onset`, `offset`. The `name` should be `blink`.

dkrako:
#1082 would get rid of the `data_loss_ratio`  and `data_loss_ratio_blinks` fields in the metadata dict.

dkrako:
the blink ratio is implemented in #1452.

As a last remaining task we should check if there are any fields left in `Gaze._metadata` that we didn't take care of.
If all of the fields are redundant or handled by specific preprocessing methods we can finally remove setting the private field `Gaze._metadata` in `from_asc()`.

cbueth:
Currently, `gaze._metadata` includes:
- `DISPLAY_COORDS`, `resolution`, `sampling_rate`, `tracked_eye`, `recorded_eye`
- `version_1`, `version_2`, `version_number`, `model`, `recorded_by`
- `weekday`, `month`, `day`, `year`, `time`, `datetime`
- `pupil_data_type`, `mount_configuration`
- `calibrations`, `validations`, `recording_config`, `total_recording_duration_ms`
- Any user-defined metadata patterns (`metadata_1`, `metadata_2`, etc.)

[...]

PR #942, opened 2025-02-11
Added Parsing Of BlinksFrom ASC To Data Frame

AnnaBhlr:
## Description

This is partly a solution to issue [#893 ](https://github.com/aeye-lab/pymovements/issues/893#issue-2645929832)

At the moment blinks are stored as a private attribute `GazeDataFrame._metadata`.

Fixes #

## Implemented changes

Blinks parsed from asc files are now parse into passed as an `EventDataFrame` to `GazeDataFrame.__init__()` as suggested by @dkrako 

The result looks like this:

```
data = pm.gaze.from_asc(file = file_path)
data.blinks
shape: (225, 4)
┌─────────────────┬────────────────┬─────────────┬─────────────┐
│ start_timestamp ┆ stop_timestamp ┆ duration_ms ┆ num_samples │
│ ---             ┆ ---            ┆ ---         ┆ ---         │
│ f64             ┆ f64            ┆ f64         ┆ i64         │
╞═════════════════╪════════════════╪═════════════╪═════════════╡
│ 2.034965e6      ┆ 2.035055e6     ┆ 91.0        ┆ 0           │
│ 2.052089e6      ┆ 2.052198e6     ┆ 110.0       ┆ 0           │
│ 2.052822e6      ┆ 2.052822e6     ┆ 1.0         ┆ 0           │
│ 2.053735e6      ┆ 2.053971e6     ┆ 237.0       ┆ 0           │
│ 2.054211e6      ┆ 2.054216e6     ┆ 6.0         ┆ 0           │
│ …               ┆ …              ┆ …           ┆ …           │
│ 3.255086e6      ┆ 3.25516e6      ┆ 75.0        ┆ 0           │
│ 3.276228e6      ┆ 3.276445e6     ┆ 218.0       ┆ 0           │
│ 3.276604e6      ┆ 3.276659e6     ┆ 56.0        ┆ 0           │
│ 3.277155e6      ┆ 3.27725e6      ┆ 96.0        ┆ 0           │
│ 3.278153e6      ┆ 3.278173e6     ┆ 21.0        ┆ 0           │
└─────────────────┴────────────────┴─────────────┴─────────────┘
```

- [x] blinks are stored as an event dataframe instead of metadata
- [x] GazeDataFrame contains both gaze data (`data.frame`) and blink events (`eventDataFram`e).
- [x] Proper initialization of the blinks DataFrame
## Type of change

## How Has This Been Tested?

- [x] Run locally with test data set

## Checklist:

### DIFF
--- a/src/pymovements/gaze/gaze_dataframe.py
+++ b/src/pymovements/gaze/gaze_dataframe.py
@@ -50,6 +50,8 @@ class GazeDataFrame:
         The experiment definition. (default: None)
     events: pm.EventDataFrame | None
         A dataframe of events in the gaze signal. (default: None)
+    blinks: pm.EventDataFrame | None
+        A dataframe of detected blink events. If None, an empty EventDataFrame is created. (default: None)
     auto_column_detect: bool
         Flag indicating if the column names should be inferred automatically. (default: False)
     trial_columns: str | list[str] | None
@@ -182,6 +184,7 @@ class GazeDataFrame:
             data: pl.DataFrame | None = None,
             experiment: Experiment | None = None,
             events: pm.EventDataFrame | None = None,
+            blinks: pm.EventDataFrame | None = None,
             *,
             auto_column_detect: bool = False,
             trial_columns: str | list[str] | None = None,
@@ -290,6 +293,19 @@ class GazeDataFrame:
 
         # Remove this attribute once #893 is fixed
         self._metadata: dict[str, Any] | None = None
+        
+        if blinks is None:
+            if self.trial_columns is None:
+                self.blinks = pm.EventDataFrame()
+            else:  
+                self.blinks = pm.EventDataFrame(
+                    data=pl.DataFrame(
+                        schema={column: self.frame.schema[column] for column in self.trial_columns},
+                    ),
+                    trial_columns=self.trial_columns,
+                )
+        else:
+            self.blinks = blinks.clone()
 
     def apply(
             self,
--- a/src/pymovements/gaze/gaze_dataframe.py
+++ b/src/pymovements/gaze/gaze_dataframe.py
@@ -293,11 +293,11 @@ class GazeDataFrame:
 
         # Remove this attribute once #893 is fixed
         self._metadata: dict[str, Any] | None = None
-        
+
         if blinks is None:
             if self.trial_columns is None:
                 self.blinks = pm.EventDataFrame()
-            else:  
+            else:
                 self.blinks = pm.EventDataFrame(
                     data=pl.DataFrame(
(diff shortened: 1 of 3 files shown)


dkrako (changes requested):
Thank you so much for taking care!

Instead of having a separate `EventDataFrame` for blinks, I would be in favor of having a single dataframe that holds all kind of eye movements. Each `EventDataFrame` contains column `name`, that specifies the name of the event. This could be `saccade`, `fixation` or in your case it would be `blink`.

Still your idea for quick access to blinks via `GazeDataFrame.blinks` is great. I would propose to implement this as a python property though.

So let's assume all events are stored in `GazeDataFrame.events`.

You could implement a `blink` property like this:

```python
@property
def blinks(self) -> pl.DataFrame:
    return self.events.filter(pl.col("name").str.starts_with('blink'))
```

This would return a dataframe with all events where the name starts with `blink`. (So technically you could have names `blink_eyelink` or `blink_custom` included)

Moreover, such a property would also be really nice to have available for `saccade` and `fixation`.

AnnaBhlr:
No worries. I can see your point re having all types of movements in `GazeDataFrame.events`. However, I was wondering if this is the best option for further data processing steps. Just as an example, microsaccade algorithm identifies saccades that are technical a blink artefact. Typically, the user would then exclude all identified saccades within 50 ms around the blink event. Having that as a function would also be nice.

saeub:
Hi @AnnaBhlr, thanks a lot for working on this! I've been working on an alternative implementation and just opened a draft PR (#945). It would be great if you could take a look and see if that implementation would be sufficient for your use cases.

dkrako:
> No worries. I can see your point re having all types of movements in `GazeDataFrame.events`. However, I was wondering if this is the best option for further data processing steps. Just as an example, microsaccade algorithm identifies saccades that are technical a blink artefact. Typically, the user would then exclude all identified saccades within 50 ms around the blink event. Having that as a function would also be nice.

Great idea! I just think it would be perfectly feasible to achieve this with the property-approach I proposed. I created issue #969 where we can further discuss if we want to have combined or separate event dataframes for the different kinds of events.

For this PR though I would like to stay consistent with the existing codebase and use a single `events` dataframe for all kinds of events.

I created a few issues to address some specific points in the discussion. Some of them are very brief, so feel free to fill in some more details

- #968 is the proposal for having read-only properties that allow quick-access to specific event types like fixations
- #969 is for discussing combined vs separate event data frames 
- #970 is for removing events that occur during blink events
- #971 is for setting samples to null that occur during blink events

dkrako:
Closing this in favor of #945. I hope that PR addresses your request. Anyway, thanks a lot for your initiative!

PR #946, opened 2025-02-12
reorder links in README.md

dkrako:


README.md
```diff
@@ -26,11 +26,11 @@ and render plots to visually analyze your results.
...
 - **Documentation:** https://pymovements.readthedocs.io
 - **Source code:** https://github.com/aeye-lab/pymovements
-- **Mailing list:** pymovements-list@uni-potsdam.de
-- **Contributing:** https://github.com/aeye-lab/pymovements/blob/main/CONTRIBUTING.md
-- **Bug reports:** https://github.com/aeye-lab/pymovements/issues
 - **PyPI package:** https://pypi.org/project/pymovements
 - **Conda package:** https://anaconda.org/conda-forge/pymovements
+- **Bug reports:** https://github.com/aeye-lab/pymovements/issues
+- **Contributing:** https://github.com/aeye-lab/pymovements/blob/main/CONTRIBUTING.md
+- **Mailing list:** pymovements-list@uni-potsdam.de
```
SiQube, line 33: I think we should move pymovements mailing list below discord because we will lose access to it once we graduated from UP (only retrieve accounts for 6 months)

SiQube (commented):
thank you for the update. I think this is a good change, just a minor discussion point left.

README.md
dkrako: Actually I intentionally put discord at the end of the list.

My line of thinking was that the last entry in the list will be much more noticeable than the ones before it (apart from the first ones).

I guess we won't find out before conducting an eye tracking study for this, but this may be overkill for this PR :smile: 

I created #972 for migrating to a new mailing list.

PR #914, opened 2025-01-01
feat: support yaml dataset definitions

SiQube:
## :boom: Breaking Changes
- `DatasetLibrary.get()` now returns an instantiated `DatasetDefinition` object instead of a `DatasetDefinition` class.

## Description

Supports specifying a  `DatasetDefinition` as a yaml file.

Example file:
```yaml
name: YourDataset

has_files:
  gaze: true
  precomputed_events: false
  precomputed_reading_measures: false

mirrors:
  gaze:
    - https://osf.io/download/

resources:
  gaze:
    - resource: your_resource/
      filename: target_filename.zip
```

PR #961, opened 2025-02-20
dataset: add RaCCooNS by Frank and Aumeistere

### LINKED ISSUE #954: add the RaCCooNS dataset
The Radboud Coregistration Corpus of Narrative Sentences (RaCCooNS)

https://data.ru.nl/collections/ru/cls/eeg_et_sentence_reading_dsc_556

Frank, S.L., Aumeistere, A. An eye-tracking-with-EEG coregistration corpus of narrative sentences. Lang Resources & Evaluation 58, 641–657 (2024). https://doi.org/10.1007/s10579-023-09684-x

SiQube:
resolves #954

### DIFF
--- a/docs/source/bibliography.bib
+++ b/docs/source/bibliography.bib
@@ -317,6 +317,17 @@ Milham, Michael P.},
   publisher={Nature Publishing Group}
 }
 
+@dataset{RaCCooNS,
+  author = {Frank, S.L. and Aumeistere, A.},
+  title = {{EEG+ET sentence reading}},
+  year = 2022,
+  publisher = {Radboud University},
+  version = 1,
+  doi = {10.34973/3g69-ha39},
+  url = {https://doi.org/10.34973/3g69-ha39}
+}
+
+
 @inproceedings{GazeGraph,
   title={{GazeGraph}: Graph-based few-shot cognitive context sensing from human visual behavior},
   author={Lan, Guohao and Heit, Bailey and Scargill, Tim and Gorlatova, Maria},
--- a/src/pymovements/datasets/raccoons.py
+++ b/src/pymovements/datasets/raccoons.py
@@ -1,0 +1,178 @@
+# Copyright (c) 2022-2025 The pymovements Project Authors
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+"""Provides a definition for the RaCCooNS dataset."""
+from __future__ import annotations
+
+from dataclasses import dataclass
+from dataclasses import field
+from typing import Any
+
+from pymovements.dataset.dataset_definition import DatasetDefinition
+from pymovements.dataset.dataset_library import register_dataset
+
+
+@dataclass
+@register_dataset
+class RaCCooNS(DatasetDefinition):
+    """RaCCooNS dataset :cite:p:`RaCCooNS`.
+
+    The Radboud Coregistration Corpus of Narrative Sentences (RaCCooNS) dataset consists
+    simultaneously recorded eye-tracking and EEG data from Dutch sentence reading,
+    aimed at studying human sentence comprehension and evaluating computational
+    language models. The dataset includes 37 participants reading 200 narrative sentences,
+    with eye movements and brain activity recorded to analyze reading behavior
+    and neural responses. The dataset provides both raw and preprocessed data,
+    including fixation-related potentials, enabling comparisons between cognitive
+    and neural processes.
+
+    Check the respective paper :cite:p:`RaCCooNS` for details.
+
+    Attributes
+    ----------
+    name: str
+        The name of the dataset.
+
+    has_files: dict[str, bool]
+        Indicate whether the dataset contains 'gaze', 'precomputed_events', and
+        'precomputed_reading_measures'.
+
+    mirrors: dict[str, tuple[str, ...]]
+        A tuple of mirrors of the dataset. Each entry must be of type `str` and end with a '/'.
+
+    resources: dict[str, tuple[dict[str, str | None], ...]]
+        A tuple of dataset gaze_resources. Each list entry must be a dictionary with the following
+        keys:
+        - `resource`: The url suffix of the resource. This will be concatenated with the mirror.
+        - `filename`: The filename under which the file is saved as.
+        - `md5`: The MD5 checksum of the respective file.
+
+    extract: dict[str, bool]
+        Decide whether to extract the data.
+
+    filename_format: dict[str, str]
+        Regular expression which will be matched before trying to load the file. Namedgroups will
+        appear in the `fileinfo` dataframe.
+
+    filename_format_schema_overrides: dict[str, dict[str, type]]
+        If named groups are present in the `filename_format`, this makes it possible to cast
+        specific named groups to a particular datatype.
+
+    trial_columns: list[str]
+            The name of the trial columns in the input data frame. If the list is empty or None,
+            the input data frame is assumed to contain only one trial. If the list is not empty,
+            the input data frame is assumed to contain multiple trials and the transformation
+            methods will be applied to each trial separately.
+
+    column_map: dict[str, str]
+        The keys are the columns to read, the values are the names to which they should be renamed.
+
+    custom_read_kwargs: dict[str, Any]
+        If specified, these keyword arguments will be passed to the file reading function.
+
+    Examples
+    --------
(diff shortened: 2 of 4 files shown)


SiQube:
there is additional raw gaze data. unfortunately, the files are ascii and we currently only support [`csv`, `tsv` and `feather` data](https://github.com/aeye-lab/pymovements/blob/9073c2814152e86ffa9622476d967c54f5352381/src/pymovements/dataset/dataset.py#L190-L192), see #962

SiQube:
yes, but we have to have the right parsing criterions. I'd want to do this in a seperate PR and reopen an issue. maybe someone from the original authors can help?

dkrako:
Ah I see, it's not the EyeLink format then? can you provide some example lines?

SiQube:
the files are actually eyelink ascii files. however I am not entirely sure how trials are split and I don't want to propagate mistakes. maybe we can setup a meeting with Stefan Frank where we can discuss how the trials were split? CC @izaskr

izaskr:
Alright, let me check. Will get back to you with this.

SiQube:
from @izaskr:

```quote
There's a variable TRIAL_INDEX that is increased by 1 at the end of each trial.
If you look for this in the text files you find the separations between consecutive trial.
I hope that answers David's question!
```

I'll implement it and then we can merge the dataset to pymovements.

saeub:
@SiQube I'll take care of the ASCII parsing (hopefully today)

saeub:
Two problems:

1. Participant IDs in TSV files are numbers, but participant IDs in ASC filenames are not in a number format.
2. Trial IDs start at 0 in ASC files but at 1 in TSV files, meaning the user won't be able to join the event and gaze frames.
3. Trial variable messages (`!V TRIAL_VAR`) are at the end of the trial, meaning we can't add info like stimulus ID to the gaze dataframe.

Regarding 2., I guess we can't really fix that right now? @SiQube (One option would be to allow passing functions with patterns in `from_asc()`, where a function takes the `match.groupdict()` and returns a dictionary of column values)

For 3., I created an issue #990.

SiQube:
maybe we can discuss with Frank and @izaskr ?

saeub:
@SiQube Should we wait with merging this until we have a solution for parsing trial variables (#990) and custom postprocessing functions (https://github.com/aeye-lab/pymovements/pull/961#issuecomment-2703823841)? Or do you want to merge this now and improve it later?

(I think the dataset is perfectly usable as it is now, just some of the information about the stimulus is missing, and the gaze and precomputed event frames don't match up.)

SiQube:
I don't mind merging it asap and fix it later => we can move one of the comments to a new issue. using pymovements for this dataset is still valuable (I think) since downloading and preprocessing works (for most of the data)

dkrako:
I'm in favor of merging this PR without the additional trial infos and improve on this later on when we have solved the underlying issues.

Also, adding a `custom_post_processing()` to a `DatasetDefinition` wouldn't help much when a user simply uses `gaze.from_asc()` to load a single file. #997 could help with this though.

dkrako:
We should probably also mention these issues in the docstring of the dataset.

src/pymovements/datasets/raccoons.py
```diff
@@ -0,0 +1,198 @@
...
+
+    column_map: dict[str, str]
+        The keys are the columns to read, the values are the names to which they should be renamed.
+
+    custom_read_kwargs: dict[str, Any]
+        If specified, these keyword arguments will be passed to the file reading function.
+
+    Examples
+    --------
+    Initialize your :py:class:`~pymovements.PublicDataset` object with the
```
dkrako, line 91: `PublicDataset` does not exist anymore. Use 

```
:py:class:`~pymovements.dataset.Dataset`
```

src/pymovements/datasets/raccoons.py
```diff
@@ -0,0 +1,198 @@
...
+    :py:class:`~pymovements.RaCCooNS` definition:
```
dkrako, line 92: use 

```
:py:class:`~pymovements.datasets.RaCCooNS`
```

dkrako (approved):
Alright, lets merge this.

I just have two comments to make the classes clickable in the documentation  (following #986).

@SiQube you can resolve the comments without changes in code if you like, as these probably aren't relevant anymore after merging #914

docs/source/bibliography.bib
dkrako, line 1: Ah I forgot one thing: can you add a line for the dataset to `docs/source/datasets/public_datasets.csv`.

src/pymovements/datasets/raccoons.py
SiQube: done

src/pymovements/datasets/raccoons.py
SiQube: done

SiQube:
@izaskr can you check with stefan why the dataset is down?
@izaskr in the paper its written `1018` pixel screen resolution, I assume its a type and it should be `1080`

docs/source/bibliography.bib
SiQube: done

src/pymovements/datasets/raccoons.py
```diff
@@ -0,0 +1,225 @@
...
+
+    filename_format_schema_overrides: dict[str, dict[str, type]] = field(
+        default_factory=lambda: {
+            'gaze': {'participant_id': str},
+            'precomputed_events': {},
+            'precomputed_reading_measures': {},
+        },
+    )
+
+    trial_columns: list[str] = field(default_factory=lambda: [])
```
saeub, line 209: @SiQube I just noticed that I forgot to add `trial_index0` to the trial columns here. Although I'm not sure we should add it actually, because it only exists in the gaze dataframe, but not in the event dataframe?

saeub:
Also, the parser is getting conflicting information regarding the eye tracker version:

- The ASC files say it's `EYELINK II CL v4.594 Jul  6 2012`, which is translated to `EyeLink 1000` in our parsing logic: https://github.com/aeye-lab/pymovements/blob/913c372f83d1d19eada650f28fd2ec52d741f0ad/src/pymovements/gaze/_utils/parsing.py#L508-L509
- The experiment metadata (and the paper) says it should be `EyeLink 1000 Plus`.

Is anyone familiar with these version mappings? (maybe @theDebbister, since she wrote the metadata parsing code?)

src/pymovements/datasets/raccoons.py
saeub: Maybe `trial_columns="trial_index"` should only be added to the `custom_read_kwargs` for gaze?

SiQube:
@izaskr @theDebbister any update?

theDebbister:
Yes, Stefan said it is possible that the paper is wrong and they used the EyeLink 1000 instead of 1000 Plus. It seems that our device detection is still correct.

src/pymovements/datasets/raccoons.py
```diff
@@ -0,0 +1,225 @@
...
+            ],
+        },
+    )
+    resources: dict[str, list[dict[str, str | None]]] = field(
+        default_factory=lambda: {
+            'gaze': [
+                {
+                    'resource': 'download/eyetracking%2FET_raw_data.zip',
+                    'filename': 'ET_raw_data.zip',
+                    'md5': None,  # type: ignore
```
dkrako, line 142: why aren't any checksums provided?

src/pymovements/datasets/raccoons.py
```diff
@@ -0,0 +1,225 @@
...
+
+    Load the data into memory:
+
+    >>> dataset.load()# doctest: +SKIP
+    """
+
+    # pylint: disable=similarities
+    # The DatasetDefinition child classes potentially share code chunks for definitions.
+
+    name: str = 'RaCCooNS'
```
dkrako, line 114: long_name is missing

tests/files/raccoons.asc
dkrako, line 1: did you obtain the rights to publish this file?

dkrako: If not, then please remove as many data from the file as possible.

The samples are not that relevant and there can be removed a lot of lines.

For testing the correct patterns, please have at least 2 trials in this file (can be few ms each).

dkrako (changes requested):

src/pymovements/datasets/raccoons.yaml
```diff
@@ -0,0 +1,81 @@
...
+
+mirrors:
+  gaze:
+    - "https://data.ru.nl/api/collectionfiles/ru/id/ru_395469/files/"
+  precomputed_events:
+    - "https://data.ru.nl/api/collectionfiles/ru/id/ru_395469/files/"
+  precomputed_reading_measures:
+    - "https://data.ru.nl/api/collectionfiles/ru/id/ru_395469/files/"
+
+resources:
```
dkrako, line 5: this way of specifying resources was deprecated.

Instead you should now write:

```yaml
resources:
  - content: gaze
    resource: 'download/eyetracking%2FET_raw_data.zip'
    filename: "ET_raw_data.zip"
    md5: null
  - content: precomputed_events
    resource: 'download/eyetracking%2FET_fix_data.tsv'
    filename: "ET_fix_data.tsv"
    md5: null
  - content: precomputed_reading_measures
    resource: 'download/eyetracking%2FET_word_data.tsv'
    filename: "ET_word_data.tsv"
    md5: null
```

dkrako:
@cbueth not yet ready for review

dkrako:
@SiQube Can you identify which file(s) raise the error? I am currently downloading the dataset but it's very slow.

To me it seems that one of the files switch the tracked eye. We currently do not support that, see issue #1401. CoLaGaze has a similar problem. I guess we should raise the priority of that issue.

However, as that's out of scope of this PR, I would be in favor to just put a warning like with colagaze: https://pymovements.readthedocs.io/en/latest/reference/api/pymovements.datasets.CoLAGaze.html

src/pymovements/datasets/raccoons.yaml
```diff
@@ -0,0 +1,53 @@
...
+    md5: c40886c4515c43187aba8fbc32c8c935
+    filename_pattern: "ET_word_data.tsv"
+    load_kwargs:
+      separator: "\t"
+      encoding: "latin-1"
+
+experiment:
+  screen:
+    width_px: 1920
+    # double check with stefan -- 1018 in the paper
```
dkrako: did you check this already? if not, please create an issue. in either case, please remove that comment.

src/pymovements/datasets/raccoons.py
```diff
@@ -0,0 +1,153 @@
...
+                },
+            ],
+        ),
+    )
+
+    experiment: Experiment = field(
+        default_factory=lambda: Experiment(
+            screen=Screen(
+                width_px=1920,
+                # in the paper its written 1018, but assume typo
```
dkrako: same

tests/files/raccoons.asc
dkrako: I don't remember if we discussed this already, but I'll like to make sure before merge.

PR #965, opened 2025-02-24
chore: remove unnecessary .frame from import

SiQube:


dkrako (approved):
Great, thanks for catching this!

PR #966, opened 2025-02-24
docs: improve datasets error message

SiQube:
`gaze_resources` is an artifcat from a previous implementation.


dkrako (approved):
Great! Thanks for improving the docs!

PR #976, opened 2025-02-26
depr: Deprecate `utils.parsing` in favor of `gaze.from_asc()`

### LINKED ISSUE #973: deprecate `utils.parsing` and move functionality to private `gaze._utils.parsing`
## Description of the problem

Due to the current developments on our parsing module we get a lot of breaking changes due to enhancements in functionality.
For example #945 is one of those.

## Description of a solution

The `parse_eyelink()` method and the whole parent module `utils.parsing` should be private.

The user should use `gaze.from_asc()` instead.

To be backwards compatible for the next few we releases we should deprecate the function first and then remove it in one of the future versions. All the updates to the private parsing function interface won't lead to breaking changes if we do it like this:

```python

def parse_eyelink(...):
    #raise deprecation warning. include version on which function will be removed.

    # only use gaze_data and metadata for backwards compatible return values
    gaze_data, _, metadata = pm.gaze._utils.parsing.parse_asc(...)

   return gaze_data, metadata
```

dkrako:
## :warning: Deprecation

`utils.parsing` is deprecated. please use `gaze.from_asc()` instead.
`utils.parsing` will be removed in pymovements v0.26.0

## Description

Deprecate `utils.parsing` in favor of `gaze.from_asc()`

The following warning is printed:

```
DeprecationWarning: Call to deprecated function (or staticmethod) parse_eyelink. (Please use gaze.from_asc() instead of utils.parsing. This module will be removed in v0.26.0.) -- Deprecated since version v0.21.0.
```

Fixes #973 

Should be merged before #945

## Implemented changes

- [x] added `deprecated` as a dependency for `@deprecated` decorator
- [x] moved `utils/parsing.py` to `gaze/_utils/parsing.py` and adjusted to new location
- [x] use `gaze._utils.parsing` in `gaze.io` 
- [x] created depracted function in `utils/parsing.py`
- [x] moved `tests/unit/utils/parsing_test.py` to `tests/unit/gaze/_utils/parsing_test.py`
- [x] add test that function is correctly deprecated
- [x] add test that function is removed in planned version


dkrako:
finally

dkrako:
let's merge this after releasing v0.20.0

dkrako:
this is now ready for being merged

dkrako:
Why was #989 merged before this?

SiQube:
whoops -- #954 was the pain point 🙈

SiQube:
very true -- I just forgot about this PR. I think the "easiest" way would be to revert #989, merge #976, and then adjust #989 in a new PR.... sorry

dkrako:
I try to resolve the conflicts in this PR

SiQube:
see #993

SiQube:
okay -- if you decide to revert and remerge, lmk and I'll fix #989

dkrako:
shit happens, we're all doing errors here and now. luckily there weren't many changes in #989.

please review the merge commit against the changes in #989

tests/unit/gaze/_utils/_parsing_test.py
```diff
@@ -0,0 +1,920 @@
+# Copyright (c) 2023-2025 The pymovements Project Authors
```
SiQube, line 1: I think the filename should be `tests/unit/gaze/_utils/parsing_test.py` or does this conflict with `tests/units/utils/parsing_test.py`?

SiQube (commented):
just a couple of feedback questions:

1. comment below
2. why `gaze/_utils` instead of `gaze/utils`?

dkrako:
> just a couple of feedback questions:
> 
>     1. comment below
> 
>     2. why `gaze/_utils` instead of `gaze/utils`?

I wanted to make sure that users understand that this is a private utils module that is not intended for the end-user.

tests/unit/gaze/_utils/_parsing_test.py
dkrako: We can also use `gaze/_utils/_parsing.py` or `gaze/_parsing.py` for the source module if you like
As it is a private module, we can change the name anytime.

I wouldn't be much in favor of `gaze/utils/_parsing.py`.

SiQube (approved):
okay reasonable, I have to preference and let make pytest happy :-)

PR #977, opened 2025-02-26
docs: change mail address to @python.org

SiQube:


dkrako (approved):
Wow so fast!

SiQube:
@saeub I think it would be good to add it to the `README.md`, can you send a PR?

PR #985, opened 2025-03-04
docs: add Datasets tab to main page in documentation

dkrako:


SiQube:
given #914 how would this look?

we have then a `datasets.yaml` which can be automatically created. maybe we can just this list somehow?

dkrako:
yes I thought about it today. Just copying the old docstring won't work. I just wanted to test adding the tab for a first draft.

I did a bit of research and maybe I'd just create a table in the rst file.

There's the possibility of reading in csv files into a table: https://sublime-and-sphinx-guide.readthedocs.io/en/latest/tables.html#csv-files

Maybe that's a feasible approach for this PR? We could create a script which downloads all datasets and creates a csv with some summary data of the datasets. This script should be run on every dataset update, probably run manually for now.

The csv would then be parsed by sphinx.

SiQube:
currently I create a yaml file but we can probably also just put the dataset into a csv. I push everything before tomorrows meeting so you can have a look. I'm missing a couple of tests and correct `to_yaml` due to types but rest is ready to go #914

dkrako:
Alright, here's the first draft for the dataset tab page:

https://pymovements.readthedocs.io/en/docs-datasets-tab/datasets/index.html

I'd like to delegate the adding of all the dataset info columns to someone else if possible.
I'll leave a message on discord to request help soon

dkrako:
As I just released a new pymovements version I would like to merge this PR with just the minimal table.

Adding all further information (sampling rate, types of stimuli, number of users, number of sessions per user) and correcting the copy-paste X's (from the samples, events, reading measures columns) could then be done in a follow-up PR

SiQube (approved):

PR #1003, opened 2025-03-06
test: add global gaze fixtures

dkrako:
WIP 

#### required by:
- #1002

#### requires:
- #914
- #1004


SiQube:
great idea, I also thought of that a while ago, thank you for tackling the issue.

initial feedback/thought: since these are not directly tests, maybe have another repository like `testing` where we put all fixtures etc.

dkrako:
I'm not sure that I understand. fixtures are a very common part of test suites. fixtures should be tested though.

The way I implemented this and imported in `conftest.py` is actually the intended way by pytest, so I'm rather confused about what you mean that fixtures are not directly tests.

What I really wouldn't like is creating another repository just for testing.

Can you provide a link to repository where your idea is implemented?

SiQube:
[here](https://github.com/pre-commit/pre-commit/blob/main/testing/fixtures.py) with combination [this](https://github.com/pre-commit/pre-commit/blob/main/tests/conftest.py)

dkrako:
> [here](https://github.com/pre-commit/pre-commit/blob/main/testing/fixtures.py?rgh-link-date=2025-04-22T12%3A02%3A28Z) with combination [this](https://github.com/pre-commit/pre-commit/blob/main/tests/conftest.py?rgh-link-date=2025-04-22T12%3A02%3A28Z)

Ah thank for the comment, now I understand what you mean. This isn't a separate repository, it's a dedicated module. Please make sure you use the correct words, they have a meaning.

But as a side note: the `fixtures.py` you linked is really bad, as it doesn't provide any fixtures. It's really unclear what that module really achieves as the fixtures are actually defined in `conftest.py`.

I also thought about that originally as this is done in some packages, but currently I don't see much benefit to add this to the user API. If done bad, we would need to add `pytest` as a dependency (probably that's why the linked `fixtures.py` is so bad to avoid this). Additionally, this would lead to inflexibilities in development, as then we would need to take care of breaking changes.

I guess it makes sense in cases, where users extend the original package, so providing users with fixtures can make sense. For example in polars you can [extend the api](https://docs.pola.rs/api/python/stable/reference/api.html), so having fixtures available in a testing module is quite nice.

It will take some time until this will be finished, so we'll have some time to discuss this further.

Issue #559, opened 2023-09-15
fix circular import in gaze_dataframe.py


dkrako:
This PR includes an EventsDataFrame in a GazeDataFrame, which leads to circular dependencies when importing both.

Doing something like
```python
from pymovements.gaze.gaze_dataframe import GazeDataFrame
```
results in a hard python circular import error.

I have found a workaround by importing the pymovements top-level package, but then pylint is not happy with me again.

Right now I just can't find a way to untie this knot. As this is just a linting issue, I would like to create a follow-up issue on how to solve this problem and leave it like this for this PR.

_Originally posted by @dkrako in https://github.com/aeye-lab/pymovements/pull/553#discussion_r1325803013_

romajaku:
Hi all, when trying to import pymovements within a jupyter notebook created in the tutorials directory (both old and newly created) i get the following error.

```
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
Cell In[9], [line 2](vscode-notebook-cell:?execution_count=9&line=2)
      [1](vscode-notebook-cell:?execution_count=9&line=1) data_path = Path("../../../tests/files/eyelink_monocular_example.asc")
----> [2](vscode-notebook-cell:?execution_count=9&line=2) data, metadata = pm.gaze.from_asc(
      [3](vscode-notebook-cell:?execution_count=9&line=3)     data_path,
      [4](vscode-notebook-cell:?execution_count=9&line=4)     patterns=[
      [5](vscode-notebook-cell:?execution_count=9&line=5)         # TODO: Update patterns (https://github.com/theDebbister/multipleye-preprocessing/issues/5#issuecomment-2230701372)
      [6](vscode-notebook-cell:?execution_count=9&line=6)         r"start_recording_(?P<trial>(?:PRACTICE_)?trial_\d+)_(?P<screen>.+)",
      [7](vscode-notebook-cell:?execution_count=9&line=7)         {"pattern": r"stop_recording_", "column": "trial", "value": None},
      [8](vscode-notebook-cell:?execution_count=9&line=8)         {"pattern": r"stop_recording_", "column": "screen", "value": None},
      [9](vscode-notebook-cell:?execution_count=9&line=9)         {
     [10](vscode-notebook-cell:?execution_count=9&line=10)             "pattern": r"start_recording_(?:PRACTICE_)?trial_\d+_page_\d+",
     [11](vscode-notebook-cell:?execution_count=9&line=11)             "column": "activity",
[...]

PR #1004, opened 2025-03-06
feat: support passing `DatasetDefinition` to gaze init and io

### LINKED ISSUE #997: pass a `DatasetDefinition` to `gaze.io` functions
## Description of the problem
Currently, a user has to manually forward all the relevant fields from a `DatasetDefinition` to the io functions `gaze.from_csv()`, `gaze.from_asc()`, etc. if they just want to load in a single gaze file.

## Description of a solution
It would be much nicer, if a user could simply pass a dataset definition instead of being burdened with forwarding all these fields manually.

dkrako:
## Description

Allows for passing a `DatasetDefinition` to `GazeDataFrame` instead of specifying all columns etc. explicitly.
This is very handy when dealing with single files of a dataset that already has a definition implemented.

The enhancement is also added to `gaze.from_csv()`, and `gaze.from_asc()`. `gaze.from_ipc()` was excluded, as this function is rudimentary and currently a very special case in our codebase with barely any explicit arguments, as we expect it to be a processed file adhering to our formatting standard.

All arguments take precedence, so passing explicitly any arguments to these functions (like `experiment` or `pixel_columns`) will override the corresponding fields of `definition`.

## Implemented changes

- [x] add `definition` argument to `GazeDataFrame.__init__()`
- [x] add `definition` argument to `gaze.from_csv()`
- [x] add `definition` argument to `gaze.from_asc()`
- [x] simplified existing usages in source where appropriate
- [x] simplified existing usages in tests where appropriate

## Type of change

## How Has This Been Tested?

- [x] Tested passing `definition` to `GazeDataFrame.__init__()`
- [x] Tested `definition` overwrite by explicit arguments in `GazeDataFrame.__init__()`
- [x] Tested passing `definition` to `gaze.from_csv()`
- [x] Tested `definition` overwrite by explicit arguments in `gaze.from_csv()`
- [x] Tested passing `definition` to `gaze.from_asc()`
- [x] Tested definition overwrite in `gaze.from_asc()`

## Context

Resolves #997

#### required by:
- #1003

#### requires:
- [x] #1006
- [x] #1068 
- [x] #1069 
- [x] #1070
- [x] #1076 
- [x] #1084 
- [x] #1085
- [x] #1092

## Checklist:


dkrako:
I'm left with a few questions on how to decide on argument precedence:

For the first iteration I implemented this in the way that `definition` takes precedence over all other arguments like `time_column`, `position_columns` etc., as this is the easiest way to implement.

But I am not sure of this is the right way. Wouldn't the opposite make more sense?

Let's take this for an example:

```python
gaze = pm.GazeDataframe(data, definition, pixel_columns=['xl_new', 'xr_new'])
```

The user would clearly expect that the explicitly specified `pixel_columns` have precedence over the ones in `definition`.

But there's a big problem with the `time_unit` argument. Its default is `ms`. If the explicit arguments take precedence over the definition, the default should be set to `None`. This should be easy to adapt to.

Also, what if someone wants to explicitly set `None` to one of the values? This use case is probably very rare and there's a simple workaround: just set the value to `None` in the dataset definition and your fine. So that's probably a non-issue.

We could also make `definition` and the other arguments mutually exclusive, but this would be very limiting I think.

Additionally, it would probably be great to allow the names of the datasets in the library:

```
definition: pm.DatasetDefinition | str | None
```

This would make the argument even more useful, but I'll probably leave this to a follow-up PR.

dkrako:
Unfortunately I can't use `from pymovements.dataset.DatasetDefinition` here due to a circular import:

```
ImportError while importing test module '/home/krakowczyk/workspace/pymovements/tests/functional/dataset_processing_test.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib/python3.12/importlib/__init__.py:90: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/functional/dataset_processing_test.py:23: in <module>
    import pymovements as pm
src/pymovements/__init__.py:22: in <module>
    from pymovements import datasets
src/pymovements/datasets/__init__.py:58: in <module>
    from pymovements.datasets.bsc import BSC
src/pymovements/datasets/bsc.py:27: in <module>
    from pymovements.dataset.dataset_definition import DatasetDefinition
src/pymovements/dataset/__init__.py:41: in <module>
    from pymovements.dataset.dataset import Dataset
src/pymovements/dataset/dataset.py:32: in <module>
    from pymovements.dataset import dataset_download
src/pymovements/dataset/dataset_download.py:26: in <module>
    from pymovements.dataset.dataset_definition import DatasetDefinition
src/pymovements/dataset/dataset_definition.py:27: in <module>
    from pymovements.gaze.experiment import Experiment
src/pymovements/gaze/__init__.py:82: in <module>
    from pymovements.gaze.gaze_dataframe import GazeDataFrame
src/pymovements/gaze/gaze_dataframe.py:34: in <module>
    from pymovements.dataset import DatasetDefinition
E   ImportError: cannot import name 'DatasetDefinition' from partially initialized module 'pymovements.dataset' (most likely due to a circular import) (/home/krakowczyk/workspace/pymovements/src/pymovements/dataset/__init__.py)
```

This is a greater issue that needs to be addressed separately and will probably need some restructuring of our codebase, see #559.

src/pymovements/gaze/io.py
```diff
@@ -25,6 +25,7 @@
 
 import polars as pl
 
+import pymovements as pm  # pylint: disable=cyclic-import
```
dkrako, line 28: this is needed as this PR creates a circular dependency between the `dataset` (which contains `DatasetDefinition` and for example imports `GazeDataFrame` and `from_csv()`) and `gaze` module (which imports `DatasetDefinition`).

in case of circular dependencies there are basically two ways to go:

1. split one of the modules
2. merge some of the modules

I'd say that all the definition classes like `DatasetDefinition`, `Experiment`, `Screen` or `EyeTracker` have very much in common. Also they are expected to be imported and not expected to import any higher order functionality.

So I would propose to create a `definition` module in the future which holds these (data)classes. I create an issue for this before merge.

Related to #559 but there it's a circular import between the `events` and `gaze` module

tests/functional/gaze_file_processing_test.py
```diff
@@ -22,7 +22,10 @@
 
 import pytest
 
-import pymovements as pm
+from pymovements import datasets
+from pymovements import Experiment
+from pymovements import EyeTracker
+from pymovements import gaze as gaze_module
```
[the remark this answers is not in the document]
SiQube: tbh we should just rename `gaze` in the loop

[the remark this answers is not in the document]
dkrako: you're probably right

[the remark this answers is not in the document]
dkrako: So I reconsidered a bit and created an issue related to this #1124
I would like to leave the variable `gaze` in the test function like it is. When implementing #1124 the current import line can be changed accordingly.

tests/functional/gaze_file_processing_test.py
```diff
@@ -53,117 +56,83 @@ def fixture_gaze_init_kwargs(request):
...
             ),
         },
         'didec': {
             'file': 'tests/files/didec_example.txt',
-            'time_column': pm.DatasetLibrary.get('DIDEC').time_column,
-            'time_unit': pm.DatasetLibrary.get('DIDEC').time_unit,
-            'pixel_columns': pm.DatasetLibrary.get('DIDEC').pixel_columns,
-            'experiment': pm.DatasetLibrary.get('DIDEC').experiment,
-            **pm.DatasetLibrary.get('DIDEC').custom_read_kwargs['gaze'],
+            'definition': datasets.DIDEC(),
```
SiQube, line 103: adding these lines makes it harder and harder to get rid of datasets.PUBLICDATASET

SiQube (approved):

tests/functional/gaze_file_processing_test.py
dkrako: don't worry, we will get rid of the class implementations without any breaking change by forwarding to the `DatasetLibrary` item for each dataset: `pymovements.datasets.DIDEC = pymovements.DatasetLibrary.get('DIDEC')`.

dkrako: I don't expect any user to use `DatasetLibrary` ever at all.

Issue #919, opened 2025-01-08
create method to check Experiment, Screen and EyeTracker for equality


dkrako:
## Description of the problem
There are a lot of places in our source and test code that manually checks for equality of `Experiment`, `Screen` or `EyeTracker` objects.

## Description of a solution
Create a method like `Experiment.equals(Experiment: other) -> bool` with corresponding methods for the `Screen` and `EyeTracker` classes.

## Minimum acceptance criteria
- [ ] `Experiment.equals(Experiment: other)`
- [ ] `Screen.equals(Screen: other)`
- [ ] `EyeTracker.equals(EyeTracker: other)`

dkrako:
resolved by #1008 and #1010 . `EyeTracker` was already implemented as a dataclass.

PR #1008, opened 2025-03-07
feat: implement `Screen` as dataclass

dkrako:
# :boom: (not really a) Breaking Change
This PR removes `utils.decorators`. As the documentation clearly states that the module is only intended for developers, I consider this not a breaking change. I recommend using `@dataclass` instead.

resolves partly: #919 

step towards equality operators for `Experiment` and `DatasetDefinition`


src/pymovements/gaze/gaze_dataframe.py
```diff
@@ -166,7 +166,7 @@ class GazeDataFrame:
     >>> gaze = GazeDataFrame(data=df_no_time, experiment=experiment, pixel_columns=['x', 'y'])
     >>> gaze
     Experiment(sampling_rate=100, screen=Screen(width_px=1024, height_px=768, width_cm=38,
-    height_cm=30, distance_cm=60, origin=center), eyetracker=None)
+    height_cm=30, distance_cm=60, origin='center'), eyetracker=None)
```
SiQube, line 169: probably didnt complain but we have floats with 1 decimal point for all other floats (at least in this patch)

src/pymovements/gaze/screen.py
```diff
@@ -20,20 +20,21 @@
...
-from pymovements.utils import decorators
 
 
-@decorators.auto_str
+@dataclass
 class Screen:
     """Screen class for holding screen properties.
 
-     Also transforms pixel coordinates to degrees of visual angle.
+    Also transforms pixel coordinates to degrees of visual angle.
```
SiQube, line 35: this comment might be confusing if people want to use pixel coordinates instead of dva.

SiQube (commented):
I think one minor docstring update could improve the PR

src/pymovements/gaze/screen.py
dkrako: I removed the line completely from the docstring to reduce some of the confusion.

The thing with the `pix2deg()` inside this class is a tough knot to unfold.

It relies on our legacy numpy functions that still aren't officially deprecated.
The `y_max_dva` etc. properties rely on this. The heatmap plot relies on the `y_max_dva` properties.

I think the functionality itself shouldn't be removed, but maybe it could fit somewhere else better? I don't know.

SiQube (approved):

Issue #978, opened 2025-02-27
Add more datasets?


AnnaBhlr:
Those are some open data sets from the cogneuro space that might be good candidates for adding:

Forrest Gump Dataset
   - Simultaneous fMRI and eye tracking during a long naturalistic viewing task
   - Data: https://openneuro.org/datasets/ds000113/versions/1.3.0
   - Reference: https://doi.org/10.1038/sdata.2016.92

EyeT4Empathy
   - Dataset for foraging visual information, gaze typing, and empathy assessment 
   - Reference: https://doi.org/10.1038/s41597-022-01862-w

Eye-Tracking Dataset for Autism Spectrum Disorder  
   - Data: https://figshare.com/articles/dataset/Eye-Tracking_Dataset_to_Support_the_Research_on_Autism_Spectrum_Disorder/20113592?file=35980367

Hollywood Video Clips Eye-Tracking Dataset
   - eye movements recorded while watching video clips.  
   - Reference: https://doi.org/10.1016/j.dib.2019.103991
   - Data: https://osf.io/g64tk/

Eye Movements During Viewing of Complex Images  
   - Reference: https://doi.org/10.1038/sdata.2016.126
   - Data: https://datadryad.org/dataset/doi:10.5061/dryad.9pf75

Gaze Tracking Dataset for Smooth vs. Saccadic Eye Movements
   - Reference: https://doi.org/10.1016/j.dib.2021.106730)
   - Data is in the supplementary materials

SiQube:
PRs:
- #1009

xiaotiansu:
One more dataset from CHI'23

- [ ] UEyes: a large eye-tracking-based dataset including 62 participants looking at 1,980 UI screenshots, covering four major UI types: webpage, desktop UI, mobile UI, and poster
	- Reference: https://dl.acm.org/doi/10.1145/3544548.3581096
	- Data: https://zenodo.org/records/8010312

PR #1009, opened 2025-03-07
dataset: add mousecursor dataset by Pantanowitz et al.

SiQube:
to download the dataset we need to add a `User-Agent` to the opener.

part of #978


SiQube:
I'll adjust if #914 is merged first

dkrako (changes requested):
This is a dataset which provides gaze samples but has no screen or sampling rate specified. Can you add this? If there's some more info on the used eyetrackers it would be great if you could fill the respective fields in the eyetracker definition.

SiQube:
@dkrako

docs/source/datasets/public_datasets.csv
```diff
@@ -14,6 +14,7 @@ Dataset,Citation,Samples,Events,Reading Measures,# Subjects,# Sessions per Subje
 :py:class:`~pymovements.datasets.HBN`,:cite:p:`HBN`,X,,,1246,1,120 Hz,Video
 :py:class:`~pymovements.datasets.InteRead`,:cite:p:`InteRead`,X,X,,50,1,1200 Hz,Text (documents)
 :py:class:`~pymovements.datasets.JuDo1000`,:cite:p:`JuDo1000`,X,,,150,4,1000 Hz,Random Dots
+:py:class:`~pymovements.datasets.JuDo1000`,:cite:p:`MouseCursor`,X,,,2,1,3 Hz, Mouse Cursor (saccadic vs smooth pursuit)
```
dkrako, line 17: There's a copypaste error in the first column.

In the last column I would just list it as `Mouse Cursor` and specify the details in the docstring, as the `saccadic vs smooth pursuit` is missing there.

dkrako (changes requested):
just minor change requests in the csv and maybe add the smooth pursuit info to the docstring instead.

docs/source/datasets/public_datasets.csv
SiQube: done

dkrako (approved):
Great, thanks for the addition!

Issue #891, opened 2024-11-09
Avoid type ambivalence in Experiment.screen and Experiment.eyetracker


saeub:
## Current Behavior

At least `Experiment.eyetracker` can be `None` (if `sampling_rate` is passed in `Experiment.__init__()`). `Experiment.screen` should always be of type `Screen`, unless it is tampered with after initialization.

Details: https://github.com/aeye-lab/pymovements/pull/884#discussion_r1833448210

## Expected Behavior

`Experiment.screen` should always be of type `Screen`, `Experiment.eyetracker` should alwys be of type `EyeTracker`.

## Minimum acceptance criteria

- Implement `Experiment.screen` and `Experiment.eyetracker` as non-null properties.
- [ ] Remove unnecessary `None` check in `from_asc()`
   -  see https://github.com/aeye-lab/pymovements/pull/884#discussion_r1833448210)
   - resolved in #1010 

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I have provided sufficient information for the team

lattaai5:
Hello, I tried to fix the issue

This is what I did:

Implemented read-only properties for Experiment.screen and Experiment.eyetracker to ensure they are always of the correct type. Removed unnecessary None check in the __init__ method.

I also created pull request: https://github.com/aeye-lab/pymovements/pull/892


> [!CAUTION]
> Disclaimer: This fix was created by Latta AI and you should never merge before you check the correctness of generated code!
>
> The fix provided by Latta AI might not be complete and it can serve as an inspiration.

---
This bug was fixed for free by Latta AI - https://latta.ai/ourmission

If you no longer want Latta AI to attempt fixing issues on your repository, you can block this account.

dkrako:
I labeled it nonessential as it's only protecting the user against themselves. feel free to raise the priority in case

dkrako:
I marked #1010 to resolve this issue, as the main issue lies in the type ambivalence of the eyetracker field.

I created #1011 as a followup for making attributes tamper-free.

Issue #402, opened 2023-05-11
Use import pymovements as pm in all test files


dkrako:
## Description of the problem
In some test files pymovements functionality is still imported using the complete path like `from pymovements.events.events import EventDataFrame`

## Description of a solution
As we recommend to use `import pymovements as pm` in all tutorials we should also check that we can actually import it that way.

If we use the recommended import in all test files, 
Some tests are testing stuff that is not part of the public API, that must be left as it is then.

## Minimum acceptance criteria
Specify what is necessary for the issue to be closed.

PR #1010, opened 2025-03-07
feat: improve `Experiment` class

dkrako:
# :boom: (not really a) Breaking Change
Restricts eyetracker argument to keyword-only. As `sampling_rate` and `eyetracker` already was mutually exclusive, I don't see how this would break any existing workflow.

## Description

- restricts eyetracker argument to keyword-only
- implements equality operator #919 
- resolves type ambivalence of eyetracker field #891 
- adds screen argument
- necessary changes to experiment field in `DatasetDefinition` as `Experiment` now allows for empty init
- remove irrelevant experiment field from some `DatasetDefinitions`

This PR now also allows for full equality comparisons between `DatasetDefinition` instances.

Blocked by #1008


src/pymovements/gaze/experiment.py
```diff
@@ -93,42 +95,48 @@ def __init__(
...
-
+        checks.check_is_mutual_exclusive(screen_width_px=screen_width_px, screen=screen)
+        checks.check_is_mutual_exclusive(screen_height_px=screen_height_px, screen=screen)
+        checks.check_is_mutual_exclusive(screen_width_cm=screen_width_cm, screen=screen)
+        checks.check_is_mutual_exclusive(screen_height_cm=screen_height_cm, screen=screen)
+        checks.check_is_mutual_exclusive(distance_cm=distance_cm, screen=screen)
         checks.check_is_mutual_exclusive(sampling_rate=sampling_rate, eyetracker=eyetracker)
 
+        # the origin default value needs special care.
+        if origin != 'upper left' or (screen is not None and screen.origin != origin):
```
dkrako, line 110: this needs to be improved, otherwise a user needs to make sure that the explicit `origin` is equal to `screen.origin`

SiQube (approved):

tests/unit/gaze/experiment_test.py
```diff
@@ -20,19 +20,59 @@
 """Test for Experiment class."""
 import pytest
 
-from pymovements.gaze.experiment import Experiment
-from pymovements.gaze.eyetracker import EyeTracker
+import pymovements as pm
```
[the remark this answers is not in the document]
dkrako: hm I see, fair point.

There's a related issue #402 which I created a long time ago.

My line of thinking was there, that we should test the actual paths that we want to expose to the end-user, not those long paths that nobody should use.

But maybe both viewpoints can be combined.

What do you think about this?

```python
from pymovements import Experiment
from pymovements import EyeTracker
```

src/pymovements/gaze/experiment.py
dkrako: I rewrote this to that:

https://github.com/aeye-lab/pymovements/blob/a96571b4345ec2c9507dc3d781560ae37f330680/src/pymovements/gaze/experiment.py#L109-L112

This should be fine now.

Issue #980, opened 2025-02-27
DIDEC dataset does not load


SiQube:
## Current Behavior
DIDEC dataset does **NOT** load.

## Expected Behavior
DIDEC dataset does load into memory.

## Minimum acceptance criteria
- [ ] implement BeGaze parser.
- [ ] Integration test for DIDEC passes.

## Failure Information (for bugs)
the DIDEC dataset currently fails because there is currently no parser for BeGaze datasets.

### Steps to Reproduce
install pymovements from source via 

```console
git clone git@github.com:aeye-lab/pymovements
cd pymovements
pip install -e .
```

then execute the following script:

```python
import pymovements as pm


dataset = pm.Dataset('DIDEC', path='data')
dataset.download()
dataset.load()
```

the dataset does download and extract but then unfortunately errors:

```console
polars.exceptions.ComputeError: could not parse `# Message: UE-keypress N` as dtype `f64` at column 'column_4' (column number 4)
```

more details here:

<details>
```console
Traceback (most recent call last):
  File "/home/siqube/lab/aeye-lab/pymovements/tt.py", line 11, in <module>
    dataset.load()
  File "/home/siqube/lab/aeye-lab/pymovements/src/pymovements/dataset/dataset.py", line 131, in load
    self.load_gaze_files(
  File "/home/siqube/lab/aeye-lab/pymovements/src/pymovements/dataset/dataset.py", line 207, in load_gaze_files
    self.gaze = dataset_files.load_gaze_files(
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/siqube/lab/aeye-lab/pymovements/src/pymovements/dataset/dataset_files.py", line 257, in load_gaze_files
    gaze_df = load_gaze_file(
              ^^^^^^^^^^^^^^^
  File "/home/siqube/lab/aeye-lab/pymovements/src/pymovements/dataset/dataset_files.py", line 374, in load_gaze_file
    gaze_df = from_csv(
              ^^^^^^^^^
  File "/home/siqube/lab/aeye-lab/pymovements/src/pymovements/gaze/io.py", line 218, in from_csv
    gaze_data = pl.read_csv(file, **read_csv_kwargs)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/siqube/lab/aeye-lab/pymovements/venv/lib/python3.12/site-packages/polars/_utils/deprecation.py", line 92, in wrapper
[...]

PR #1013, opened 2025-03-07
feat: Parse SMI BeGaze raw data files

### LINKED ISSUE #981: implement BeGaze parser.
sample files can be found when downloading the `DIDEC` dataset

saeub:
## Description

Read BeGaze text files. Not aiming for full support for now (I don't have enough data for testing), but at least the raw samples and events in DIDEC should be parsed successfully (#980).

#945 should be merged first (so that events can be implemented and tested the same way as for EyeLink)

Fixes #981

## Implemented changes

- [ ] Implement `parse_begaze()` like `parse_eyelink()`
- [ ] Implement `from_begaze()` like `from_asc()`
- [ ] :question: Figure out a way to route BeGaze files through `from_begaze()` when loading datasets, even if the file extension is `.txt` (currently they are routed through `from_csv()`, not sure if this behavior is required by other datasets)

## Type of change

## How Has This Been Tested?

- [x] `test_parse_begaze()`
- [ ] Update `test_parse_begaze()` after #945 is merged
- [ ] DIDEC tests

## Checklist:

### DIFF
--- a/tests/unit/gaze/_utils/_parsing_test.py
+++ b/tests/unit/gaze/_utils/_parsing_test.py
@@ -104,6 +104,77 @@ EBLINK R 10000018	10000020	2
 END	10000022 	SAMPLES	EVENTS	RES	  38.54	  31.12
 """
 
+BEGAZE_TEXT = r"""
+## [BeGaze]
+## Converted from:	C:\test.idf
+## Date:	08.03.2023 09:25:20
+## Version:	BeGaze 3.7.40
+## IDF Version:	9
+## Sample Rate:	1000
+## Separator Type:	Msg
+## Trial Count:	1
+## Uses Plane File:	False
+## Number of Samples:	11
+## Reversed:	none
+## [Run]
+## Subject:	P01
+## Description:	Run1
+## [Calibration]
+## Calibration Area:	1680	1050
+## Calibration Point 0:	Position(841;526)
+## Calibration Point 1:	Position(84;52)
+## Calibration Point 2:	Position(1599;52)
+## Calibration Point 3:	Position(84;1000)
+## Calibration Point 4:	Position(1599;1000)
+## Calibration Point 5:	Position(84;526)
+## Calibration Point 6:	Position(841;52)
+## Calibration Point 7:	Position(1599;526)
+## Calibration Point 8:	Position(841;1000)
+## [Geometry]
+## Stimulus Dimension [mm]:	474	297
+## Head Distance [mm]:	700
+## [Hardware Setup]
+## System ID:	IRX0470703-1007
+## Operating System :	6.1
+## IView X Version:	2.8.26
+## [Filter Settings]
+## Heuristics:	False
+## Heuristics Stage:	0
+## Bilateral:	True
+## Gaze Cursor Filter:	True
+## Saccade Length [px]:	80
+## Filter Depth [ms]:	20
+## Format:	LEFT, POR, QUALITY, PLANE, MSG
+##
+Time	Type	Trial	L POR X [px]	L POR Y [px]	L Pupil Diameter [mm]	Timing	Pupil Confidence	R Plane	Info	R Event Info	Stimulus
+10000000123	SMP	1	850.71	717.53	714.00	0	1	1	1	-	test.bmp
+10000000234	MSG	1	# Message: START_A
+10000002123	SMP	1	850.71	717.53	714.00	0	1	1	1	-	test.bmp
+10000003234	MSG	1	# Message: STOP_A
+10000004123	SMP	1	850.71	717.53	714.00	0	1	1	1	-	test.bmp
+10000004234	MSG	1	# Message: METADATA_1 123
+10000005234	MSG	1	# Message: START_B
+10000006123	SMP	1	850.71	717.53	714.00	0	1	1	1	-	test.bmp
+10000007234	MSG	1	# Message: START_TRIAL_1
+10000008123	SMP	1	850.71	717.53	714.00	0	1	1	1	-	test.bmp
+10000009234	MSG	1	# Message: STOP_TRIAL_1
+10000010234	MSG	1	# Message: START_TRIAL_2
+10000011123	SMP	1	850.71	717.53	714.00	0	1	1	1	-	test.bmp
+10000012234	MSG	1	# Message: STOP_TRIAL_2
+10000013234	MSG	1	# Message: START_TRIAL_3
+10000014234	MSG	1	# Message: METADATA_2 abc
+10000014235	MSG	1	# Message: METADATA_1 456
+10000014345	SMP	1	850.71	717.53	714.00	0	1	1	1	-	test.bmp
+10000015234	MSG	1	# Message: STOP_TRIAL_3
+10000016234	MSG	1	# Message: STOP_B
+10000017234	MSG	1	# Message: METADATA_3
+10000017345	SMP	1	850.71	717.53	714.00	0	0	-1	-1	-	test.bmp
+10000019123	SMP	1	850.71	717.53	714.00	0	0	-1	-1	Blink	test.bmp
+10000020123	SMP	1	850.71	717.53	714.00	0	0	-1	-1	Blink	test.bmp
+10000021123	SMP	1	850.71	717.53	714.00	0	0	1	-1	-	test.bmp
+"""  # noqa: E501
+
+
 PATTERNS = [
     {
         'pattern': 'START_A',
@@ -150,7 +221,7 @@ EXPECTED_DF = pl.from_dict(
     },
 )
 
-EXPECTED_METADATA = {
+EXPECTED_METADATA_EYELINK = {
     'weekday': 'Wed',
     'month': 'Mar',
     'day': 8,
@@ -191,6 +262,26 @@ EXPECTED_METADATA = {
     'metadata_4': None,
 }
 
+# TODO: Add more metadata
+EXPECTED_METADATA_BEGAZE = {
+    'sampling_rate': 1000.00,
+    'tracked_eye': 'L',
+    'data_loss_ratio_blinks': 0.18181818181818182,
+    'data_loss_ratio': 0.2727272727272727,
+    'total_recording_duration_ms': 11,
+    'datetime': datetime.datetime(2023, 3, 8, 9, 25, 20),
+    'blinks': [{
+        'duration_ms': 2,
+        'num_samples': 2,
+        'start_timestamp': 10000019.123,
+        'stop_timestamp': 10000021.123,
+    }],
+    'metadata_1': '123',
+    'metadata_2': 'abc',
+    'metadata_3': True,
+    'metadata_4': None,
+}
+
 
 def test_parse_eyelink(tmp_path):
     filepath = tmp_path / 'sub.asc'
@@ -203,7 +294,7 @@ def test_parse_eyelink(tmp_path):
     )
 
     assert_frame_equal(df, EXPECTED_DF, check_column_order=False)
-    assert metadata == EXPECTED_METADATA
+    assert metadata == EXPECTED_METADATA_EYELINK
 
 
 @pytest.mark.parametrize(
@@ -918,3 +1009,17 @@ def test_parse_eyelink_encoding(tmp_path, bytestring, encoding, expected_text):
     )
 
     assert parsed_metadata['text'] == expected_text
+
+
+def test_parse_begaze(tmp_path):
+    filepath = tmp_path / 'sub.txt'
+    filepath.write_text(BEGAZE_TEXT)
+
+    df, metadata = pm.gaze._utils.parsing.parse_begaze(
+        filepath,
+        patterns=PATTERNS,
+        metadata_patterns=METADATA_PATTERNS,
+    )
+
+    assert_frame_equal(df, EXPECTED_DF, check_column_order=False)
+    assert metadata == EXPECTED_METADATA_BEGAZE


cbueth:
I have pushed some commits and updated the PRs text, trying to reach an MVP for some of the features. @saeub, if you have availability, feel free to read through.

src/pymovements/gaze/io.py
```diff
@@ -746,3 +751,133 @@ def _fill_experiment_from_parsing_metadata(
...
+
+    if column_schema_overrides is not None:
+        samples = samples.with_columns([
+            pl.col(fileinfo_key).cast(fileinfo_dtype)
+            for fileinfo_key, fileinfo_dtype in column_schema_overrides.items()
+        ])
+
+    # Fill experiment with parsed metadata.
+    # BeGaze exports do not provide the same metadata keys as EyeLink ASC. Use a
+    # minimal, tolerant fill here instead of the strict ASC-specific function.
```
dkrako: can you create a similar method like `_fill_experiment_from_parsing_metadata`? that function warns in case an experiment field is not None and would be overwritten with a different value.

you can call the function `_fill_experiment_from_parsing_begaze_metadata` and rename the old one to ``_fill_experiment_from_parsing_eyelink_metadata`

tests/unit/gaze/_utils/_parsing_begaze_test.py
```diff
@@ -0,0 +1,491 @@
...
+    assert_frame_equal(
+        ev_actual.select(common_cols),
+        BEGAZE_EXPECTED_EVENT_DF.select(common_cols),
+        check_column_order=False,
+        rtol=0,
+    )
+
+    # Experiment should be filled from metadata (sampling_rate) and metadata attached to gaze.
+    assert pytest.approx(gaze.experiment.sampling_rate, rel=0, abs=1e-9) == 1000.0
+    assert gaze._metadata == EXPECTED_METADATA_BEGAZE
```
dkrako: there seem to be no tests for filling in the experiment.

can you create a parametrized test for this like this? 

```python
@pytest.mark.parametrize(
    ('header', 'kwargs', 'expected_experiment'),
    [
        pytest.param(
            '## Sample Rate:	1000',
            {},
            Experiment(sampling_rate=1000),
            id='sampling_rate_1000',
        ),
    ],
)
def test_from_begaze_loads_expected_experiment(header, kwargs, expected_experiment, make_text_file):
```

This way it's clearer which fields are implemented and each test is more specific.

To prevent missing metadata warnings you might consider one of the two options:
- ignore missing metadata warnings in relevant test functions
- create a `make_begaze_file` fixture with a minimal header.

tests/unit/gaze/_utils/_parsing_begaze_test.py
```diff
@@ -0,0 +1,491 @@
...
+        'pupil': [
+            714.0, 714.0, 714.0, 714.0, 714.0, 714.0, 714.0, 714.0, np.nan, 0.0,
+            0.0,
+        ],
+        'task': [None, 'A', None, 'B', 'B', 'B', 'B', None, None, None, None],
+        'trial_id': [None, None, None, None, '1', '2', '3', None, None, None, None],
+    },
+)
+
+BEGAZE_EXPECTED_EVENT_DF = pl.from_dict(
```
dkrako, line 162: I think this is fine for a start but unfortunately testing that way is inflexible when extending the feature set. this issue already persists in the existing eyelink parsing tests, leading to changes in test inputs to unrelated functions etc.

Using parametrization and fixtures is more flexible as you can create very minimal test data on the fly.

src/pymovements/gaze/_utils/parsing_begaze.py
```diff
@@ -0,0 +1,578 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+"""BeGaze parsing module.
+
+This module exists to provide a dedicated namespace for BeGaze-specific parsing
+logic.
+
+Public API:
```
dkrako: actually the whole parent module `_utils` is not part of the api

src/pymovements/gaze/_utils/_parsing_begaze.py
```diff
@@ -0,0 +1,578 @@
...
+        metadata_patterns: list[dict[str, Any] | str] | None = None,
+        encoding: str = 'ascii',
+        prefer_eye: str = 'L',
+) -> tuple[pl.DataFrame, pl.DataFrame, dict[str, Any]]:
+    """Parse BeGaze raw data export file.
+
+    Minimal "good enough" support for DIDEC-like exports:
+    - Parses sample rows (`SMP`) using the tabular header when available; falls back to a
+      LEFT-only regex for older/simpler files.
+    - Supports monocular and binocular files by selecting one eye via `prefer_eye`.
```
dkrako, line 117: I'm unsure if I fully understand: full binocular support would mean that we have x/y-coordinates for both files. It seems this is not implemented, right?

src/pymovements/gaze/io.py
```diff
@@ -746,3 +751,133 @@ def _fill_experiment_from_parsing_metadata(
...
+    gaze = Gaze(
+        samples=samples,
+        experiment=experiment,
+        events=Events(event_data),
+        trial_columns=trial_columns,
+        time_column='time',
+        time_unit='ms',
+        pixel_columns=detected_pixel_columns,
+    )
+    gaze._metadata = metadata  # pylint: disable=protected-access
```
dkrako: It would be so great if we could get rid of this private field as soon as possible. The metadata fields should either be integrated as a distinct field in `Gaze` (e.g., `calibrations`) or we should provide preprocessing methods that calculate them (e.g. `blink_ratio`).

Let's keep it in this PR, but your work would have been much easier if you didn't need to care about these things. sorry for that. see #893

src/pymovements/gaze/_utils/parsing_begaze.py
```diff
@@ -0,0 +1,578 @@
...
+
+    for idx, line in enumerate(lines):
+        if line.startswith('##'):
+            # extract simple key/value pairs
+            if line.startswith('## Date:'):
+                # format: ## Date:\tDD.MM.YYYY HH:MM:SS
+                try:
+                    _, value = line.split(':', 1)
+                    value = value.strip().strip('\n')
+                    if value.startswith('\t'):
```
dkrako: this is really hard to follow, can you use a regular expression here?

src/pymovements/gaze/_utils/parsing_begaze.py
```diff
@@ -0,0 +1,578 @@
...
+                        value = value[1:]
+                    value = value.replace('\t', ' ').strip()
+                    header_datetime = datetime.datetime.strptime(value, '%d.%m.%Y %H:%M:%S')
+                except (ValueError, TypeError):
+                    header_datetime = None
+            elif line.startswith('## Sample Rate:'):
+                try:
+                    _, value = line.split(':', 1)
+                    value = value.strip().strip('\n').replace('\t', ' ').strip()
```
dkrako: I'd very much prefer a regular expression here.

If you keep a dictionary of regular expressions that include named groups, you can create a more generic implementation:

```python
            elif line.startswith('##'):
                try:
                    line_metadata = _parse_begaze_meta_line(line)
                    metadata.update(line_metadata)
                except ...
```

with

```python
def _parse_begaze_meta_line(line: str) -> dict[str, Any]:
    for regex in BEGAZE_META_REGEXES:
        if match := regex.match(line):
            return match.groupdict()  # you could also do some casting / strptime / processing here
    return {}
```

and

```python
BEGAZE_META_REGEXES = [
        r'##\s+Date:\s+(?P<date>.*)\s+',
        r'##\s+Sample Rate:\s+(?P<sampling_rate>.*)\s+',
]

File: src/pymovements/gaze/_utils/parsing_begaze.py [hunk 9]
```diff
@@ -0,0 +1,578 @@
...
+    for idx, line in enumerate(lines):
+        if line.startswith('##'):
+            # extract simple key/value pairs
+            if line.startswith('## Date:'):
+                # format: ## Date:\tDD.MM.YYYY HH:MM:SS
+                try:
+                    _, value = line.split(':', 1)
+                    value = value.strip().strip('\n')
+                    if value.startswith('\t'):
+                        value = value[1:]
```
dkrako: we made the experience with eyelink files that (depending on model, version) their output is not always strict in their whitespace formatting (there might be a space instead of a tab although documented otherwise).

a regular expression would be more robust, performant and readable.

src/pymovements/gaze/_utils/_parsing_begaze.py
```diff
@@ -0,0 +1,578 @@
...
+                    'duration_ms': blink_last_ts - blink_start_prev_ts,
+                    'num_samples': blink_sample_count,
+                    'start_timestamp': blink_start_prev_ts,
+                    'stop_timestamp': blink_last_ts,
+                })
+            current_event = '-'
+            current_event_onset = None
+            current_event_additional = {key: {} for key in current_event_additional}
+
+    else:
```
dkrako: a lot of this code is duplicated between those branches. could you refactor this?

cbueth: Sure, will do!

src/pymovements/gaze/_utils/parsing_begaze.py
cbueth: Ok, important hint!

src/pymovements/gaze/_utils/parsing_begaze.py
cbueth: Thank you for the example code 👍

src/pymovements/gaze/_utils/_parsing_begaze.py
cbueth: The changes support binocular files, but only selecting and importing on eye, this might not be considered 'full', true.

src/pymovements/gaze/io.py
cbueth: Done!

cbueth: See

https://github.com/pymovements/pymovements/blob/79e0046ccdbd397336a54858fc4c64a9b8212c49/src/pymovements/gaze/io.py#L752-L763

tests/unit/gaze/_utils/_parsing_begaze_test.py
cbueth: I restructured the BeGaze testing partially, adding few more cases.

src/pymovements/gaze/_utils/parsing_begaze.py
cbueth: True, it is not really public, just locally speaking.

src/pymovements/gaze/io.py
cbueth: I removed BeGaze’s functional dependency on `gaze._metadata`. Now, `from_begaze` integrates key values into `Experiment`. Let me know if you prefere them somewhere else.

src/pymovements/gaze/_utils/parsing_begaze.py
cbueth: https://github.com/pymovements/pymovements/blob/6678a7df1e20d249a380f5e5d9f25fa9d63649f5/src/pymovements/gaze/_utils/parsing_begaze.py#L43-L55

src/pymovements/gaze/_utils/_parsing_begaze.py
cbueth: Adding `_maybe_finalize_blink_meta()`and `_finalize_current_event()` should remove most redundancy.

tests/unit/gaze/_utils/_parsing_begaze_test.py
cbueth: See

https://github.com/pymovements/pymovements/blob/8dc4d135dfa77fecb6b240845792c6ff6daebc7c/tests/unit/gaze/_utils/_parsing_begaze_test.py#L638-L640

cbueth:
> Wow this PR is massive and is taking a great direction!

Thank you for the valuable review!

> I think there are some things that could be improved, but as this PR is so big, I would be in favor to focus on the easy tasks to merge this early and improve on it later on.

Yes, I was unsure due to the large scope of the PR, not making it even bigger. The clean testing suffered a bit under this.

> One thing that would be really great is full binocular support, as I see it's just monocular currently. But that could be improved in a follow up.

Actual binocular support would be nice. I do think it would be fine to have a breaking change when adding binocular support, naturally delivering both eyes, unless selecting only one. As of now there is the `BEGAZE_SAMPLE` regex fallback case, where we read in the file format without the header, assuming a fixed format, we might want to drop that for simplicity. We would need more tests, e.g. one eye blinks while the other eye has valid data. We can do it like eyelink and have two event rows including `eye` column, or keep the current single stream, adding suffix in `name` and `eye` column.

cbueth:
Agreed, after having this run with the DIDEC tests, I would like to double-check the implementation with the DIDEC files. I will change the draft status when it's done.

src/pymovements/gaze/_utils/_parsing.py
saeub, line 1: I just noticed that in the other private `_utils` packages (e.g., `pm._utils`), all the modules also start with an underscore. Should we rename these files `_parsing.py`, `_parsing_begaze.py` etc. for consistency?

src/pymovements/gaze/_utils/parsing_begaze.py
```diff
@@ -0,0 +1,557 @@
...
+    """Parse a single BeGaze '##' metadata line using predefined regexes.
+
+    Returns an empty dict when the line does not match any known pattern.
+    Performs light casting for known fields (e.g., float for sampling rate, datetime for date).
+    """
+    for regex in BEGAZE_META_REGEXES:
+        if match := regex.match(line):
+            groupdict = match.groupdict()
+            # Casting and processing for known fields
+            if 'sampling_rate' in groupdict and groupdict['sampling_rate'] is not None:
```
saeub: Can we simplify this to:
```suggestion
            if groupdict.get('sampling_rate') is not None:
```

src/pymovements/gaze/_utils/parsing_begaze.py
```diff
@@ -0,0 +1,557 @@
...
+                try:
+                    groupdict['sampling_rate'] = float(groupdict['sampling_rate'])
+                except ValueError:
+                    pass
+            if 'date' in groupdict and groupdict['date']:
```
saeub: Same here:
```suggestion
            if groupdict.get('date') is not None:
```
(unless there is a possibility that `groupdict['date'] == ""`)

src/pymovements/gaze/_utils/_parsing_begaze.py
```diff
@@ -0,0 +1,557 @@
...
+                _maybe_finalize_blink_meta()
+
+                _finalize_current_event()
+                current_event = event
+                current_event_onset = timestamp
+                current_event_additional[current_event] = {**current_additional}
+            previous_timestamp = timestamp
+
+        # add last event (header-parsing branch)
+        if current_event != '-':
```
saeub: Is this check necessary? `_finalize_current_event()` and `_maybe_finalize_blink_meta()` already seem to check the `current_event`? (also for the `else` branch below)

src/pymovements/gaze/_utils/parsing_begaze.py
```diff
@@ -0,0 +1,557 @@
...
+                current_event_additional[current_event] = {**current_additional}
+            previous_timestamp = timestamp
+
+        # add last event (header-parsing branch)
+        if current_event != '-':
+            _finalize_current_event()
+            _maybe_finalize_blink_meta()
+            current_event = '-'
+            current_event_onset = None
+            current_event_additional = {key: {} for key in current_event_additional}
```
saeub: Are the three lines above necessary, since this is the last event? (also for the `else` branch below)

src/pymovements/gaze/io.py
```diff
@@ -746,3 +747,190 @@ def _fill_experiment_from_parsing_metadata(
...
+
+    # Tracked eye flags if present (metadata may provide 'L'/'R')
+    tracked = (metadata.get('tracked_eye') or '')
+    left_parsed = 'L' in tracked
+    right_parsed = 'R' in tracked
+
+    if experiment.eyetracker.left is None:
+        experiment.eyetracker.left = left_parsed
+    elif experiment.eyetracker.left != left_parsed:
+        warnings.warn(
```
saeub, line 865: In `_fill_experiment_from_parsing_eyelink_metadata()`, we are raising a `ValueError` with all the data mismatches, here it's just warnings. Should we make this consistent for the two functions? I don't have a preference for either, but when I implemented these checks for EyeLink in #884, @dkrako said he would prefer warnings: https://github.com/pymovements/pymovements/pull/884#discussion_r1833450965 So maybe change the EyeLink ones to warnings too?

tests/unit/dataset/dataset_files_test.py
```diff
@@ -335,7 +335,7 @@ def test_load_gaze_file_unsupported_load_function(make_example_file):
     msg, = exc.value.args
     assert msg == (
         'Unsupported load_function "from_a_land_down_under". '
```
saeub, line 337: I had forgotten about this, and it brought a smile to my face :)

saeub (commented):
Thanks a lot for adopting this neglected child of mine @cbueth! :pray:

Just added a few minor comments and questions.

src/pymovements/gaze/_utils/_parsing.py
cbueth: Good point, will do! Not sue if this is necessary, as the folder already has an underscore, but I'll do it for homogeneity.

src/pymovements/gaze/io.py
cbueth: Good point. This could be done in another PR:
- Change `_fill_experiment_from_parsing_eyelink_metadata()` to mirror BeGaze: warn on differences, only error on truly missing, required fields.
- Update tests

src/pymovements/gaze/_utils/_parsing_begaze.py
cbueth: It is not strictly needed for no-op if `_finalize_current_event()` already checks the state, but cheap and a safeguard. But calls to `_finalize_current_event()` and `_maybe_finalize_blink_meta()` are needed at the very end to flush a last ongoing event/blink if the file ends while an event is open.
I removed the unneeded resets in d89d6ffc.

src/pymovements/gaze/_utils/parsing_begaze.py
cbueth: Added it in ccbf5469.

src/pymovements/gaze/_utils/parsing_begaze.py
cbueth: Added it in ccbf5469.

src/pymovements/gaze/_utils/_parsing_begaze.py
```diff
@@ -0,0 +1,463 @@
...
+    Returns
+    -------
+    tuple[pl.DataFrame, pl.DataFrame, dict[str, Any]]
+        A tuple containing the parsed gaze sample data, the parsed event data, and the metadata.
+    """
+    # pylint: disable=too-many-branches, too-many-statements, too-many-nested-blocks
+    msg_prefix = r'\d+\tMSG\t\d+\t# Message:\s+'
+
+    # consume unused argument to satisfy linters (schema is still unused)
+    _ = schema
```
cbueth, line 158: Fyi, as of now, `schema` is accepted, but not used.

src/pymovements/gaze/_utils/_parsing.py
dkrako: I agree that the second leading underscore is redundant. In this PR it's probably best to stay consistent, but feel free to open a follow-up that renames all the private modules inside private modules.

src/pymovements/gaze/_utils/_parsing_begaze.py
```diff
@@ -0,0 +1,463 @@
...
+import warnings
+
+import numpy as np
+import polars as pl
+
+from pymovements.gaze._utils._parsing import compile_patterns, get_pattern_keys, \
+    check_nan, _calculate_data_loss_ratio
+
+# Regular expressions for BeGaze header metadata lines with named groups.
+# Note: we compile regexes for performance and to support minor whitespace variations.
```
dkrako, line 48: actually, I think that we increase import times with that, but we can handle that separately.
I created an issue for this in #1380

src/pymovements/gaze/_utils/_parsing_begaze.py
```diff
@@ -0,0 +1,463 @@
...
+    """
+    for regex in BEGAZE_META_REGEXES:
+        if match := regex.match(line):
+            groupdict = match.groupdict()
+            # Casting and processing for known fields
+            if groupdict.get('sampling_rate') is not None:
+                try:
+                    groupdict['sampling_rate'] = float(groupdict['sampling_rate'])
+                except ValueError:
+                    pass
```
dkrako, line 69: this should probably raise a warning, although a warning may be better suited inside the `from_asc()` function as that one is public (although we would warn in a private helper function anyway so it doesn't really matter).

Regardless, I think the approaches for `datetime` and `sampling_rate` should be consistent.

We should probably discuss where we want to cast and warn, I'm fine with both.

src/pymovements/gaze/_utils/_parsing_begaze.py
```diff
@@ -0,0 +1,463 @@
...
+    - Converts `Time` from microseconds to milliseconds (float).
+    - Derives BeGaze per-sample event labels (Fixation/Saccade/Blink) into consolidated
+      event rows named `<lowercase>_begaze` with onset/offset in milliseconds.
+    - Applies user-supplied `patterns` and `metadata_patterns` (same semantics as EyeLink)
+      to populate additional columns on samples/events and to extract metadata.
+
+    Parameters
+    ----------
+    filepath: Path | str
+        file name of file to convert.
```
dkrako: file path

src/pymovements/gaze/_utils/_parsing_begaze.py
```diff
@@ -0,0 +1,463 @@
...
+            return row[header_idx['Info']]
+    return '-'
+
+
+def parse_begaze(
+        filepath: Path | str,
+        patterns: list[dict[str, Any] | str] | None = None,
+        schema: dict[str, Any] | None = None,
+        metadata_patterns: list[dict[str, Any] | str] | None = None,
+        encoding: str = 'ascii',
```
dkrako, line 110: the lines starting with `#` are only used via `patterns` in this PR, right?

in a follow-up it would be great to add the same `message_patterns` as in `parse_eyelink`

src/pymovements/gaze/_utils/_parsing_begaze.py
dkrako: you can then remove that argument from this PR and add it in a follow-up later on

src/pymovements/gaze/_utils/_parsing_begaze.py
```diff
@@ -0,0 +1,463 @@
...
+        if 'R Event Info' in header_idx:
+            return row[header_idx['R Event Info']]
+        if 'Info' in header_idx:
+            return row[header_idx['Info']]
+    return '-'
+
+
+def parse_begaze(
+        filepath: Path | str,
+        patterns: list[dict[str, Any] | str] | None = None,
```
dkrako, line 107: i would be in favor of adding a `*` after filepath

src/pymovements/gaze/_utils/_parsing_begaze.py
```diff
@@ -0,0 +1,463 @@
...
+
+    # Blink tracking for metadata (declared before helper functions for mypy)
+    blinks_meta: list[dict[str, Any]] = []
+    blink_active = False
+    blink_start_prev_ts: float | None = None
+    blink_last_ts: float | None = None
+    blink_sample_count = 0
+
+    # Small helpers to unify event handling across branches
+    def _finalize_current_event() -> None:
```
dkrako, line 211: I think the inner function approach is great for the code structure in this state machine

src/pymovements/gaze/_utils/_parsing_begaze.py
```diff
@@ -0,0 +1,463 @@
...
+                blink_start_prev_ts = previous_timestamp
+                blink_sample_count = 1
+            else:
+                blink_sample_count += 1
+        elif not np.isnan(x_pix) and not np.isnan(y_pix) and not np.isnan(pupil):
+            num_valid_samples += 1
+
+        # event segmentation
+        if event != current_event:
+            _maybe_finalize_blink_meta()
```
dkrako, line 445: side note: it would be great if we could have a method that transforms between sample segmentations and events. this way we could offload some logic from the parser to a preprocessing step inside `from_begaze`. that function could also be reused for event detection and processing.

src/pymovements/gaze/_utils/_parsing_begaze.py
```diff
@@ -0,0 +1,463 @@
...
+            line_meta = _parse_begaze_meta_line(line.rstrip('\n'))
+            if 'datetime' in line_meta:
+                header_datetime = line_meta['datetime']  # type: ignore[assignment]
+            if 'sampling_rate' in line_meta and isinstance(
+                    line_meta['sampling_rate'], (int, float),
+            ):
+                header_sampling_rate = float(line_meta['sampling_rate'])
+            # Merge any parsed metadata keys
+            for k, v in line_meta.items():
+                if k not in ('datetime',):
```
dkrako: why is 'datetime' excluded here?

src/pymovements/gaze/_utils/_parsing_begaze.py
```diff
@@ -0,0 +1,463 @@
...
+    header_row_index: int | None = None
+    header_cols: list[str] | None = None
+    header_idx: dict[str, int] = {}
+
+    for idx, line in enumerate(lines):  # pragma: no branch
+        if line.startswith('##'):
+            # Parse meta line via regexes with named groups
+            line_meta = _parse_begaze_meta_line(line.rstrip('\n'))
+            if 'datetime' in line_meta:
+                header_datetime = line_meta['datetime']  # type: ignore[assignment]
```
dkrako: is that variable and additional code necessary or couldn't we just store directly in `metdata`?

src/pymovements/gaze/_utils/_parsing_begaze.py
```diff
@@ -0,0 +1,463 @@
...
+            )
+            selected_eye = other_eye
+
+    # Decide if we can parse samples: requires tabular header with eye columns
+    parse_samples = False
+    if header_row_index is not None and header_cols is not None:
+        parse_samples = has_eye_columns('L') or has_eye_columns('R')
+
+    # iterate over data lines following the header row
+    assert header_row_index is not None
```
dkrako: is this assert necessary here? usually asserts in production code should assert things that should never happen regardless of any user input (e.g., loop invariants). if something fails due to invalid input a warning or exception should be raised.

src/pymovements/gaze/_utils/_parsing_begaze.py
```diff
@@ -0,0 +1,463 @@
...
+    for line in lines[header_row_index + 1:]:
+        # Apply message-driven additional columns first
+        for pattern_dict in compiled_patterns:
```
dkrako, line 339: in this case a dedicated preprocessing method could also help with this in the future and offload some logic.

What we would need for this is transforming `messages` to `samples` and `events` (related to the sidenote in #1290).

src/pymovements/gaze/_utils/_parsing_begaze.py
```diff
@@ -0,0 +1,463 @@
...
+
+        parts = [p.strip() for p in line.rstrip('\n').split('\t')]
+        if len(parts) < 3:
+            # also try metadata patterns on non-sample lines (use unprefixed compiled patterns)
+            for pattern_dict in compiled_metadata_patterns.copy():
+                if match := pattern_dict['pattern'].match(line):
+                    if 'value' in pattern_dict and 'key' in pattern_dict:
+                        metadata[pattern_dict['key']] = pattern_dict['value']
+                    else:
+                        metadata.update(match.groupdict())
```
dkrako: This should raise warning.

Moreover the `metadata_patterns` documentation doesn't address that recurring metadata is overwritten.

The behavior is more or less consistent with the argument in `parse_eyelink`, but that argument has shown to be not really useful in most situations (as most messages that you will be interested in can occur multiple times).

I think it's fine to leave out the `messages_pattern` in this PR but it should be added in a follow up.

src/pymovements/gaze/_utils/_parsing_begaze.py
```diff
@@ -0,0 +1,463 @@
...
+        previous_timestamp = timestamp
+
+    # add last event (header-parsing branch)
+    if current_event != '-':
+        _finalize_current_event()
+        _maybe_finalize_blink_meta()
+
+    # Finalise metadata for BeGaze
+    # total_recording_duration_ms per test equals number of samples for this minimal fixture
+    total_recording_duration_ms = len(samples['time']) if samples['time'] else 0
```
dkrako: I don't think this is how we define total recording duration. IMO it's the actual time between recording starts and stops.

The way it is defined here, it's not really possible to calculate a data loss ratio in case of missing samples.

From my side, it's totally fine to leave out these calculations completely in this PR and implement them as preprocessing methods (data quality measures) outside the parsing modules.

src/pymovements/gaze/io.py
dkrako: yes probably best to do this in a separate PR

src/pymovements/gaze/_utils/_parsing_begaze.py
cbueth: That makes sense, thanks for the comment. I keep it untouched for this PR.

src/pymovements/gaze/_utils/_parsing_begaze.py
cbueth: Agreed!

src/pymovements/gaze/_utils/_parsing_begaze.py
cbueth: Yes, this could be done in some follow-up PR. As example creating `samples_to_events()` and `messages_to_{samples,events}()` to be used in `from_begaze()`.

src/pymovements/gaze/_utils/_parsing_begaze.py
cbueth: I am removing the exclusion. Then `datetime` and `sampling_rate` will be written directly into `metadata` during the header pass.

src/pymovements/gaze/_utils/_parsing_begaze.py
cbueth: True, direct assign parse when `##` is better. I am changing it.

src/pymovements/gaze/_utils/_parsing_begaze.py
cbueth: I am replacing it with a `ValueError` for a clearer failure.

src/pymovements/gaze/_utils/_parsing_begaze.py
cbueth: Good point. For now I will keep the behaviour here.

src/pymovements/gaze/_utils/_parsing_begaze.py
cbueth: For this I am adding a `RuntimeWarning`, one-time and per-key overwrite warning through `_set_metadata_key()`.

src/pymovements/gaze/_utils/_parsing_begaze.py
cbueth: Thank you for calling this out. I remove it, better implemented as separate preprocessing.

PR #1023, opened 2025-03-12
fix: return DatasetDefinition copy by DatasetLibrary.get()

dkrako:
## Description

#914 exposed a bug in `resample()`: the experiment of the dataset definition is not just updated in the `GazeDataFrame` itself, but also in the `DatasetLibrary`. This really shouldn't happen.

## Implemented changes

- [ ] Change 1
- [ ] Change 2
- [ ] Change 3

## Type of change

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Checklist:

### DIFF
--- a/src/pymovements/dataset/dataset_library.py
+++ b/src/pymovements/dataset/dataset_library.py
@@ -63,6 +63,17 @@ class DatasetLibrary:
         """
         return cls.definitions[name]
 
+    @classmethod
+    def list(cls) -> list[str]:
+        """Get list of names of all added datasets in library.
+
+        Returns
+        -------
+        list[str]
+            List of dataset names that are included in the dataset library.
+        """
+        return list(cls.definitions.keys())
+
 
 DatsetDefinitionClass = TypeVar('DatsetDefinitionClass', bound=type[DatasetDefinition])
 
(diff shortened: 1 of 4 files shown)


dkrako:
I just don't get it...

How can `test_returned_definition_is_copy` fail if I clearly use `deepcopy()` in `DatasetLibrary.get()`?

dkrako:
Ah now I think I know why this happened. The returned definition was the class not an instance. #914 has changed this accordingly which is good. The same class `is` always the same, regardless of using deepcopy. 

Hope the tests will pas now.

src/pymovements/dataset/dataset.py
```diff
@@ -379,7 +379,7 @@ def apply(
 
         Use apply for upsampling, downsampling or making the sampling rate constant
         using resample:
-        >>> dataset.apply('resample', resampling_rate=2000)# doctest:+ELLIPSIS
+        >>> dataset.apply('resample', resampling_rate=1000)# doctest:+ELLIPSIS
```
SiQube, line 382: sampling rate of ToyDataset is 1000, here nothing happens if we change this to 1000 ?

SiQube (commented):
one minor comment

src/pymovements/dataset/dataset.py
dkrako: good catch, I thought I had to revert a line from #914 but it was already reverted there and 2000 was the original value.

I reverted this change

PR #1027, opened 2025-03-12
feat: implement DatasetLibrary.names as class property

dkrako:
## Description

followup to #1022 

inspired by https://stackoverflow.com/questions/76249636/class-properties-in-python-3-11

there's really some next level magic involved. need to level up my skill tree.

### DIFF
--- a/src/pymovements/dataset/dataset_library.py
+++ b/src/pymovements/dataset/dataset_library.py
@@ -20,7 +20,6 @@
 """DatasetLibrary module."""
 from __future__ import annotations
 
-from collections.abc import Callable
 from typing import Any
 from typing import TypeVar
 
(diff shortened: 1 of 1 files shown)


dkrako:
pylint is complaining:

```
src/pymovements/dataset/dataset_library.py:81: [E0213(no-self-argument), DatasetLibrary.names] Method 'names' should have "self" as first argument
```

This one would be easy to fix: either rename `cls` to `self` or ignore `no-self-argument`.

What's worse is this:
```
tests/unit/dataset/dataset_library_test.py:60: [E1133(not-an-iterable), test_list_names_is_list_of_str] Non-iterable value names is used in an iterating context
```

This means pylint doesn't get that this is a property, not simply a method.

I'll try [this approach](https://stackoverflow.com/a/13624858) soon:

```python
class classproperty(property):
    def __get__(self, owner_self, owner_cls):
        return self.fget(owner_cls)
```

PR #1031, opened 2025-03-13
depr: deprecate `utils.strings`

### LINKED ISSUE #999: deprecate utils modules that are not intended for end-users
## Description of the problem
Most of our utils modules are not really intended to be used by end-users , but should be used via high-level interfaces that call these methods.

This follows an anti-pattern where there is high coupling between the `utils` module and other modules and reduced cohesion inside modules like `dataset`.

Also this crowds our online documentation with irrelevant functions no end-user will need.

## Description of a solution

Instead, we should increase the cohesion inside each module and reduce coupling between modules.

Candidates are:
- [x] `utils.aois` #1080 
- [x] `utils.archives` -> `dataset._utils.archives` #1028 
- [x] `utils.checks` -> `_utils._checks` #1094 
- [x] remove `utils.decorators` #1008 
- [x] `utils.downloads` -> `dataset._utils.downloads` #1030 
- [x] `utils.filters` -> `events._utils._filters` #1032
- [x] `utils.parsing` -> `gaze._utils.parsing` #976 
- [x] `utils.paths` -> `_utils._paths` #1029 
- [ ] `utils.plotting` -> `plotting._utils.matlab` #1081 
- [x] `utils.strings` -> `_utils._strings` #1031

dkrako:
## :warning: Deprecation

`utils.strings` is deprecated and will be removed in pymovements v0.26.0

## Description

Deprecate `utils.strings` as none of the included are intended for the end-user.

The following warning is printed:

```
DeprecationWarning: Call to deprecated function (or staticmethod) curly_to_regex. (This module will be removed in v0.26.0.) -- Deprecated since version v0.21.0.
```

partly resolves #999

## Implemented changes

- [x] moved `utils/strings.py` to `_utils/_strings.py` and adjusted to new location
- [x] use `_utils.strings` in remaining source
- [x] created deprecated function in `utils/strings.py`
- [x] moved `tests/unit/utils/strings_test.py` to `tests/unit/_utils/_strings_test.py`
- [x] add test that functions are correctly deprecated
- [x] add test that functions are removed in planned version


src/pymovements/utils/parsing.py
```diff
@@ -19,9 +19,9 @@
 # SOFTWARE.
 """Module for parsing input data.
 
-.. deprecated:: v0.21.0
+.. deprecated:: v0.22.0
    Please use gaze.from_asc() instead of utils.parsing.
-   This module will be removed in v0.26.0.
+   This module will be removed in v0.27.0.
```
SiQube: why is parsing changed in this PR?

src/pymovements/stimulus/image.py
```diff
@@ -22,9 +22,9 @@
 
 from pathlib import Path
 
+from pymovements._utils._strings import curly_to_regex
```
SiQube, line 25: do we need the double underscore for both `_utils` and `_strings`? from my pov `utils` already implies no user interaction

src/pymovements/utils/parsing.py
```diff
@@ -37,7 +37,7 @@
 @deprecated(
     reason='Please use gaze.from_asc() instead of utils.parsing. '
            'This module will be removed in v0.26.0.',
-    version='v0.21.0',
+    version='v0.21.1',
```
SiQube, line 40: why is the version of parsing deprecated in this PR?

src/pymovements/utils/strings.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2022-2025 The pymovements Project Authors
+# Copyright (c) 2023-2025 The pymovements Project Authors
```
SiQube, line 1: why change the copyright year?

SiQube (commented):
thank you for your PR, just some minor questions/feedback

PR #1032, opened 2025-03-13
depr: deprecate `utils.filters`

### LINKED ISSUE #999: deprecate utils modules that are not intended for end-users

dkrako:
## :warning: Deprecation

`utils.filters` is deprecated and will be removed in pymovements v0.26.0

## Description

Deprecate `utils.filters` as none of the included are intended for the end-user.

The following warnings are printed:

```
DeprecationWarning: Call to deprecated function (or staticmethod) filter_candidates_remove_nans. (This function will be removed in v0.26.0.) -- Deprecated since version v0.21.0.
```

```
DeprecationWarning: Call to deprecated function (or staticmethod) events_split_nans. (This function will be removed in v0.26.0.) -- Deprecated since version v0.21.0.
```

partly resolves #999

## Implemented changes

- [x] moved `utils/filters.py` to `events/_utils/_paths.py` and adjusted to new location
- [x] use `events._utils.filters` in remaining source
- [x] created deprecated function in `utils/filters.py`
- [x] moved `tests/unit/utils/filters_test.py` to `tests/unit/events/_utils/_filters_test.py`
- [x] add test that functions are correctly deprecated
- [x] add test that functions are removed in planned version


src/pymovements/utils/parsing.py
```diff
@@ -19,9 +19,9 @@
 # SOFTWARE.
 """Module for parsing input data.
 
-.. deprecated:: v0.21.0
+.. deprecated:: v0.22.0
    Please use gaze.from_asc() instead of utils.parsing.
-   This module will be removed in v0.26.0.
+   This module will be removed in v0.27.0.
```
SiQube: I think this is the third change to `utils/parsing` I'm seeing in PRs submitted within a day. is there any story to it?

src/pymovements/utils/parsing.py
```diff
@@ -37,7 +37,7 @@
 @deprecated(
     reason='Please use gaze.from_asc() instead of utils.parsing. '
            'This module will be removed in v0.26.0.',
-    version='v0.21.0',
+    version='v0.21.1',
```
SiQube, line 40: why is `utils/parsing` deprecated in this PR?

SiQube (commented):
thank you for your PR, it seems to me that you did not 

`git checkout origin/main -b NEWBRANCH` before creating the last few pull requests. doing so makes it much easier for a reviewer to see only the changes that you want to make in a pull request.

dkrako:
> thank you for your PR, it seems to me that you did not
> 
> `git checkout origin/main -b NEWBRANCH` before creating the last few pull requests. doing so makes it much easier for a reviewer to see only the changes that you want to make in a pull request.

I guess you mean the other PRs but I always do it this way where possible.

The thing with the last deprecation PRs is, that there's a dependency (downloads depends on archives depends on paths). If a PR depends on another PR it is not possible to create a branch from main, as the changes from the other PR need to be included.

src/pymovements/utils/parsing.py
dkrako: Parsing was already deprecated two weeks ago. #976 

What you see here is an autoreplace mistake. I had to change the version for all the current deprecation PRs because the PRs weren't merged for `v0.21.0`, so I had to replace the deprecations with `v0.21.1`.

I didn't notice that it also commited `parsing.py`. Good that you noticed!

I reverted this change.

dkrako:
In case of dependencies between PRs I advise you as a reviewer to do the following:
1. review the first PR in the dependency queue.
2. wait until the PR is merged.
3. merge main into the next PR and goto 1. for reviewing the next PR.

Step 2 can take a while (especially if there are changes requested), but that's unavoidable.

SiQube (approved):

PR #1028, opened 2025-03-13
depr: deprecate `utils.archives` in favor of `Dataset.extract()`

dkrako:
## :warning: Deprecation

`utils.archives` is deprecated. please use `Dataset.extract()` instead.
`utils.archives` will be removed in pymovements v0.26.0

## Description

Deprecate `utils.archives` in favor of `Dataset.extract()`

The following warning is printed:

```
DeprecationWarning: Call to deprecated function (or staticmethod) extract_archive. (Please use Dataset.extract() instead. This function will be removed in v0.26.0.) -- Deprecated since version v0.21.0.
```

partly resolves #999

blocked by #1029

## Implemented changes

- [x] moved `utils/archives.py` to `dataset/_utils/_archives.py` and adjusted to new location
- [x] use `dataset._utils.archives` in `dataset.dataset_download` 
- [x] created deprecated function in `utils/archives.py`
- [x] moved `tests/unit/utils/archives_test.py` to `tests/unit/dataset/_utils/_archives_test.py`
- [x] add test that function is correctly deprecated
- [x] add test that function is removed in planned version

PR #1034, opened 2025-03-13
perf: resume archive extraction by default

dkrako:
## Description

Take use of  #786 by default as I don't see any actual harm in this just benefits.

This is blocked by #1030 to prevent a breaking change.


SiQube (approved):
I think the change is great, I thought about changing it in #488 but didn't wanna break defaults. @dkrako there seems to be a merge conflict, can you resolve it? I already preapproved since the change is "minimal". 

also please try to use `git checkout origin/main -b NEWBRANCH` before submitting a PR, there is a lot of spam from previous PRs which do not need to be in this PR.

dkrako:
Thanks, I'll take care of the merge conflicts.

In this case it was not possible to `git checkout origin/main -b NEWBRANCH` as this PR depends on #1030 (which depends on #1028 which depends on #1029) otherwise it would introduce breaking changes to the utils interface (as `resume` was missing from `download_and_extract_archive()`).

SiQube:
then let's rebase #1030 into this branch after #1030 is merged

PR #1035, opened 2025-03-13
ci: change documentation emoji in release drafter

dkrako:
The bookmark emoji 🔖 doesn't look right.

I changed it to the open book emoji 📖


SiQube (approved):
good change!

PR #1046, opened 2025-03-14
feat: raise helpful error message for existing event properties

xiaotiansu:
Hi, my first commit of code, hope I'm doing it correctly :)

Fixes #

## Implemented changes

- [x] raise ValueError if the computed property already exists in the event dataframe

## Type of change

## How Has This Been Tested?

- [x] passed 579 tests for `dataset`: `tox -e py39 -- tests/unit/events`

## Checklist:


src/pymovements/dataset/dataset.py
```diff
@@ -763,6 +765,12 @@ def compute_event_properties(
             if column != 'filepath'
         ]
 
+        existing_columns = set(self.events[0].columns) & set(event_properties)
```
dkrako, line 768: `event_properties` can be passed either as a list of `str` or as a dictionary, that's why some of the tests fail:

```python
       existing_columns = set(self.events[0].columns) & set(event_properties)
       TypeError: unhashable type: 'dict'
```

Just include these lines before and then use `set(event_property_names)`:

```python
if isinstance(event_properties, dict):
   event_property_names = event_properties
elif isinstance(event_properties, dict):
   event_property_names = list(event_properties.keys())
else:
   raise TypeError(f'event_properties must be of type str or dict, but is {type(event_properties)}')
```

dkrako (changes requested):
Great! Thank you so much for your initiative!

Some of the tests don't pass because `event_properties` can also be a dict.
I posted a short proposal on how to change your code accordingly.

You will also need to add at least two test cases here:

https://github.com/aeye-lab/pymovements/blob/87487d0368cbf8919dc05b59c1706c8abc20764a/tests/unit/dataset/dataset_test.py#L1603-L1614

One for the `ValueError` and one for the `TypeError`.

src/pymovements/dataset/dataset.py
```diff
@@ -763,6 +770,25 @@ def compute_event_properties(
...
         ]
 
+        if isinstance(event_properties, str):
+            event_property_names = [event_properties]
+        elif isinstance(event_properties, list):
+            event_property_names = event_properties
+        elif isinstance(event_properties, dict):
+            event_property_names = event_properties.keys()
+        elif isinstance(event_properties, tuple):
+            event_property_names = event_properties[1].keys()
```
dkrako, line 780: I just noticed, that it's probably much easier to reuse the properties of the initialized processor:

https://github.com/aeye-lab/pymovements/blob/e7d45528e8d8cbb1ead6eda447e158fc40947bde/src/pymovements/events/processing.py#L124

Here it is already guaranteed that it's a list of tuples, where the first item of each tuple is the name of the property and the second are optional keyword arguments.
Example: `[('dispersion', {}), ('location', {'method': 'median'})]`

So this way getting all the names is much easier:

```python
event_property_names = [property[0] for property in processor.event_properties]
```

Sorry that I previously lead you to this more complicated code.

src/pymovements/dataset/dataset.py
```diff
@@ -749,12 +749,19 @@ def compute_event_properties(
...
             If specified event name ``name`` is missing from ``events``.
+        ValueError
+            If the computed property already exists in the event dataframe.
 
         Returns
         -------
         Dataset
             Returns self, useful for method cascading.
         """
+        if not isinstance(event_properties, (str, tuple, list, dict)):
```
dkrako, line 760: Can you move the type check here to the `EventGazeProcessor`?

https://github.com/aeye-lab/pymovements/blob/e7d45528e8d8cbb1ead6eda447e158fc40947bde/src/pymovements/events/processing.py#L104-L114

This way the checks are already in place if other function use the processor.

dkrako: Unfortunately the `event_properties` argument is not that well documented.

Just for clarification, if `event_properties` is given just as a tuple,
then this form is expected: `('location', {'method': 'median'})`

So the first if is for for the two ways of passing a single property, once simply as a string, once with additional kwargs.

If it's a list it would be nice to check that all items in the list are either `str` or a `tuple`.

All in all, it would be best to do the checking in the processor like this:

```python
if isinstance(event_properties, str): 
     event_properties = [event_properties]
elif isinstance(event_properties, tuple):
     # here check first, that tuple has a length of two, else raise ValueError
     # then check that first item is str, second item dict, else raise TypeError
     event_properties = [event_properties]
elif isinstance(event_properties, list):
    # here we just check  each list item for correct types
    for event_property in event_properties:
        if not isinstance(event_property, (str, tuple)):
            raise TypeError(...)
        if isinstance(event_property, tuple):
            # here check first, that tuple has a length of two, else raise ValueError
            # then check that first item is str, second item dict, else raise TypeError
else:
    raise TypeError(...) 
```

This way the error messages will better guide confused users on how to correctly pass the properties.

Also this would then guarantee type safety in the following for loop, where the `else` would lead to confusing errors if users for example pass a list of dicts:

https://github.com/aeye-lab/pymovements/blob/e7d45528e8d8cbb1ead6eda447e158fc40947bde/src/pymovements/events/processing.py#L107-L114

src/pymovements/events/processing.py
```diff
@@ -42,8 +42,46 @@ class EventProcessor:
     """
 
     def __init__(self, event_properties: str | list[str]):
-        if isinstance(event_properties, str):
-            event_properties = [event_properties]
+        def __init__(self, event_properties: str | list[str]):
```
dkrako, line 45: this line here is duplicated thus the init is actually never executed

src/pymovements/events/processing.py
```diff
@@ -41,9 +41,62 @@ class EventProcessor:
...
-    def __init__(self, event_properties: str | list[str]):
+    def __init__(self, event_properties: str | tuple[str, dict] | list[str | tuple[str, dict]]):
         if isinstance(event_properties, str):
             event_properties = [event_properties]
+        elif isinstance(event_properties, tuple):
+            if len(event_properties) != 2:
+                raise ValueError('Tuple must have a length of 2.')
+            if not isinstance(event_properties[0], str):
+                raise TypeError(
+                    f"First item of tuple must be a string, but received {
```
dkrako, line 52: this line leads to a syntax error on python 3.9-3.11. the curly braces must be on the same line for these python versions.

Try to format it like you wrote it in `Dataset.compute_event_properties()` then this should work.

dkrako (changes requested):
Great, this looks quite good already, but please make sure that all tests pass.

There are a few syntax errors left in your code, that's why the tests fail. The error messages in the test runs point you to the error: https://github.com/aeye-lab/pymovements/actions/runs/14032843428?pr=1046

SiQube:
this PR gets screwed over by autopep8, see https://github.com/hhatto/autopep8/issues/765 for the issue. maybe we should deactivate autopep8?

src/pymovements/dataset/dataset.py
```diff
@@ -772,7 +774,13 @@ def compute_event_properties(
             for column in self.fileinfo['gaze'].columns
             if column != 'filepath'
         ]
-
+        event_property_names = [property[0] for property in processor.event_properties]
+        existing_columns = set(self.events[0].columns) & set(event_property_names)
+        if existing_columns:
+            raise ValueError(
```
dkrako, line 780: This is not covered by any tests anymore. Please add a test here:

https://github.com/aeye-lab/pymovements/blob/6245bbfceacd018a6f880a4b8f0f3f44ba2acacd/tests/unit/dataset/dataset_test.py#L1603-L1619

src/pymovements/events/processing.py
```diff
@@ -212,3 +214,50 @@ def process(
...
+def _check_event_properties(
+        event_properties: str | tuple[str, dict[str, Any]] | list[str]
+        | list[str | tuple[str, dict[str, Any]]],
+) -> None:
+    """Validate event properties."""
+    if isinstance(event_properties, str):
+        pass
+    elif isinstance(event_properties, tuple):
+        if len(event_properties) != 2:
+            raise ValueError('Tuple must have a length of 2.')
```
dkrako, line 228: Most of the exceptions are not covered by tests. Please add tests for the following cases:

- [ ] `[('peak_velocity', {}, None)]` -> `raise ValueError('Tuple must have a length of 2.')`
- [ ] `[(1, {})]` -> `TypeError('First item of tuple must be a string ...')`
- [ ] `[('peak_velocity', 1]` -> `TypeError('Second item of tuple must be a dictionary ...`
- [ ] `[['peak_velocity', 1]]` -> `TypeError('Each item in the list must be either a string ...`
- [ ] `[('peak_velocity', {}), ('amplitude', {}, 1)]` -> `raise ValueError('Tuple must have a len...`
- [ ] `[[('peak_velocity', 1), (1, {})]]` -> `TypeError('First item of tuple must be a string ...`
- [ ] `[[('peak_velocity', 1]), ('amplitude', 1)]]` -> `TypeError('Second item of tuple must ...`
- [ ] `[1]` -> `TypeError('event_properties must be of type str, tuple, or list ...`


Here's where they need to go:

https://github.com/aeye-lab/pymovements/blob/6245bbfceacd018a6f880a4b8f0f3f44ba2acacd/tests/unit/events/processing_test.py#L115-L125

dkrako (changes requested):
Great, finally linting and all tests pass! Was a bit complicated with all the type unions. There seemed to be a confusion between `EventProcessor` and `EventGazeProcessor`, so I helped out a bit to not burden you with these details.

The last thing what's missing now is covering all the new exceptions with tests.

src/pymovements/dataset/dataset.py
dkrako: Hi @xiaotiansu maybe you overlooked this, there's one final test needed to get full coverage.

After adding the test we'll be ready to merge. Thanks for all the work so far!

dkrako (approved):
I finalized this PR accordingly and now we're ready to merge. Thank you for your work!

PR #1049, opened 2025-03-17
ci: run integration tests for all datasets in library

dkrako:
## Description

This runs a very simple integration test for all datasets in the matrix:

```python
    dataset = pm.Dataset(dataset_name, path=dataset_path)

    # Download and load in dataset.
    dataset.download()
    dataset.scan()

    assert len(dataset.fileinfo) > 0
```

I implemented this simplified test instead of the existing integration test, as we currently have a memory consumption problem and the GitHub workers would crash. This can be improved on after finishing #1005

The dataset matrix is dynamically created by this command in the workflow without needing to create any further files:

```
python -c "from pymovements import DatasetLibrary; print(f'matrix={DatasetLibrary.names()}')"
```

Only commits that have a version tag of the form `v[0-9]+.[0-9]+.[0-9]+*` will run the integration tests.

For the next releases I will:
1. create and push tag for release candidate e.g. `v0.21.0rc0`
2. github will automatically create the integration test jobs
3. if all tests pass, publish release and create tag e.g. `v0.21.0`
3a. if tests fail, resolve failing tests in hotfix PR. merge and tag new release candidate. (goto 1.)

This will run the integration tests twice (once for the release candidate, once for the release), but that should be alright I think.

We could also add something like this to run these tests on changes in dataset definitions:

```yaml
on:
  push:
    paths:
      - src/pymovements/datasets/*
```
But this would also mean that for each new commit on the PR all of the datasets will be downloaded, not just the changed one. This would be a waste of resources. I guess we can try to improve on this in the future.

It's always possible to trigger the integration tests on the most recent commit of a PR this way (may you will have to increase the pre-counter):

```bash
git checkout your/branch
git tag v0.21.0-pre
git push origin tag v0.21.0-pre1
```

## Implemented changes

- [x] prepare dataset matrix in dataset workflow
- [x] create separate github workers for testing each dataset
- [x] run integration tests on version tags only


dkrako:
Blocked by #1050

dkrako:
Extracting GazeBase seems to be just too much for the worker to complete during 6 hours. Probably due to the nested archives.

The previous GazeBase run failed because it exceeded the 6 hour limit. The last run failed because the worker has just too much to the stdout which lead to a lost connection.

I will simplify the test further and just download the resource. If it passes the md5 this should be perfectly sufficient for this PR. We can still do the preprocessing tests locally on our servers, especially for new datasets.

Additionally we could check if there's a way to make extraction even more efficient. Maybe if we encounter an archive inside of an archive, (`if recursive==True`) don't extract the archive but extract its members?

There's also an issue I noticed in the progressbar for the nested archives. There are hundreds of new progressbars created. This is what probably caused the lost communication in the last run. It could be much more efficient to update the original progress bar instead.

@SiQube what do you think about these suggestions?

dkrako:
the dataset integration tests were run for commit [e7b6b28](https://github.com/aeye-lab/pymovements/pull/1049/commits/e7b6b283f9f15cfdaa318c1ecdc6580a9ee04360)

SiQube (approved):

SiQube:
I think the next big step would be to optimize extraction to focus on the files in question and delete all unnecessary files

dkrako:
> I think the next big step would be to optimize extraction to focus on the files in question and delete all unnecessary files

I am not sure if I understand correctly. Deleting unnecessary files would be even more inefficient, as it adds even more disk io operations instead of preventing them (removing a file is a disk io op).

The issue here is, that GitHub workers have very slow disk io (hence `dataset_test.py` takes so long because we are writing dataset files to disk. this can be optimized by #1002 as then most of the dataset fixtures could be rewritten to not require any disk io).

Maybe you mean ignoring unnecessary files for extraction? This could be an additional approach. How would we make sure that files are not needed? How large is the proprotion of unnecessary files in GazeBase? Are these stimulus related files?

Anyway, I created #1060 for my previous proposal of reducing disk io by avoiding writing nested archives to disk.
Feel free to create an issue for your proposal.

PR #1050, opened 2025-03-17
fix: deduplicate resource filenames in UCL dataset

dkrako:
The UCL dataset fails the tests in #1049 if `remove_finished=True`.

Reason: resource filename is duplicated, hence archive is already removed before being extracted (again).

This is just a workaround before fixing the underlying issue of handling filename duplicates in resources.

Here's one of the failed jobs: https://github.com/aeye-lab/pymovements/actions/runs/13903514974/job/38903529945?pr=1049


SiQube (approved):

SiQube:
@dkrako I think it would be great if we could reuse previously downloaded archives, there is some duplicate downloads in some datasets.

PR #1052, opened 2025-03-19
chore: simplify public datasets tests

SiQube:
we havd a lot of copy pasted code which forced the user to enter the dataset twice to our tests.


dkrako:
I am very confused. It seems you have forgotten major points that we discussed (same with #1053).

The tests are needed until we remove the classes. They test if the classes are consistent with the yaml definitions.

We currently rely on these classes as we currently do not have any other way to create documentation pages for yaml-only datasets.

So the only way to move forward is creating dataset pages out of pure yaml dataset definitions.

Then we will deprecate the classes. And five versions later (which will take 6-12 months) we will remove the classes.
Only then we can start the discussion on removing the these consistency tests.

PR #1053, opened 2025-03-19
remove datasets.yaml

SiQube:
somehow I missed that `_add_shipped_datasets` relies on `datasets.yaml` which is kind of annoying because you're not able to add a dataset without tests failing.


dkrako:
`importlib.resources` doesn't have an `iterdir()` method, so there's no way that we can remove `datasets.yaml`.

The tests only work, because tox extracts all package archives during installation.

dkrako:
The only way forward to simplify the data contribution process is implementing autogeneration of the `datasets.yaml` (e.g., via a local pre-commit hook).

PR #1057, opened 2025-03-25
Dataset info

### LINKED ISSUE #987: fix erroneous docstrings in dataset definitions
Several docstrings of the new dataset definitions have copy paste errors.

Here's some examples:

- https://github.com/aeye-lab/pymovements/blob/38c3bcb8ac9141b6e6053696a3f315f5907dcc33/src/pymovements/datasets/daemons.py#L35 wrong citation
- https://github.com/aeye-lab/pymovements/blob/38c3bcb8ac9141b6e6053696a3f315f5907dcc33/src/pymovements/datasets/bsc.py#L108 wrong names in example

SiQube:
add information about the dataset as a property.  

additionally, resolves #987. 

a final version of info `property` could be integrated into the download process, to make the user aware of the underlying information and citation.

### DIFF
--- a/src/pymovements/dataset/dataset.py
+++ b/src/pymovements/dataset/dataset.py
@@ -1058,6 +1058,14 @@ class Dataset:
         return self
 
     @property
+    def info(self) -> None:
+        """The information about the dataset.
+
+        Print dataset information and citation key.
+        """
+        print(self.definition.info)
+
+    @property
     def path(self) -> Path:
         """The path to the dataset directory.
 
--- a/src/pymovements/dataset/dataset_definition.py
+++ b/src/pymovements/dataset/dataset_definition.py
@@ -47,6 +47,9 @@ class DatasetDefinition:
     ----------
     name: str
         The name of the dataset. (default: '.')
+    info: str
+        Information about the dataset including but not limited to original citation,
+        general information. (default: '.')
     has_files: dict[str, bool]
         Indicate whether the dataset contains 'gaze', 'precomputed_events', and
         'precomputed_reading_measures'.
@@ -142,6 +145,9 @@ class DatasetDefinition:
 
     # pylint: disable=too-many-instance-attributes
     name: str = '.'
+
+    info: str = ''
+
     has_files: dict[str, bool] = field(default_factory=dict)
 
     mirrors: dict[str, list[str]] | dict[str, tuple[str, ...]] = field(default_factory=dict)
--- a/src/pymovements/datasets/bsc.py
+++ b/src/pymovements/datasets/bsc.py
@@ -44,6 +44,10 @@ class BSC(DatasetDefinition):
     name: str
         The name of the dataset.
 
+    info: str
+        Information about the dataset including but not limited to original citation,
+        general information.
+
     has_files: dict[str, bool]
         Indicate whether the dataset contains 'gaze', 'precomputed_events', and
         'precomputed_reading_measures'.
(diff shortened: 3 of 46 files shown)


src/pymovements/dataset/dataset.py
```diff
@@ -1057,6 +1057,14 @@ def extract(
         )
         return self
 
+    @property
+    def info(self) -> None:
+        """The information about the dataset.
+
+        Print dataset information and citation key.
```
dkrako, line 1064: the property should return a string instead of printing it. A user can easily call `print(dataset.info)` if necessary

dkrako:
Great! Let's discuss this PR tomorrow.

I guess some of the formatting should be edited into a more human-readable format.

Currently the publications are cited in two non-human-readable formats: first as the sphinx-citation and then a bibtex citation.

Furthermore, I would like to split the `info` field into `info` and `citation`, as I think printing the whole info on downloading would be just too much. Instead, a much shorter download disclaimer could be something like:

```
You are downloading the BSC dataset. Please be aware that pymovements does not
If you intend to use the dataset in your research, please cite the publication as:

Jinger Pan, Ming Yan, Eike M. Richter, Hua Shu, and Reinhold Kliegl. The Beijing Sentence Corpus: a Chinese sentence corpus with eye movement data and predictability norms. Behavior Research Methods, 2022.
```

This way we can simply store the main part of the disclaimer somewhere else and just fill in the name and the citation of the dataset. I would be in favor of a human-readable citation format instead of bibtex, because something like this is just hell to parse visually:

```
    @inproceedings{CopCoL1Hollenstein,
      title = "The {C}openhagen Corpus of Eye Tracking Recordings from
      Natural Reading of {D}anish Texts",
      author = {Hollenstein, Nora  and
        Barrett, Maria  and
        Bj{\\"o}rnsd{\\'o}ttir, Marina},
      editor = "Calzolari, Nicoletta  and
        B{\\'e}chet, Fr{\\'e}d{\\'e}ric  and
        Blache, Philippe  and
        Choukri, Khalid  and
        Cieri, Christopher  and
        Declerck, Thierry  and
        Goggi, Sara  and
        Isahara, Hitoshi  and
        Maegaard, Bente  and
        Mariani, Joseph  and
        Mazo, H{\\'e}l{\\`e}ne  and
        Odijk, Jan  and
        Piperidis, Stelios",
      booktitle = "Proceedings of the Thirteenth Language Resources and Evaluation Conference",
      month = jun,
      year = "2022",
      address = "Marseille, France",
      publisher = "European Language Resources Association",
      url = "https://aclanthology.org/2022.lrec-1.182",
      pages = "1712--1720",
    }
```

Moreover, not everyone uses bibtex and we don't want to force our citation preferences on users. If you want to keep the bibtex in someway, then please add a `bibtex` field. We will then have to make sure that the `bibliography.bib` and the definition bibtex are consistent, but this could be handled in a followup, where we autopopulate the `bibliography.bib` with the `bibtex` field of each dataset definition (which would be great as everything would be specified neatly in each definition file). (maybe we could even autogenerate the `citation` field from the `bibtex` field in the future)

Of course we could additionally link to our documentation page in the disclaimer, but I don't know if this is necessary.

Planning further ahead, the `info` field could then be used to autogenerate dataset pages and should include the sphinx citation.

dkrako:
Maybe instead of `info` it would be better to call the field `description`, as that name indicates a text description of the dataset, while `info` is more ambiguous about its content (as all data is information).

We already have the `fileinfo` field in `Dataset` and `info` is unrelated so I'm afraid the naming could be even more confusing.

I like `info` as it's nice and short, but the ambiguities weigh more I guess.

src/pymovements/dataset/dataset_definition.py
```diff
@@ -47,6 +47,9 @@ class DatasetDefinition:
     ----------
     name: str
         The name of the dataset. (default: '.')
+    info: str
+        Information about the dataset including but not limited to original citation,
+        general information. (default: '.')
```
dkrako, line 52: the default is an empty string isn't it?

PR #1059, opened 2025-03-26
chore: remove mirrors from datasets with no real mirrors

dkrako:
The `mirror` field is removed from all dataset definitions that have no actual mirrors (only one mirror specified).

Instead, the resources now provide the full URLs.


dkrako:
I removed the breaking-change label, as I wouldn't consider this a real breaking change anymore.

Technically it is, but I just cannot imagine any workflow where someone uses this attribute directly.

But feel free to comment or re-add the breaking-change label.

SiQube (approved):
nice red PR, love it! keep up the good refactoring, we're on the right track to natural usability

PR #1058, opened 2025-03-26
feat: make `mirrors` optional for downloading datasets

dkrako:
## Description

This PR makes `mirrors` optional for downloading datasets.

Previously, every `DatasetDefinition` had to specify a mirror for resources, even if there's just one mirror specified.

Now it is possible to specify the full url for each resource, omitting the mirror.

## Implemented changes

- [x] support downloading resources without mirrors
- [x] warn instead of print on failed mirror
- [x] refactored `dataset_download.py`

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

- [x] extend dataset definition fixture with `NoMirrors` versions
- [x] add `NoMirrors` test parametrizations for all existing tests
- [x] refactored `dataset_download_test.py`
- [x] remove all `DatasetDefinition` implementatios from `dataset_download_test.py` instead instantiate `DatasetDefinition`
- [x] add test for `@register_dataset` as no tests were using it anymore
- [x] add test for passing `DatasetDefinition` class to `Dataset()` as no tests were using it anymore
- [x] fix annoying tests that wrote files into `tests/files` instead of using `tmp_path`

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

PR #1062, opened 2025-03-26
chore: consistent pylint for pre-commit and tox

### LINKED ISSUE #1061: pylint more restrictive in pre-commit than in tox
## Current Behavior
Linting the `tests` directory is much more restrictive in the local pre-commit hook than it is in the tox environment  (which is used by the github workflow).

## Expected Behavior
The github workflow should be as restrictive as the pre-commit hook.

## Context

Here is the pre-commit definition:

https://github.com/aeye-lab/pymovements/blob/6245bbfceacd018a6f880a4b8f0f3f44ba2acacd/.pre-commit-config.yaml#L138-L152

and here's the one in tox:

https://github.com/aeye-lab/pymovements/blob/6245bbfceacd018a6f880a4b8f0f3f44ba2acacd/tox.ini#L88-L97

The pre-commit definition is actually less restrictive, because it allows for duplicate code. They both use the pylintrc config. How can this be?

## Steps to Reproduce

1. Add a comment line that's longer than 120 characters to one of the test files.

2. Add file change to git via `git add tests`
 
3. run pylint using pre-commit
```
$ pre-commit run pylint
pylint...............................................(no files to check)Skipped
pylint-tests-directory...................................................Failed
- hook id: pylint
- exit code: 16

************* Module dataset_download_test
tests/unit/dataset/dataset_download_test.py:1375: [C0301(line-too-long), ] Line too long (133/100)

-------------------------------------------------------------------
Your code has been rated at 9.96/10 (previous run: 10.00/10, -0.04)
```

4. run pylint using tox

```
$ tox -e pylint
.pkg: _optional_hooks> python /usr/lib/python3.13/site-packages/pyproject_api/_backend.py True setuptools.build_meta
.pkg: get_requires_for_build_sdist> python /usr/lib/python3.13/site-packages/pyproject_api/_backend.py True setuptools.build_meta
.pkg: build_sdist> python /usr/lib/python3.13/site-packages/pyproject_api/_backend.py True setuptools.build_meta
pylint: install_package> python -I -m pip install --force-reinstall --no-deps /home/krakowczyk/workspace/pymovements/.tox/.tmp/package/207/pymovements-0.21.0.post26+git.750b78e0.dirty.tar.gz
pylint: commands[0]> pylint --rcfile=pylintrc --output-format=parseable --ignore-paths=/home/krakowczyk/workspace/pymovements/src/pymovements/_version.py /home/krakowczyk/workspace/pymovements/src/pymovements

--------------------------------------------------------------------
Your code has been rated at 10.00/10 (previous run: 10.00/10, +0.00)

pylint: commands[1]> pylint --rcfile=pylintrc --output-format=parseable --disable=missing-class-docstring,missing-function-docstring,protected-access /home/krakowczyk/workspace/pymovements/tests
  pylint: OK (36.01=setup[5.26]+cmd[30.03,0.73] seconds)
  congratulations :) (36.12 seconds)
```

dkrako:
## Description

Fixes #1061

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -94,7 +94,7 @@ commands =
     pylint --rcfile=pylintrc --output-format=parseable --ignore-paths={toxinidir}/src/pymovements/_version.py \
     {toxinidir}/src/pymovements
     pylint --rcfile=pylintrc --output-format=parseable \
-    --disable=missing-class-docstring,missing-function-docstring,protected-access {toxinidir}/tests
+    --disable=duplicate-code,missing-class-docstring,missing-function-docstring,protected-access {toxinidir}/tests/*
 
 
 [flake8]


dkrako:
This will be a chore for sure. Some of the errors are fixed in #1058

tests/unit/dataset/dataset_test.py
```diff
@@ -39,6 +39,12 @@
 from pymovements.exceptions import InvalidProperty
 
 
+# pylint: disable=too-many-lines
+
+class _UNSET:
+    ...
+
```
SiQube, line 46: why? we could use `from dataclasses import MISSING`

tox.ini
```diff
@@ -94,7 +94,7 @@ commands =
     pylint --rcfile=pylintrc --output-format=parseable --ignore-paths={toxinidir}/src/pymovements/_version.py \
     {toxinidir}/src/pymovements
     pylint --rcfile=pylintrc --output-format=parseable \
-    --disable=missing-class-docstring,missing-function-docstring,protected-access {toxinidir}/tests
+    --disable=duplicate-code,missing-class-docstring,missing-function-docstring,protected-access {toxinidir}/tests/*
```
SiQube, line 97: why?

SiQube (commented):
seems like we're exlucding duplicate-code, why?

regarding `dataset_test.py`, we can probably be smarter with fixtures or move it smth along the lines of `testing_utilities/dataset_fixtures.py`.

if we use this option, `testing_utilities` should be tested as well

IMO having `testing_utilities` or `testing` would be good, there is some logic in our tests that is not tested.

tox.ini
dkrako: We are already disabling it in the otherwise more restrictive pre-commit config:

https://github.com/aeye-lab/pymovements/blob/deae2fa75b97761ee3024b223fee6f82f3db633b/.pre-commit-config.yaml#L147

I don't think that duplicate test code is much of an issue, but I can change this accordingly if you like.

tests/unit/dataset/dataset_test.py
dkrako: Although it would be nice, it is not intended to be used this way:

> No code should directly use the [MISSING](https://docs.python.org/3/library/dataclasses.html#dataclasses.MISSING) value.

https://docs.python.org/3/library/dataclasses.html#dataclasses.field

Relevant stackoverflow post: https://stackoverflow.com/a/69520963

dkrako:
> regarding `dataset_test.py`, we can probably be smarter with fixtures or move it smth along the lines of `testing_utilities/dataset_fixtures.py`.

This will be adressed in #1001 which still has a long dependency chain until development can really start.

tox.ini
SiQube: 🤷 no strong opinion, but I do think we have way too much logic in our tests

SiQube (approved):

tox.ini
dkrako: you're right, I removed the disable.

There's a single test file where I added an ignore:

https://github.com/aeye-lab/pymovements/blob/ec3f99235661e251bf32b02ec3d42c9c3408208b/tests/unit/plotting/traceplot_test.py#L34

It is for a gaze fixture that is copy pasted from `tsplot_test.py`.

I will get rid of that fixture in #1003

PR #1064, opened 2025-03-26
feat: add dynamic `has_resources` property to `DatasetDefinition`

dkrako:
## Description

This PR adds `DatasetDefinition.has_resources` as a getter property that is inferred dynamically on each call.

The `has_resources` property functions both as `bool`:

```python
>>> definition = pm.DatasetLibrary.get('ToyDataset')
>>> definition.has_resources
True
```

and as an indexable object:
```python
>>> definition.has_resources['gaze']
True
```

This is achieved by a helper class `_HasResourcesIndexer` that is forwarded by the getter property.

The contribution has two shortcomings that should be improved on in follow-up PRs:
- Resources have to be reset on each call of `has_resources`. This could be resolved by implementing a `DatasetResources` class that would be the type of `DatasetDefintion.resources`. `DatasetDefintion.has_resources` would then simply call `DatasetDefintion.resources.has_resources`.
- It doesn't fully replace `DatasetDefinition.has_files`, as its usage in [`dataset_files.py`](https://github.com/aeye-lab/pymovements/blob/9d4bbf348d2f420147b8eee0806696c8b81d8afb/src/pymovements/dataset/dataset_files.py#L67-L122) has different semantics: it does not refer to downloadable resources, but to local resources (which could have been downloaded before). Moving resource-dependent attributes to each resource would make a full replacement possible.

Both solutions to the shortcomings are addressed in #1120.

## Implemented changes

- [x] implement `_HasResourcesIndexer` to make `has_resources` indexable
- [x] implement `has_resources()` property and forward `_HasResourceIndexer`
- [x] change `has_files` to `has_resources` in `public_dataset_processing_test.py`

## Type of change

- New feature (non-breaking change which adds functionality)
- This change is or requires a documentation update

## How Has This Been Tested?

- [x] tested various cases for boolean context
- [x] tested various cases for indexable context

#### related issues:

- #1120 

#### requires:
- #1088 

 
## Checklist:


src/pymovements/dataset/dataset_definition.py
```diff
@@ -254,3 +321,41 @@ def to_yaml(self, path: str | Path, exclude_private: bool = True) -> None:
...
+        >>> definition.has_resources['gaze']
+        True
+
+        In this definition there are gaze resources defined, but no precomputed events.
+        >>> definition.has_resources['precomputed_events']
+        False
+        """
+        # Resources may have changed, so update indexer before returning.
+        # A better way to update the resources would be through a resources setter property.
+        self._has_resources.set_resources(self.resources)
```
dkrako, line 360: Unfortunately we have do set the resources on each call, as we cannot guarantee that `self.resources` has not changed.

This could be improved by having a `DatasetResources` class with a `has_resources` property and then forwarding the call to `DatasetDefinition.resources.has_resources` in the property here instead.
related to #1120.

src/pymovements/dataset/dataset_definition.py
```diff
@@ -39,6 +40,67 @@
 yaml.add_multi_constructor('!', type_constructor, Loader=yaml.SafeLoader)
 
 
+# sphinx automatically substitutes _Resources with the Union using the | operator.
+_Resources = Union[dict[str, list[dict[str, str]]], dict[str, tuple[dict[str, str], ...]]]
+
+
+class _HasResourcesIndexer:
```
dkrako, line 47: this helper class is necessary to allow something like:

```
definition.has_resources['gaze']
```

The `['gaze']` index is equivalent to `__getitem__('gaze')`, so the `DatasetDefinition.has_resources` property has to return an object that implements this method.

SiQube: I think having this helper function  within `dataset_definition` makes it even more intimidating to add new datasets for outside users. maybe it could go to `dataset/_utils`?

src/pymovements/dataset/dataset_definition.py
```diff
@@ -39,6 +40,67 @@
 yaml.add_multi_constructor('!', type_constructor, Loader=yaml.SafeLoader)
 
 
+# sphinx automatically substitutes _Resources with the Union using the | operator.
+_Resources = Union[dict[str, list[dict[str, str]]], dict[str, tuple[dict[str, str], ...]]]
```
SiQube, line 44: why is this not possible?

`dict[str, list[dict[str, str]]] | dict[str, tuple[dict[str, str], ...]]`

SiQube (commented):
couple comments/questions.

src/pymovements/dataset/dataset_definition.py
dkrako: moved to `dataset/_utils/_resources.py`

src/pymovements/dataset/dataset_definition.py
SiQube: this is just a `TypeAlias`, right?

this works for tests and pre-commit:

`_Resources: TypeAlias = dict[str, list[dict[str, str]]] | dict[str, tuple[dict[str, str], ...]]`

dkrako: Now I remember why it didn't work out: `|` is not a valid operator in python 3.9. I'm really looking forward to drop py39 support.

dkrako: @SiQube unfortunately this was not possible (see the workflows of my last two commits), but will be possible for python310+.

Can you approve if there aren't any more change requests?

dkrako: @SiQube just a small reminder that this PR was already reviewed and needs a final approve.

dkrako: @SiQube small reminder that this PR was already reviewed and just needs your final approve.

SiQube:
ah I was confused about | not changed but I guess 3.9 ****

SiQube (approved):

PR #1069, opened 2025-03-29
dataset: add trial_columns to ToyDatasetEyeLink

dkrako:
## Description

the dataset was missing the trial columns. I noticed it in #1004 and updated and tested it there, but probably it's better to have a separate PR for the dataset definition change.

## Implemented changes

- [x] added `trial_columns` to `ToyDatasetEyeLink` python class
- [x] added `trial_columns` to `ToyDatasetEyeLink` yaml file

## How Has This Been Tested?

- [x] Tests are written in #1004 with new functionality.
        I wouldn't like to rewrite the tests with the old-style explicit argument passing for this PR here.

## Checklist:

### DIFF
--- a/src/pymovements/datasets/toy_dataset_eyelink.py
+++ b/src/pymovements/datasets/toy_dataset_eyelink.py
@@ -78,7 +78,7 @@ class ToyDatasetEyeLink(DatasetDefinition):
             The name of the trial columns in the input data frame. If the list is empty or None,
             the input data frame is assumed to contain only one trial. If the list is not empty,
             the input data frame is assumed to contain multiple trials and the transformation
-            methods will be applied to each trial separately. (default: None)
+            methods will be applied to each trial separately.
 
     time_column: str
         The name of the timestamp column in the input data frame. This column will be renamed to
--- a/src/pymovements/datasets/toy_dataset_eyelink.py
+++ b/src/pymovements/datasets/toy_dataset_eyelink.py
@@ -74,6 +74,12 @@ class ToyDatasetEyeLink(DatasetDefinition):
         If named groups are present in the `filename_format`, this makes it possible to cast
         specific named groups to a particular datatype.
 
+    trial_columns: list[str]
+            The name of the trial columns in the input data frame. If the list is empty or None,
+            the input data frame is assumed to contain only one trial. If the list is not empty,
+            the input data frame is assumed to contain multiple trials and the transformation
+            methods will be applied to each trial separately. (default: None)
+
     time_column: str
         The name of the timestamp column in the input data frame. This column will be renamed to
         ``time``.
@@ -180,6 +186,10 @@ class ToyDatasetEyeLink(DatasetDefinition):
                 'session_id': int,
             },
         },
+    )
+
+    trial_columns: list[str] = field(
+        default_factory=lambda: ['task', 'trial_id', 'screen_id', 'point_id'],
     )
 
     time_column: str = 'time'
--- a/src/pymovements/datasets/toy_dataset_eyelink.py
+++ b/src/pymovements/datasets/toy_dataset_eyelink.py
@@ -189,7 +189,7 @@ class ToyDatasetEyeLink(DatasetDefinition):
     )
 
     trial_columns: list[str] = field(
-        default_factory=lambda: ['task', 'trial_id', 'screen_id', 'point_id'],
+        default_factory=lambda: ['task', 'trial_id'],
     )
 
     time_column: str = 'time'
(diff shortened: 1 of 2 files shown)


src/pymovements/datasets/toy_dataset_eyelink.py
```diff
@@ -74,6 +74,12 @@ class ToyDatasetEyeLink(DatasetDefinition):
         If named groups are present in the `filename_format`, this makes it possible to cast
         specific named groups to a particular datatype.
 
+    trial_columns: list[str]
+            The name of the trial columns in the input data frame. If the list is empty or None,
+            the input data frame is assumed to contain only one trial. If the list is not empty,
+            the input data frame is assumed to contain multiple trials and the transformation
+            methods will be applied to each trial separately. (default: None)
```
SiQube, line 81: `(default: None)` doesn't make sense here since we specify it to be a `list[str]`

src/pymovements/datasets/toy_dataset_eyelink.py
```diff
@@ -182,6 +188,10 @@ class ToyDatasetEyeLink(DatasetDefinition):
         },
     )
 
+    trial_columns: list[str] = field(
+        default_factory=lambda: ['task', 'trial_id', 'screen_id', 'point_id'],
```
SiQube, line 192: iirc `point_id` is not a trial but part of of judo sequence? @prassepaul

SiQube (commented):
couple of comments, additionally pinged @prassepaul whether `point_id` is a trial column.

src/pymovements/datasets/toy_dataset_eyelink.py
dkrako: good catch. probably both screen_id and point_id are not actual trial columns. just the task and the trial id

src/pymovements/datasets/toy_dataset_eyelink.py
dkrako: you're right, was a copy-paste error

src/pymovements/datasets/toy_dataset_eyelink.py
SiQube: unsure -- `screen_id` might be from the reading trials, but probably is covered already with `trial_id`. @jakobchwastek knows best

dkrako: The reading task consists of a just a single text (the lion and the hare as far as I remember).
The multiple screens of the text should be assigned to the same trial.

To further clarify, `task` is not given in the file itself, but gained from the patterns passed to `from_asc()`:

https://github.com/aeye-lab/pymovements/blob/913c372f83d1d19eada650f28fd2ec52d741f0ad/src/pymovements/datasets/toy_dataset_eyelink.py#L193-L211

src/pymovements/datasets/toy_dataset_eyelink.py
dkrako: I removed mentioning the default at the end. all dataset definition classes seem to conveniently leave the defaults out. so maybe it wasn't a copy-paste error after all.

src/pymovements/datasets/toy_dataset_eyelink.py
dkrako: I removed `point_id` and `screen_id` from the trials

src/pymovements/datasets/toy_dataset_eyelink.py
dkrako: I changed the typehint of the field to `trial_columns: list[str] | None` to allow overwriting these without having mypy complain in my tests in #1004 

I wrote an issue for it to discuss this further #1071

src/pymovements/datasets/toy_dataset_eyelink.py
SiQube: many researchers treat one screen as one trial but should then probably be easy for them to split across screens if they want to. e.g. sb-sat did it in their paper, and hence we adopted it for our papers and the [datasets](https://github.com/aeye-lab/pymovements/blob/d5c4cafd72dfe02cd7feda14b7a7f213b53bf715/src/pymovements/datasets/sb_sat.py#L197-L202) but it can be up to the researcher. I think have finer-grained trials is generally better but it does not really matter for the toy dataset

SiQube (approved):

src/pymovements/datasets/toy_dataset_eyelink.py
dkrako: yes, i think we would make sure that we define and document these things clearly and than first take care of being consistent within pymovements. We can then further support different workflows regarding special needs

Issue #921, opened 2025-01-08
rename `GazeDataFrame` to `Gaze` and `EventDataFrame` to `Events`


dkrako:
## Description of the problem
The class names are much too long and can be shortened drastically

## Description of a solution
rename `GazeDataFrame` to `Gaze` and `EventDataFrame` to `Events` 

to avoid a breaking change without further notice, it would be nice to deprecate the old classes.

## Minimum acceptance criteria
- [ ] rename classes
- [ ] deprecate old classes to avoid breaking change

SiQube:
additionally, we have within `gaze`, [`gaze_dataframe.py`](https://github.com/aeye-lab/pymovements/blob/main/src/pymovements/gaze/gaze_dataframe.py), but for `events` and `reading_measures` just `frame.py`. might be worth to refactor as well

dkrako:
Sure, we can rename `gaze_dataframe.py` to `frame.py` or maybe even just `gaze.py` as it's not just a frame anymore but has some metadata included.

dkrako:
Here are some ways on how to handle class name deprecation:
https://stackoverflow.com/questions/9008444/how-to-warn-about-class-name-deprecation

I'm in favor of something like this:

```python
from warnings import warn

class ClassDeprecationHelper(object):
    def __init__(self, new_class, old_name):
        self.new_class = new_class
        self.old_name = old_name

    def _warn(self):
        warn(
            f'{self.old_name} is deprecated in favor of {self.new_class.__name__}.'
            ' Please change your code accordingly!',
        )

    def __call__(self, *args, **kwargs):
        self._warn()
        return self.new_class(*args, **kwargs)

    def __getattr__(self, attr):
        self._warn()
        return getattr(self.new_class, attr)

GazeDataFrame = ClassDeprecationHelper(Gaze, "GazeDataFrame")
EventDataFrame = ClassDeprecationHelper(Event, "EventDataFrame")

```

dkrako:
After further thinking I would like to get rid of `EventsDataFrame` completely. I don't see the benefit anymore when `Gaze` already includes the events and syncing the `trial_columns` is error-prone.

SiQube:
I think so too -- then it would also fit with #942 ?

PR #1077, opened 2025-04-03
chore: remove unnecessary ignore of cyclic-import

SiQube:
towards #559


dkrako (approved):
already included in #1004 but I give you that :)

unfortunately this doesn't help much for #559 as the circular dependency is between `events` and `gaze` (that's why we need to import `pm.EventDataFrame` in `gaze_dataframe.py` via `import pymovements as pm`).
This will be resolved when we get rid of `EventDataFrame` und subsume it under `Gaze` as proposed in #921

PR #1084, opened 2025-04-09
feat: use locale encoding by default in from_asc()

dkrako:
## Description

This PR sets the default value of `gaze.from_asc()` to `None` (was: `ascii`) .

The main reason is that `None` defaults reduce headaches when implementing #1004 

The [`open()`](https://docs.python.org/3/library/functions.html#open) function has the following documentation on the `encoding` argument:

> In text mode, if encoding is not specified the encoding used is platform-dependent: [locale.getencoding()](https://docs.python.org/3/library/locale.html#locale.getencoding) is called to get the current locale encoding.

So I guess it would also be more in line with the builtins to have `None` as the default.

This is not a breaking change, as `ascii` is a small charset that is fully included in both `utf8` (locale on mac and linux) and `latin1` (locale on windows).

For eyelink-ascii datasets the encoding should now always be specified to make sure it is independent of platform configurations.

## Implemented changes

- [x] change `encoding` default in `gaze/io.py::from_asc()`
- [x] change `encoding` default in `gaze/_utils/parsing.py::parse_eyelink()`
- [x] add `encoding` to `ToyDatasetEyeLink`

## How Has This Been Tested?

- [x] test encoding results in expected umlauts in metadata pattern
- [x] test exception if ascii encoding but umlauts in file

## Checklist:


dkrako:
@saeub I also added you as a reviewer as you added the encoding argument.
you might possibly see some shortcomings of this PR that I didn't think of

saeub:
> To avoid this, it should be sufficient to specify the encoding in every DatasetDefinition that uses eyelink asc files, right?

Yes. I think it would be enough to do it for the datasets where non-ASCII characters occur, but it might be easier just to do it for all of them.

dkrako:
If I'm correct, then currently we only have `ToyDatasetEyeLink`, which has only ascii characters.
We would just need to make sure for future datasets that encoding is specified to guarantee locale platform independence.

saeub:
RaCCooNS (https://github.com/aeye-lab/pymovements/pull/961) will be the first one with non-ASCII characters

PR #1087, opened 2025-04-11
ci: create workflow for benchmarking import times

dkrako:
## Description

This creates a first benchmarking workflow for pymovements.
The most trivial thing to measure is the import time. Currently it is about 3 seconds which should be lowerered.

## Implemented changes

- [x] add pytest-benchmark to test dependencies
- [x] implement tox benchmark workflow
- [x] implement github worker
- [x] implement benchmarking test


dkrako:
An example of a summary is here: https://github.com/aeye-lab/pymovements/actions/runs/14413954118?pr=1087

It will always compare against the latest benchmark run on main.

It is missing from the last workflow runs in this PR as no benchmark was run on main yet.

tests/benchmark/import_test.py
```diff
@@ -0,0 +1,60 @@
...
+
+def import_pymovements():
+    import pymovements  # noqa: F401 # pylint: disable=import-outside-toplevel,unused-import
+
+
+def test_import_pymovements(benchmark):
+    benchmark.pedantic(
+        import_pymovements,
+        setup=setup_import_pymovements,
+        iterations=1, rounds=1,  # only a single round due to leakage to subsequent rounds
```
dkrako, line 47: this benchmark is only valid for a single round. subsequent rounds are much shorter, so there's some leakage between rounds.

we could probably remove this test and only run the subprocess kind, as the values are very similar.

dkrako:
@SiQube just a small reminder that this PR is ready for review

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,52 @@
...
+    steps:
+      - name: Checkout pymovements repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Setup python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.13"
```
SiQube, line 21: while I think people *should* use the newest version of python, it is very unrealisitic. even bigger libraries have not adopted to 3.13 yet, so maybe benchmarking on <3.13, e.g. 3.10 or so would be more practical.

tests/benchmark/import_test.py
SiQube: for me this is not the case:

when appyling this diff
```diff
(venv) siqube@c-cube:~/lab/aeye-lab/pymovements
$ git diff
diff --git a/tests/benchmark/import_test.py b/tests/benchmark/import_test.py
index 1b6a251..85288d0 100644
--- a/tests/benchmark/import_test.py
+++ b/tests/benchmark/import_test.py
@@ -56,5 +56,5 @@ def import_pymovements_subprocess():
 def test_import_pymovements_subprocess(benchmark):
     benchmark.pedantic(
         import_pymovements_subprocess,
-        iterations=1, rounds=10,
+        iterations=1, rounds=1,
     )

```

I get

```console
(venv) siqube@c-cube:~/lab/aeye-lab/pymovements
$ tox -e benchmark
.pkg: _optional_hooks> python /home/siqube/lab/aeye-lab/pymovements/venv/lib/python3.12/site-packages/pyproject_api/_backend.py True setuptools.build_meta
.pkg: get_requires_for_build_sdist> python /home/siqube/lab/aeye-lab/pymovements/venv/lib/python3.12/site-packages/pyproject_api/_backend.py True setuptools.build_meta
.pkg: build_sdist> python /home/siqube/lab/aeye-lab/pymovements/venv/lib/python3.12/site-packages/pyproject_api/_backend.py True setuptools.build_meta
benchmark: install_package> python -I -m pip install --force-reinstall --no-deps /home/siqube/lab/aeye-lab/pymovements/.tox/.tmp/package/8/pymovements-0.21.0.post46+git.e60c44d7.dirty.tar.gz
benchmark: commands[0]> pytest --ignore=src --ignore=tests/functional --ignore=tests/integration --ignore=tests/unit --benchmark-json /home/siqube/lab/aeye-lab/pymovements/benchmark-data.json .
================================================================================================================= test session starts ==================================================================================================================
platform linux -- Python 3.12.3, pytest-8.3.5, pluggy-1.5.0
cachedir: .tox/benchmark/.pytest_cache
benchmark: 5.1.0 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /home/siqube/lab/aeye-lab/pymovements
configfile: pyproject.toml
plugins: cov-6.1.1, benchmark-5.1.0
collected 2 items                                                                                                                                                                                                                                      

tests/benchmark/import_test.py ..                                                                                                                                                                                                                [100%]
Wrote benchmark data in: <_io.BufferedWriter name='/home/siqube/lab/aeye-lab/pymovements/benchmark-data.json'>



----------------------------------------------------------------------------------------------- benchmark: 2 tests ----------------------------------------------------------------------------------------------
Name (time in ms)                             Min                   Max                  Mean            StdDev                Median               IQR            Outliers     OPS            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_import_pymovements                  154.3648 (1.0)        154.3648 (1.0)        154.3648 (1.0)      0.0000 (1.0)        154.3648 (1.0)      0.0000 (1.0)           0;0  6.4782 (1.0)           1           1
test_import_pymovements_subprocess     1,268.2934 (8.22)     1,268.2934 (8.22)     1,268.2934 (8.22)     0.0000 (1.0)      1,268.2934 (8.22)     0.0000 (1.0)           0;0  0.7885 (0.12)          1           1
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

```

similarly when using 10 runs:

```diff
```

the numbers seem to be kinda consistent

```console
(venv) siqube@c-cube:~/lab/aeye-lab/pymovements
$ git diff
diff --git a/tests/benchmark/import_test.py b/tests/benchmark/import_test.py
index 1b6a251..bc19808 100644
--- a/tests/benchmark/import_test.py
+++ b/tests/benchmark/import_test.py
@@ -44,7 +44,7 @@ def test_import_pymovements(benchmark):
     benchmark.pedantic(
         import_pymovements,
         setup=setup_import_pymovements,
-        iterations=1, rounds=1,  # only a single round due to leakage to subsequent rounds
+        iterations=1, rounds=10,  # only a single round due to leakage to subsequent rounds
     )
```

SiQube (commented):
that you for looking into this, I think this will be a cool feature but currently I think this is not yet ready to be merged.

1. see my comments about runtimes
2. I think we should find import leakage and figure out what is going on there instead of merging it 

but your vision is clearer than mine, I think we should rather focus on features etc. instead of refactoring and benchmarking

.github/workflows/benchmark.yml
dkrako: This is the definition of a GitHub workflow. Github workflows are internal to github and neither targeted at the user nor at the developer. They are handled purely by the CI.

we should benchmark on the newest python versions with the newest optimizations.
I don't care much about performance on old python systems.

dkrako:
>     2. I think we should find import leakage and figure out what is going on there instead of merging it
> > the numbers seem to be kinda consistent

You didn't add the results for 10 runs. Your first run already shows that there's import leakage already for the first run.

I removed the leaking brenchmark. I don't see any reason why we should try to hack around `pytest-benchmark` as it's clearly not designed to measure import times. see: https://github.com/ionelmc/pytest-benchmark/issues/177

Importing pymovements in a subprocess is the best way to avoid import leakage.

dkrako:
To further clarify: the major contribution of this PR is implementing a benchmarking workflow for github workers.
the benchmarking of import times is minor in that regard.

Nevertheless, import times are very long currently. You're machine must be extremely fast. Mine usually takes 3 to 4 seconds.

The import leakage that you get already in your first run may misguide you: the `test_import_pymovements_subprocess` is much closer to the real import time than your results on `test_import_pymovements`.

You can try this out yourself using 

```python
>>> import timeit
>>> timeit.timeit('import pymovements')
Backend tkagg is interactive backend. Turning interactive mode on.
3.9113128779572435
```

SiQube:
```python
>>> import timeit
>>> timeit.timeit('import pymovements')
1.0151693140001044
```

SiQube (approved):

dkrako:
Thanks!

As an additional note: I won't set this workflow up as a hard merge requirement for now.

We will get notified in case commits or PRs deteriorate performance and can then act accordingly.

Let's see how this works out first.

PR #1081, opened 2025-04-08
depr: deprecate `utils.plotting`

dkrako:
## :warning: Deprecation

`utils.plotting` is deprecated. draw_image_stimulus() please use `ImageStimulus.show()` instead.
`utils.plotting` will be removed in pymovements v0.27.0

## Description

Deprecate `utils.plotting` in favor of `ImageStimulus.show()`

The following warnings are printed:

```
DeprecationWarning: Call to deprecated function (or staticmethod) draw_image_stimulus(). (Please use ImageStimulus.show() instead. This function will be removed in v0.27.0.) -- Deprecated since version v0.27.0.
```

## Implemented changes

- [x] moved `draw_image_stimulus()` from `utils/plotting.py` to `stimulus/image.py`
- [x] moved remaining logic from `utils/plotting.py` to `plotting/_matplotlib.py`
- [x] update usage in `plotting`
- [x] created deprecated alias functions in `utils/plotting.py`

## How Has This Been Tested?

- [x] moved some tests to `image_plotting_test.py`
- [x] add test that functions are correctly deprecated
- [x] add test that functions are removed in planned version

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

# Context

##### related issues:
- #999 

##### requires:
- #1094

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

PR #1089, opened 2025-04-12
feat: automatically infer if resources need extract

dkrako:
## Description

This PR automatically infers if dataset resources need to be extracted or decompressed.
It removes the need to define `DatasetDefinition.extract`.

`DatasetDefinition.extract` is not removed yet. It will be deprecated in a follow-up PR #1159 

## Implemented changes

- [x] autodetect of resources need extraction by checking filename suffixes
- [x] added `UnknownFileType` exception
- [x] refactored `download_dataset()` and `extract_dataset()`
- [x] improved some error handling

## How Has This Been Tested?

Testing could certainly improved but this would probably need a rewrite of `dataset_download_test.py`.

This will happen in some of the upcoming PRs where I will create a `Resources.extract()` method which will be more sensible to test than the private methods currently tested in `dataset_download_test.py`.

From the outside the behavior stays the same (if `extract`was specified correctly), so the tests are mostly left as they were.

#### related issues:
- #1120

#### requires:
- [x] #1081 
- [x] #1094 

#### required by:
- [ ] #1158
- [ ] #1159 

## Checklist:


src/pymovements/utils/downloads.py
```diff
@@ -85,12 +84,20 @@ def download_and_extract_archive(
         If the downloaded file has no suffix or suffix is not supported, or in case of a
         specified MD5 checksum which doesn't match the checksum of the downloaded file.
     """
-    return _download_and_extract_archive(
+    archive_path = _download_file(
```
dkrako, line 87: implementation moved from `dataset/_utils/_downloads.py`

src/pymovements/dataset/_utils/_downloads.py
```diff
@@ -28,80 +28,10 @@
 from tqdm.auto import tqdm
 
 from pymovements._version import get_versions
-from pymovements.dataset._utils._archives import extract_archive
 
 USER_AGENT: str = f"pymovements/{get_versions()['version']}"
 
 
-def download_and_extract_archive(
```
dkrako, line 36: this function is not needed anymore. the functionality is implemented in `dataset_download.download_dataset()`.

this implementation is moved to the deprecated function `utils.download_and_extract_archive`

Issue #1075, opened 2025-04-01
add disclaimer when downloading a dataset


SiQube:
## Description of the problem

currently when a user uses a public pymovements dataset it is assumed that they will do the proper research within our method to attribute/cite the dataset correctly.

a disclaimer should be integated

## Description of a solution

- [ ] integrate a general disclaimer about dataset attribution if the dataset is used in their project. 
- [ ] add citation key(s) from pymovements bibliography to each dataset
- [ ] add a dataset description property (similar to the docstring)

## Minimum acceptance criteria

- [ ] disclaimer automatically appears when downloading a dataset
```pycon
>>> import pymovements as pm
>>> pm.Dataset('BSC', 'data').download()
You are downloading the BSC dataset. Please be aware that pymovements does not
host or distribute any dataset resources and only provides a convenient interface to
download the public dataset resources that were published by their respective authors.
```
- [ ] property showing general dataset information
```pycon
>>> import pymovements as pm
>>> pm.Dataset('BSC', 'data').description
Some general description of the BSC dataset.
```
- [ ] add citation_keys property
```pycon
>>> import pymovements as pm
>>> pm.Dataset('BSC', 'data').citation_keys
BSC
```
- [ ] add test checking that citation keys are not the empty string for public datasets
- [ ] add citation (key) o disclaimer 
```pycon
>>> import pymovements as pm
>>> pm.Dataset('BSC', 'data').download()
You are downloading the BSC dataset. Please be aware that pymovements does not
host or distribute any dataset resources and only provides a convenient interface to
download the public dataset resources that were published by their respective authors.

You can lookup the citation in our bibliography using ...
```

## Sample Code
to implement the disclaimer add to `dataset.py`

```python
def disclaimer(self) -> str:
    _disclaimer_text = f"""\
You are downloading the {self.name} dataset. Please be aware that pymovements does not
host or distribute any dataset resources and only provides a convenient interface to
download the public dataset resources that were published by their respective authors.
"""
    return _disclaimer_text
```

add to download:
```python
self.disclaimer()
```

add to `dataset_definition.py`

```python
citation_keys: Sequence[str] = ''
```

this can also be a `list` or `tuple` in case one dataset spans multiple citations without changing the type annotation

```python
citation_keys: Sequence[str] = ['citation1', 'citation2', 'citation3']
```

add an `assert` to `tests/unit/datasets/datasets_test.py` checking that `citation_keys` are not the empty string:

```python
[...]

PR #1097, opened 2025-04-16
feat: print disclaimer when downloading a public dataset

SiQube:
## Description

print disclaimer when downloading a public dataset

## Type of change

Remove irrelevant items:
- New functionality
- Documentation update

#### related issues:
- #1057 
- #1075 

#### requires:
- [ ] #1096

### DIFF
--- a/src/pymovements/dataset/dataset_download.py
+++ b/src/pymovements/dataset/dataset_download.py
@@ -20,6 +20,7 @@
 """Provides private functions for downloading and extracting datasets."""
 from __future__ import annotations
 
+import logging
 import shutil
 from pathlib import Path
 from urllib.error import URLError
@@ -29,6 +30,10 @@ from pymovements.dataset._utils._archives import extract_archive
 from pymovements.dataset._utils._downloads import download_file
 from pymovements.dataset.dataset_definition import DatasetDefinition
 from pymovements.dataset.dataset_paths import DatasetPaths
+
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger('__name__')
 
 
 def download_dataset(
@@ -77,6 +82,15 @@ def download_dataset(
     RuntimeError
         If downloading a resource failed for all given mirrors.
     """
+    disclaimer_text = f"""\
+You are downloading the {definition.long_name}. Please be aware that pymovements does not
+host or distribute any dataset resources and only provides a convenient interface to
+download the public dataset resources that were published by their respective authors.
+
+Please cite the referenced publication if you intend to use the dataset in your research.
+"""
+    logger.info(disclaimer_text)
+
     if definition.has_files['gaze']:
         if not definition.mirrors or not definition.mirrors['gaze']:
             mirrors = None
--- a/src/pymovements/dataset/dataset_download.py
+++ b/src/pymovements/dataset/dataset_download.py
@@ -33,7 +33,7 @@ from pymovements.dataset.dataset_paths import DatasetPaths
 
 
 logging.basicConfig(level=logging.INFO)
-logger = logging.getLogger('__name__')
+logger = logging.getLogger(__name__)
 
 
 def download_dataset(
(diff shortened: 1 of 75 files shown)


src/pymovements/dataset/dataset_download.py
```diff
@@ -77,6 +77,13 @@ def download_dataset(
     RuntimeError
         If downloading a resource failed for all given mirrors.
     """
+    disclaimer_text = f"""\
+You are downloading the {definition.long_name}. Please be aware that pymovements does not
+host or distribute any dataset resources and only provides a convenient interface to
+download the public dataset resources that were published by their respective authors.
```
dkrako, line 83: please also add a second paragraph with the request to cite the paper (the particular citation will be added later so we can be generic):

```
```

dkrako (changes requested):
Great step forward! Some minor change requests regarding the text and the print function.

src/pymovements/dataset/dataset_download.py
```diff
@@ -77,6 +77,13 @@ def download_dataset(
...
+"""
+    print(disclaimer_text)
```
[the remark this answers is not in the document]
SiQube: done

src/pymovements/dataset/dataset_download.py
SiQube: done

src/pymovements/dataset/dataset_download.py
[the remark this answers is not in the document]
SiQube: if use like this, since the logging level is debugging, it does not show up in e.g. terminal. is this intended or should I change the level to debug/info?

src/pymovements/dataset/dataset_download.py
```diff
@@ -77,6 +81,15 @@ def download_dataset(
     RuntimeError
         If downloading a resource failed for all given mirrors.
     """
+    disclaimer_text = f"""\
+You are downloading the {definition.long_name}. Please be aware that pymovements does not
```
dkrako, line 85: What do you think about adding the short name in parentheses?

```
You are downloading the {definition.long_name} ({definition.name}).
```

dkrako: On second thought, that won't work well for some datasets: 

```
Code Comprehension dataset (Code Comprehension)
```

Although this still would be nice:

```
Beijing Sentence Corpus (BSC)
```

So let's leave it as it is for now, maybe we'll have some further ideas in the future.

src/pymovements/dataset/dataset_download.py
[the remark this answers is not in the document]
dkrako: I just tested this on the console and this works, but the doctest just doesn't get it.

I also tried adding `logger.addHandler(logging.StreamHandler(sys.stdout))` but still no success.

So let's just use `print()` in this PR. I created #1118 to replace all `print()` statements with logging calls.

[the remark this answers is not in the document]
SiQube: I just removed the line from the doctest. I think logging is more elegant.

src/pymovements/dataset/dataset_download.py
```diff
@@ -31,6 +32,10 @@
 from pymovements.dataset.dataset_paths import DatasetPaths
 
 
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger('__name__')
```
dkrako, line 36: the string `'__name__'` should be replaced with the dunder variable `__name__`, otherwise the info source is not the module but just `__name__`.

(`__name__` will resolve to `pymovements.dataset.dataset_download`)

SiQube: good catch!

dkrako (approved):
Great! Thanks a lot for all your work! You'll just need to resolve the merge conflicts (or just checkout the datasets directory from main, i think it was something like `git checkout main -- src/pymovements/datasets`)

PR #1099, opened 2025-04-17
dataset: Chinese Reading

SiQube:
additionally adds functionality to parse excel files

adds Chinese Reading dataset as well as functionality to functionality to parse excel files in `load_precomputed_events`

## Type of change
 
* [x] New feature (non-breaking change which adds functionality)
* [x] New dataset
 
✨ Enhancements

- add feature to read precomputed event `.xlsx` files 

📀 Datasets
- add Chinese Reading dataset 

## How Has This Been Tested?
 
* [x] passing `tests/unit/datasets/datasets_test.py::test_public_dataset_registered` for `ChineseReading`
* [x] passing adjusted test for loading precomputed event files `tests/unit/dataset/dataset_files_test.py::test_load_precomputed_rm_file_unsupported_file_format`
* [x] load precomputed events for `json` files `tests/unit/dataset/dataset_files_test.py::test_load_precomputed_rm_file_xlsx`


dkrako (changes requested):
Please add a full documentation on the new functionality.

SiQube:
@dkrako I don't get the pylint error -- can you help?

dkrako:
I'm also really confused what the problem is. It seems to be a weird bug in pylint? Strange that this is only a problem for `pl.read_excel()`.

I rewrote the lines, but the formatting isn't optimal this way (having an explicit `expected_df` is better), but it's not critical there.

I am fine with leaving it this way, or maybe we could add a pylint ignore for the line / test function? Both is good enough.

SiQube:
would leave it like this and merge?

src/pymovements/datasets/chinese_reading.yaml
```diff
@@ -0,0 +1,41 @@
+name: ChineseReading
+
+has_files:
+  gaze: false
+  precomputed_events: true
+  precomputed_reading_measures: true
+
+mirrors:
+  precomputed_events:
+      - https://files.de-1.osf.io/v1/resources/94wue/providers/osfstorage/
```
dkrako: please remove `.de-1` from the url as this is the url to a german mirror

src/pymovements/datasets/chinese_reading.yaml
```diff
@@ -0,0 +1,41 @@
...
+  precomputed_reading_measures:
+      - https://files.de-1.osf.io/v1/resources/94wue/providers/osfstorage/
```
dkrako: please remove `.de-1` from the url as this is the url to a german mirror

dkrako: also, specifying the mirror in the `DatasetDefinition` isn't necessary anymore, right?

src/pymovements/datasets/chinese_reading.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+resources:
+  precomputed_events:
+    - resource: ?zip=
```
dkrako: if you're interested just in a single file, this is the actual URL to the `Raw Data.txt`: https://files.osf.io/v1/resources/94wue/providers/osfstorage/6253cb37840dd726e75c831a

tests/files/Sentences.xlsx
dkrako, line 1: this is the exact file that is uploaded by the authors. we mustn't copy dataset files without the authors' explicit permission. please create a new test file from scratch.

src/pymovements/datasets/chinese_reading.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+@dataclass
+class ChineseReading(DatasetDefinition):
+    """ChineseReading dataset :cite:p:`ChineseReading`.
+
+    This dataset includes eye tracking data from more than 300 participants recorded in a single
+    session. Precomputed events and word-level reading measures are reported.
+
+    The participant is instructed to read several sentences.
```
dkrako: `The participants are instructed...` or `Each participant is instructed...`

dkrako (changes requested):
There are minor issues I would like to have changed before merging.

1. please remove the `.de-1` from the URLs
2. please replace the `.xlsx` file with a custom file from scratch. This is critical, as once merged, the content cannot be easily removed.

src/pymovements/datasets/chinese_reading.yaml
SiQube: done

src/pymovements/datasets/chinese_reading.yaml
SiQube: done

src/pymovements/datasets/chinese_reading.yaml
SiQube: done

tests/files/Sentences.xlsx
SiQube: good catch -- rewritten

src/pymovements/datasets/chinese_reading.py
SiQube: done

dkrako (approved):
Awesome! Thank you so much for extending our library of datasets!

Issue #1072, opened 2025-03-30
Integrate tracked eye in ASC files into gaze dataframe


saeub:
## Description of the problem

The tracked eye parsed from ASC files is currently only stored as metadata in the `GazeDataFrame._metadata` dictionary. The `GazeDataFrame`'s pixel column supports right/left/binocular samples in lists (to be refactored into structs, see #453), so the information about the tracked eye could be stored directly in that column.

Note that the tracked eye can change throughout the ASC file (see discussion [here](https://github.com/aeye-lab/pymovements/pull/887#issuecomment-2764516475)).

Related to #453 and #686.

## Description of a solution

The pixel column in the `GazeDataFrame` created by `from_asc()` should contain a data structure along the lines of `{left: [123, 456]}` or `{right: [123, 456]}` for monocular recordings and `{left: [123, 456], right: [123, 456]}` for binocular recordings.

## Minimum acceptance criteria

Information about which eyes were tracked is available in the `GazeDataFrame` obtained through `from_asc()`.

dkrako:
great, thanks for writing the issue! we should probably implement #686 first. I guess #453 could take a while until started

PR #1102, opened 2025-04-17
dataset: add CoLAGaze dataset

SiQube:
there is slight problem in the raw files, will chat with @annabondar3

------

```console
ValueError: Experiment metadata does not match the metadata in the ASC file:
- Right eye tracked: True != False
```

one can reproduce it when changing

```diff
diff --git a/src/pymovements/datasets/colagaze.yaml b/src/pymovements/datasets/colagaze.yaml
index d460bda..73d5b44 100644
--- a/src/pymovements/datasets/colagaze.yaml
+++ b/src/pymovements/datasets/colagaze.yaml
@@ -4,7 +4,7 @@ has_files:
   # need to fix bug:
   #     ValueError: Experiment metadata does not match the metadata in the ASC file:
   #     - Right eye tracked: True != False
-  gaze: false
+  gaze: true
   precomputed_events: true
   precomputed_reading_measures: true
```

and running

```python
import pymovements as pm

dataset = pm.Dataset('CoLAGaze', 'data/CoLAGaze').download().load()
```

maybe @saeub can help me

#### requires:
- #1151


SiQube:
some more information @dkrako @saeub 

in the experiment, @annabondar3 tracked the dominant eye, thus monocular but changing eyes => the tracked eye is not defined in the experiment.

however, after the first ASCII file, the eye tracker is also defined within the dataset definition experiment. the mutability of the DatasetDefinition/Experiment is a curse here.

saeub:
I see the problem. I think there are several possible solutions:
1. Decouple the GazeDataFrame-level experiment definition from the Dataset-level experiment definition, i.e., create a copy of the existing Experiment object [here](https://github.com/aeye-lab/pymovements/blob/e6c074e632f5dfefcfe8717cff7b69adb954b9f9/src/pymovements/gaze/io.py#L498). Personally I don't really like this, because this means that depending on the dataset, there will be several `Experiment` objects and it's not obvious which one contains information at which level.
2. Remove the tracked eye from the `EyeTracker` definition, since this is session-level metadata. After #1072 is implemented, the information about tracked eye should be integrated in the GazeDataFrame directly anyway. Not sure what implications this would have for other parts of the code. I think this would be the cleanest solution, but it would be a breaking change and would take more time to implement.
3. Redefine `EyeTracker.left` and `EyeTracker.right` to mean "has the left/right eye been tracked in *some part* of the dataset". I think this would be the easiest fix, as we would just have to change the logic [here](https://github.com/aeye-lab/pymovements/blob/e6c074e632f5dfefcfe8717cff7b69adb954b9f9/src/pymovements/gaze/io.py#L526-L533), but it probably makes the `left`/`right` attributes less useful, so we might as well go with option 2. directly.

SiQube:
I prefer 2., but @dkrako has the bigger picture in mind

SiQube:
we can merge the current dataset, after approval, and open an issue for the gaze sample data

saeub:
An additional note: The automatic detection of right/left eye only works for ASC files. For non-ASC datasets, the tracked eye is only stored in the `EyeTracker` object, so if we go for option 2, we lose that information for other datasets. But then again, I'm not sure we *should* store it at the dataset-level at all, since it can be different for every participant/session, or even within one session.

src/pymovements/datasets/colagaze.yaml
```diff
@@ -0,0 +1,63 @@
...
+  # need to fix bug:
+  #     ValueError: Experiment metadata does not match the metadata in the ASC file:
+  #     - Right eye tracked: True != False
+  gaze: false
+  precomputed_events: true
+  precomputed_reading_measures: true
+
+mirrors:
+  gaze:
+      - https://files.au-1.osf.io/v1/resources/gj2uk/providers/osfstorage/67e14ce0f392601163f33215/?view_only=a8ac6e0091e64d0a81d5b1fdec9bab6e&zip=
```
dkrako, line 13: so this means the dataset needs to be downloaded three times? how big are the archives?

Just another sign that the current design of `DatasetDefinition` is really bad and needs to be improved.

src/pymovements/datasets/colagaze.py
dkrako, line 1: misses `long_name`

src/pymovements/datasets/colagaze.yaml
```diff
@@ -0,0 +1,63 @@
+name: "CoLAGaze"
+
+has_files:
+  # need to fix bug:
+  #     ValueError: Experiment metadata does not match the metadata in the ASC file:
+  #     - Right eye tracked: True != False
```
dkrako, line 6: please elaborate. doesn't this mean that the experiment data is wrong?

dkrako (changes requested):

dkrako:
> I prefer 2., but @dkrako has the bigger picture in mind

I also prefer the 2.

dkrako:
> An additional note: The automatic detection of right/left eye only works for ASC files. For non-ASC datasets, the tracked eye is only stored in the `EyeTracker` object, so if we go for option 2, we lose that information for other datasets. But then again, I'm not sure we _should_ store it at the dataset-level at all, since it can be different for every participant/session, or even within one session.

I am confused. Why do we lose the information for other datasets? How are these affected?

As long as it's consistent for the dataset, it can stay defined in the `DatasetDefinition`.

As we have eyelink asc files here, we are on the safe side for now.

saeub:
> I am confused. Why do we lose the information for other datasets? How are these affected?

I think what I meant is: If we simply remove the `left`/`right` attributes from the `EyeTracker` object, there is no way to specify the tracked eye in the `DatasetDefinition` anymore. For ASC-based datasets, it can be inferred from the data files, but not for CSV-based datasets. But we can of course add the information to the `DatasetDefinition` in some other way.

src/pymovements/datasets/colagaze.yaml
dkrako: ah, so let's fix it in logic. can you create an issue?

dkrako:
> > I am confused. Why do we lose the information for other datasets? How are these affected?
> 
> I think what I meant is: If we simply remove the `left`/`right` attributes from the `EyeTracker` object, there is no way to specify the tracked eye in the `DatasetDefinition` anymore. For ASC-based datasets, it can be inferred from the data files, but not for CSV-based datasets. But we can of course add the information to the `DatasetDefinition` in some other way.

I'm completely confused. Why should we remove the attributes from the class? They are optional. 

For csv datasets with varying experiment definitions, we should extend `DatasetDefinition` to support defining experiments per file(name pattern).

saeub:
That's how I understood option 2. But it's probably easier to discuss this in a meeting :sweat_smile:

src/pymovements/datasets/colagaze.yaml
saeub: The problem is that it is overwritten multiple times (once per ASC file), and if not every ASC file uses the same eye, this currently leads to an error. I didn't have this scenario in mind when implementing #884.

dkrako: Can you post the trace of the warning/exception?

There was a time when `DatasetLibrary.get()` returned the class, but you already resolved this issue in #914 by returning a copy of an instance. Is this related?

dkrako: Reviewing this error message:

```
ValueError: Experiment metadata does not match the metadata in the ASC file:
- Right eye tracked: True != False
```

`Experiment.eyetracker.right` should not be `False`, it should be `None`.

SiQube:
```console
$ python t.py --dataset CoLAGaze --download
INFO:pymovements.dataset.dataset:        You are downloading the Corpus of Eye Movements for Linguistic Acceptability. Please be aware that pymovements does not

Downloading https://files.au-1.osf.io/v1/resources/gj2uk/providers/osfstorage/67e14ce0f392601163f33215/?view_only=a8ac6e0091e64d0a81d5b1fdec9bab6e&zip= to data/CoLAGaze/downloads/raw_data.zip
raw_data.zip: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 356M/356M [03:47<00:00, 1.64MB/s]
Checking integrity of raw_data.zip
Extracting raw_data.zip to data/CoLAGaze/raw
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 42/42 [00:19<00:00,  2.12it/s]
 17%|████████████████████████████████████████▋                                                                                                                                                                                                           | 7/42 [00:51<04:30,  7.73s/it]/home/reich3/lab/aeye-lab/pymovements/src/pymovements/gaze/_utils/parsing.py:482: UserWarning: Found inconsistent values for 'tracked_eye': ['L', 'R']
  warnings.warn(f"Found inconsistent values for '{key}': {sorted_values}")
 17%|████████████████████████████████████████▋                                                                                                                                                                                                           | 7/42 [00:59<04:55,  8.44s/it]
Traceback (most recent call last):
  File "/home/reich3/lab/aeye-lab/pymovements/t.py", line 27, in <module>
    raise SystemExit(main())
  File "/home/reich3/lab/aeye-lab/pymovements/t.py", line 20, in main
    dataset.load()
  File "/home/reich3/lab/aeye-lab/pymovements/src/pymovements/dataset/dataset.py", line 146, in load
    self.load_gaze_files(
  File "/home/reich3/lab/aeye-lab/pymovements/src/pymovements/dataset/dataset.py", line 225, in load_gaze_files
    self.gaze = dataset_files.load_gaze_files(
  File "/home/reich3/lab/aeye-lab/pymovements/src/pymovements/dataset/dataset_files.py", line 257, in load_gaze_files
    gaze_df = load_gaze_file(
  File "/home/reich3/lab/aeye-lab/pymovements/src/pymovements/dataset/dataset_files.py", line 355, in load_gaze_file
    gaze_df = from_asc(
  File "/home/reich3/lab/aeye-lab/pymovements/src/pymovements/gaze/io.py", line 430, in from_asc
    experiment = _fill_experiment_from_parsing_metadata(experiment, metadata)
  File "/home/reich3/lab/aeye-lab/pymovements/src/pymovements/gaze/io.py", line 614, in _fill_experiment_from_parsing_metadata
    raise ValueError(
ValueError: Experiment metadata does not match the metadata in the ASC file:
- Right eye tracked: True != False
```

dkrako: I don't understand why `experiment.eyetracker.right` is not `None`?

dkrako: do all `Gaze`  objects in `Dataset.gaze` share the reference to the same instance of `DatasetDefinition`?

Then that's the problem. Somewhere inbetween a copy should have been added.

dkrako: but as proposed in #1134 the copy should already be applied earlier for all types of gaze files:

https://github.com/aeye-lab/pymovements/blob/cb550996e4493463b40dde462c61a5b80223430b/src/pymovements/dataset/dataset_files.py#L257-L263

dkrako: This change wasn't in your PR before the merge, the issue lies deeper.

It will be fixed by the changing line 260. This will fix this bug for all types of gaze files, not just this one.

Deepcopying on initalization isn't good practice, as it creates unexpected restrictions for the user.

dkrako: Please create a separate PR with tests for the change.

src/pymovements/datasets/colagaze.yaml
SiQube: #1136

src/pymovements/datasets/colagaze.py
```diff
@@ -0,0 +1,185 @@
...
+                    'md5': None,  # type: ignore
+                },
+            ],
+            'precomputed_events': [
+                {
+                    'resource':
+                    'https://files.au-1.osf.io/v1/resources/gj2uk/providers/osfstorage/'
+                    '67e14ce0f392601163f33215/?view_only=a8ac6e0091e64d0a81d5b1fdec9bab6e&zip=',
+                    'filename': 'fixations.zip',
+                    'md5': None,  # type: ignore
```
dkrako, line 128: why is the checksum null?

SiQube: checksum changes with every download

src/pymovements/datasets/colagaze.py
```diff
@@ -0,0 +1,185 @@
...
+                    'https://files.au-1.osf.io/v1/resources/gj2uk/providers/osfstorage/'
+                    '67e14ce0f392601163f33215/?view_only=a8ac6e0091e64d0a81d5b1fdec9bab6e&zip=',
+                    'filename': 'raw_data.zip',
+                    'md5': None,  # type: ignore
+                },
+            ],
+            'precomputed_events': [
+                {
+                    'resource':
+                    'https://files.au-1.osf.io/v1/resources/gj2uk/providers/osfstorage/'
```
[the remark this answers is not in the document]
SiQube: changed -- was the mirror I was assigned to when looking for the download link

src/pymovements/datasets/colagaze.py
dkrako: I see. I created issue #1152 to resolve this.

dkrako:
Alright, #1151 is merged so now we're ready to merge this too (after resolving the merge conflict).

dkrako (approved):

src/pymovements/datasets/colagaze.yaml
```diff
@@ -0,0 +1,57 @@
+name: "CoLAGaze"
+
+long_name: 'Corpus of Eye Movements for Linguistic Acceptability'
+
+has_files:
+  # need to fix bug:
+  #     ValueError: Experiment metadata does not match the metadata in the ASC file:
+  #     - Right eye tracked: True != False
```
dkrako, line 8: this comment can now be removed, right?

src/pymovements/datasets/colagaze.py
```diff
@@ -0,0 +1,185 @@
...
+    # pylint: disable=similarities
+    # The PublicDatasetDefinition child classes potentially share code chunks for definitions.
+
+    name: str = 'CoLAGaze'
+
+    long_name: str = 'Corpus of Eye Movements for Linguistic Acceptability'
+
+    has_files: dict[str, bool] = field(
+        default_factory=lambda: {
+            # Note: Gaze is currently disabled due to metadata mismatch issue.
```
dkrako, line 104: this comment can now be removed, right?

src/pymovements/datasets/colagaze.yaml
SiQube: will check :-) I hope so

src/pymovements/datasets/colagaze.py
SiQube: will check, I hope so

SiQube: changed

SiQube:
tested via 

```console
pytest -k colagaze
```

PR #1103, opened 2025-04-18
ref: rename aois as frame in TextStimulus

SiQube:
to be in line with the rest of the classes (GazeDataFrame, EventDataFrame, etc.), rename `aois` to `frame`.

this is a **breaking change**

### DIFF
--- a/src/pymovements/stimulus/text.py
+++ b/src/pymovements/stimulus/text.py
@@ -34,7 +34,7 @@ class TextStimulus:
 
     Parameters
     ----------
-    aois: pl.DataFrame
+    frame: pl.DataFrame
         A stimulus dataframe.
     aoi_column: str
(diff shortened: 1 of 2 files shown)


dkrako:
Thank you for your initiative! I value consistency, but I wouldn't like to apply this change due to the following reasons:

1. it replaces the very descriptive field `aois` with a generic `frame` that doesn't describe any content.

2. `GazeDataFrame.frame` will be deprecated soon in favor of `Gaze.samples` (next to `Gaze.events`).

3. All other classes that contain a `frame` field will be deprecated "soon".

4. It is a breaking change that could be easily phased out by deprecating first.

As a side note, `aois` should be (an optional) component of any stimulus class.

I'll close the PR, but feel free to reopen or comment for further discussion.

PR #1104, opened 2025-04-18
dataset: add PoTeC fixation resources

SiQube:


src/pymovements/datasets/potec.py
```diff
@@ -149,10 +150,22 @@ class PoTeC(DatasetDefinition):
                     'md5': 'cffd45039757c3777e2fd130e5d8a2ad',
                 },
             ],
+            'precomputed_events': [
+                {
+                    'resource': 'd8pyg/',
```
dkrako, line 155: this won't work as mirrors was removed from the definition

dkrako (changes requested):

src/pymovements/datasets/potec.py
SiQube: done

dkrako (approved):
Great, thanks a lot!

PR #1106, opened 2025-04-20
fix: Use GAZE_COORDS message to determine screen resolution

### LINKED ISSUE #1086: Use GAZE_COORDS instead of DISPLAY_COORDS for resolution detection
## Current Behavior

`parse_eyelink()` currently uses the `DISPLAY_COORDS` message to detect the screen resolution. This is an optional message to define the resolution for Data Viewer, and some ASC files may not contain such a message. `from_asc()` results in an error in such cases.

## Expected Behavior

The `GAZE_COORDS` should be used because (as far as I can tell) it is always logged at the beginning of a trial. Theoretically, it might be possible to use a different resolution for each trial, but I don't expect this to be common, so we could just proceed in the same way as with the sampling rate (#887), where we warn the user about inconsistent resolution settings.

- ASC files without `DISPLAY_COORDS` message parsed correctly

saeub:
## Description

Follow-up on #887. Screen resolution should be parsed from the `GAZE_COORDS` messages instead of the optional `DIAPLAY_COORDS` message. `GAZE_COORDS` is logged before the start of a new recording block, after `RECCFG`, so adding it to the `recording_config` metadata makes sense. This also checks and warns the user if the resolution has been changed throughout the session. The `resolution` key in the metadata dictionary is kept, so no breaking change.

## Implemented changes

- [x] Use `GAZE_COORDS` message instead of `DISPLAY_COORDS` message to parse screen resolution
- [x] Include screen resolution in `recording_config`, because it can change every time recording is started
- [x] Check for consistent screen resolution, analogous to #887

## How Has This Been Tested?

- [x] Added test case for inconsistent screen resolution
- [x] Tested with a real ASC files that doesn't have a `DISPLAY_COORDS` message

## Type of change

- Bug fix

## Context

Resolves #1086

If #945 is merged first, pylint's complaint about "too many statements" will disappear. (`parse_eyelink()` is quite bloated at the moment, but #945 will make it slimmer.)

## Checklist:


src/pymovements/gaze/_utils/parsing.py
```diff
@@ -48,7 +48,6 @@
             r'\s+(?P<day>\d\d?)\s+(?P<time>\d\d:\d\d:\d\d)\s+(?P<year>\d{4})\s*'
         ),
         r'\*\*\s+(?P<version_2>EYELINK.*)',
-        r'MSG\s+\d+[.]?\d*\s+DISPLAY_COORDS\s*=?\s*(?P<resolution>.*)',
```
dkrako, line 51: instead of removing this regex completely, can you add the parsed info to the metadata dictionary? the regex group name and the metadata key can be simply `DISPLAY_COORDS`.

src/pymovements/gaze/_utils/parsing.py
```diff
@@ -416,11 +424,6 @@ def _pre_process_metadata(metadata: defaultdict[str, Any]) -> dict[str, Any]:
         metadata['version_1'], metadata['version_2'],
     )
 
-    if 'resolution' in metadata:
```
dkrako, line 419: as above: instead of removing this completely, rename `resolution` to `DISPLAY_COORDS`.

tests/unit/gaze/_utils/_parsing_test.py
```diff
@@ -45,7 +45,7 @@
 lines
 MSG	2154555 RECCFG CR 1000 2 1 L
 MSG	2154555 ELCLCFG BTABLER
-MSG	2095865 DISPLAY_COORDS 0 0 1279 1023
```
dkrako, line 48: instead of replacing this, just add a second line with the `GAZE_COORDS`

tests/unit/gaze/io/asc_test.py
```diff
@@ -323,7 +323,7 @@ def test_from_asc_fills_in_experiment_metadata(file, metadata):
                 'screen_height_px': 1080,
                 'sampling_rate': 1000,
             },
-            ['Screen resolution: (1920, 1080) != (1280, 1024)'],
+            ['Screen resolution: (1920, 1080) != (1280.0, 1024.0)'],
```
dkrako, line 326: the error message is better with integers instead of floats, as the type of `Screen.width_px` is `int`.

src/pymovements/gaze/io.py
```diff
@@ -510,7 +510,9 @@ def _fill_experiment_from_parsing_metadata(
     # Screen resolution (assuming that width and height will always be missing or set together)
     experiment_resolution = (experiment.screen.width_px, experiment.screen.height_px)
     if experiment_resolution == (None, None):
-        experiment.screen.width_px, experiment.screen.height_px = metadata['resolution']
+        width, height = metadata['resolution']
+        experiment.screen.width_px = round(width)
```
dkrako, line 514: Can you use `math.ceil()` instead of `round()`?

dkrako (changes requested):
Great, thank you so much! Just a few minor change requests.

Regarding the `too-many-statements`: I'm also fine with adding a pytest ignore in the file for now, and remove it in #945

src/pymovements/gaze/_utils/parsing.py
saeub: Added it back, but as `(left, top, right, bottom)` instead of `(width, height)` (otherwise I think the name `DISPLAY_COORDS` seems confusing)

tests/unit/gaze/_utils/_parsing_test.py
saeub: Added it before the RECCFG (where it is usually logged)

saeub:
@dkrako Thanks a lot for the review! I think I fixed all the remaining issues now.

I'll remove the `pylint: disable` in #945.

src/pymovements/gaze/_utils/parsing.py
```diff
@@ -416,10 +426,9 @@ def _pre_process_metadata(metadata: defaultdict[str, Any]) -> dict[str, Any]:
...
-        coordinates = [int(coord) for coord in metadata['resolution'].split()]
-        resolution = (coordinates[2] - coordinates[0] + 1, coordinates[3] - coordinates[1] + 1)
-        metadata['resolution'] = resolution
+    if 'DISPLAY_COORDS' in metadata:
+        display_coords = tuple(float(coord) for coord in metadata['DISPLAY_COORDS'].split())
```
dkrako, line 430: the display coords are always integers, aren't they? if so, then cast to `int` instead.

You can resolve this comment by yourself to auto-merge.

dkrako (approved):
Great, thank you for this! There's just a minor comment about the type cast that you can resolve yourself to automerge.

PR #1114, opened 2025-04-22
ci: set fail-fast to false in integration tests

dkrako:
## Description

`fail-fast` is `true` by default, which leads to cancelled integration tests if one of the datasets fails. This PR sets `fail-fast` to `false`.

### DIFF
--- a/.github/workflows/datasets.yml
+++ b/.github/workflows/datasets.yml
@@ -30,6 +30,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: [prepare-matrix]
     strategy:
+      fail-fast: false
       matrix:
         dataset: ${{ fromJSON(needs.prepare-matrix.outputs.all_datasets) }}
     steps:


SiQube:
reasoning: at least check whether the rest works? kind of waste of compute (even though its not ours) since they all rerun when the failing one is (hopefully) fixed.

dkrako:
> reasoning: at least check whether the rest works? kind of waste of compute (even though its not ours) since they all rerun when the failing one is (hopefully) fixed.

You're right, my reasoning is to check whether the rest works, because I had to click and wait more than 15 times for this run: https://github.com/aeye-lab/pymovements/actions/runs/14600372728

The integration tests run only for tagged commits, so about 1-3 times a month.
I would assume that the amount of traffic is kept within reasonable limits and shouldn't make much of a difference for any hosting server.

Also when we use the MouseCursor fail from today, I ran all the cancelled integration tests either way to make sure that the remaining datasets pass. so failing fast actually didn't reduce any traffic at all.

So although there's some trade-off to be made, I think that the waste of compute/traffic is negligible as we already make sure to run the integration tests very rarely.

Instead of making our life harder when tests fail, instead we could further reduce traffic by using github artifacts. I created an issue for this: #1117

dkrako:
@SiQube can you please approve this PR? We just had again a lot of test runs cancelled because `GazeBaseVR` had some kind of problem.

SiQube (approved):

SiQube:
IMHO: as main developer you should just be able to merge withoujt review

dkrako:
haha it's probably best to keep my power in check :smiling_imp:

PR #1121, opened 2025-04-24
feat: make `DatasetDefinition.long_name` optional

dkrako:
## Description

Make `DatasetDefinition.long_name` optional and use `DatasetDefinition.name` in disclaimer in case `DatasetDefinition.long_name` is `None`.

## Implemented changes

- [x] move disclaimer to `Dataset._disclaimer()` and log in `Dataset.download()`
- [x] change attribute to `Dataset.long_name: str | None = None`

## How Has This Been Tested?

No additional tests written. I should probably write at least two tests.

## Type of change

- New functionality

## Checklist:

### DIFF
--- a/src/pymovements/dataset/dataset.py
+++ b/src/pymovements/dataset/dataset.py
@@ -1108,7 +1108,7 @@ class Dataset:
         if len(self.gaze) == 0:
             raise AttributeError('no files present in gaze attribute')
 
-    def _disclaimer() -> str:
+    def _disclaimer(self) -> str:
         """Return string for dataset download disclaimer."""
         if self.definition.long_name is not None:
             dataset_name = self.definition.long_name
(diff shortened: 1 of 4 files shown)


src/pymovements/dataset/dataset.py
```diff
@@ -40,6 +41,10 @@
 from pymovements.reading_measures import ReadingMeasures
 
 
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
```
dkrako, line 45: After some reconsideration I moved the disclaimer to `dataset.py`, similar to your original proposal, but returning a string instead.

The main reason was, that `logger.info` has `dataset/dataset_download.py` as a source. I plan to rename this module to `dataset/_download.py` and then it would just be strange to present an info source from there. `dataset/dataset.py` as a info source would be much more comprehensible by a user.

SiQube (approved):
lgtm

PR #1129, opened 2025-04-27
feat: split EventDataFrame by column

### LINKED ISSUE #1126: `GazeDataFrame.split()` should also split the corresponding `EventDataFrame`
## Description of the problem

When using `gaze.split()` and `gaze` already has events associated with it, those events are lost in the resulting split `GazeDataFrames`. It would be useful to keep those events, especially when they are precomputed (e.g., when EyeLink events are parsed from an ASC file).

## Description of a solution

At least when splitting on `trial_columns` (which is probably the most common use case), I would expect the `EventDataFrame` to also be split and attached to the corresponding `GazeDataFrame`. When splitting on columns that are not present in the `EventDataFrame`, that is of course not possible -- in that case I would print a warning and leave the `EventDataFrame`s empty (as is currently done).

- When using `gaze.split(gaze.trial_columns)`, the resulting `GazeDataFrames` should have corresponding split `EventDataFrames` attached to them.

## Sample Code

```python
>>> gaze = pm.gaze.from_csv("gaze.csv", trial_columns="trial")
>>> gaze.detect("ivt")
>>> split_gaze = gaze.split(["trial"])
>>> split_gaze[0].events
(should contain events)
```

SiQube:
resolves #1126

to reproduce run the script `s.py`:

```python
import pymovements as pm

# get single csv file with multiple trials
pm.Dataset('JuDo1000', 'data').download()

# get experiment for event extraction
ddef = pm.DatasetLibrary.get('JuDo1000')

# load file from csv
gaze = pm.gaze.from_csv(
    './data/raw/JuDo1000/110_1.csv',
    separator='\t',
    trial_columns='trialId',
    pixel_columns=['x_left', 'x_right', 'y_left', 'y_right'],
    experiment=ddef.experiment,
)
gaze.pix2deg()
gaze.pos2vel()

# detect events
gaze.detect('idt')
split_gaze = gaze.split('trialId')
print(split_gaze[0].events)
```

which produces

```console
(venv) siqube@c-cube:~/lab/aeye-lab/pymovements
$ python s.py 
shape: (6, 5)
┌─────────┬──────────┬──────────┬──────────┬──────────┐
│ trialId ┆ name     ┆ onset    ┆ offset   ┆ duration │
│ ---     ┆ ---      ┆ ---      ┆ ---      ┆ ---      │
│ i64     ┆ str      ┆ i64      ┆ i64      ┆ i64      │
╞═════════╪══════════╪══════════╪══════════╪══════════╡
│ 1       ┆ fixation ┆ 22052506 ┆ 22052668 ┆ 162      │
│ 1       ┆ fixation ┆ 22052720 ┆ 22053659 ┆ 939      │
│ 1       ┆ fixation ┆ 22053719 ┆ 22053888 ┆ 169      │
│ 1       ┆ fixation ┆ 22053908 ┆ 22054272 ┆ 364      │
│ 1       ┆ fixation ┆ 22054301 ┆ 22056699 ┆ 2398     │
│ 1       ┆ fixation ┆ 22056755 ┆ 22057533 ┆ 778      │
└─────────┴──────────┴──────────┴──────────┴──────────┘
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -94,6 +94,7 @@ repos:
     rev: v2.3.2
     hooks:
     -   id: autopep8
+        args: [--global-config=.flake8]
 -   repo: https://github.com/PyCQA/autoflake
     rev: v2.3.1
     hooks:
--- a/src/pymovements/gaze/gaze_dataframe.py
+++ b/src/pymovements/gaze/gaze_dataframe.py
@@ -296,15 +296,22 @@ class GazeDataFrame:
             A list of new GazeDataFrame instances, each containing a partition of the
             original data with all metadata and configurations preserved.
         """
+        frames = self.frame.partition_by(by=by)
+        # we cannot directly zip because we need to check if by is in events columns
+        events_list = self.events.split(by) if by in self.events.columns else [
+            pm.EventDataFrame(),
+        ] * len(frames)
+
         return [
             GazeDataFrame(
-                new_frame,
+                frame,
                 experiment=self.experiment,
                 trial_columns=self.trial_columns,
                 time_column='time',
                 distance_column='distance',
+                events=events,
             )
-            for new_frame in self.frame.partition_by(by=by)
+            for frame, events in zip(frames, events_list)
         ]
 
     def transform(
--- a/src/pymovements/events/frame.py
+++ b/src/pymovements/events/frame.py
@@ -269,6 +269,37 @@ class EventDataFrame:
         """
         return EventDataFrame(data=self.frame.clone())
 
+    def split(self, by: Sequence[str]) -> list[EventDataFrame]:
+        """Split the EventDataFrame into multiple frames based on specified column(s).
+
+        Parameters
+        ----------
+        by: Sequence[str]
+            Column name(s) to split the DataFrame by. If a single string is provided,
+            it will be used as a single column name. If a list is provided, the DataFrame
+            will be split by unique combinations of values in all specified columns.
+
+        Returns
+        -------
+        list[EventDataFrame]
+            A list of new EventDataFrame instances, each containing a partition of the
+            original data with all metadata and configurations preserved.
+        """
+        event_pl_df_list = list(self.frame.partition_by(by=by))
+
+        # Ensure column order: trial columns, name, onset, offset.
+        if self.trial_columns is not None:
+            event_pl_df_list = [
+                frame.select([*self.trial_columns, *self._minimal_schema.keys()])
+                for frame in event_pl_df_list
+            ]
+        return [
+            EventDataFrame(
+                frame,
+                trial_columns=self.trial_columns,
+            ) for frame in event_pl_df_list
+        ]
+
     def _add_minimal_schema_columns(self, df: pl.DataFrame) -> pl.DataFrame:
         """Add minimal schema columns to :py:class:`polars.DataFrame` if they are missing.
 
--- a/tests/unit/events/frame_test.py
+++ b/tests/unit/events/frame_test.py
@@ -486,3 +486,46 @@ def test_event_dataframe_add_trial_column_raises_exception(events, kwargs, excep
         events.add_trial_column(**kwargs)
 
     assert message == excinfo.value.args[0]
+
+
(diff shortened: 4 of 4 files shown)


src/pymovements/gaze/gaze_dataframe.py
```diff
@@ -296,15 +296,22 @@ def split(self, by: Sequence[str]) -> list[GazeDataFrame]:
             A list of new GazeDataFrame instances, each containing a partition of the
             original data with all metadata and configurations preserved.
         """
+        frames = self.frame.partition_by(by=by)
+        # we cannot directly zip because we need to check if by is in events columns
+        events_list = self.events.split(by) if by in self.events.columns else [
+            pm.EventDataFrame(),
```
dkrako, line 302: Are you sure this is correct? If `by` is not in the events columns just empty events are created?

dkrako (changes requested):
Please add tests to `gaze_dataframe_test.py` to make sure the functionality is correct

src/pymovements/gaze/gaze_dataframe.py
SiQube: what else would you propose?

src/pymovements/gaze/gaze_dataframe.py
```diff
@@ -296,15 +296,22 @@ def split(self, by: Sequence[str]) -> list[GazeDataFrame]:
             A list of new GazeDataFrame instances, each containing a partition of the
             original data with all metadata and configurations preserved.
         """
+        frames = self.frame.partition_by(by=by)
+        # we cannot directly zip because we need to check if by is in events columns
+        events_list = self.events.split(by) if by in self.events.columns else [
```
dkrako, line 301: this won't work if `by` is a list of strings. please add a test case to make sure list of strings are supported for `by`.

src/pymovements/gaze/gaze_dataframe.py
dkrako: if `by` is not included in the events, then log a warning and clone the events to each split.

saeub:
@SiQube If there's anything I can help with (writing tests or implementation), just let me know.
This feature would be very helpful for a course I'm teaching, so I would love to see this merged soon.

src/pymovements/gaze/gaze_dataframe.py
SiQube: I've been thinking about this a bit more. I believe we should raise a warning and either return an empty result or raise an error.

why?

the only plausible use case for using `split` is when multiple trials are present in a single dataframe. if we proceed by cloning the events across splits, when missing `by`, -- even with a warning -- there’s a risk that a user might unknowingly continue working with incorrect or misleading data.

additionally, since we already compute events (magically) at the trial level, not the dataframe level, replicating events in this way feels misleading and potentially dangerous.

src/pymovements/gaze/gaze_dataframe.py
SiQube: done

src/pymovements/gaze/gaze_dataframe.py
SiQube: we could even think about restricting it to `trial_columns`, which are used for event extraction?

.pre-commit-config.yaml
```diff
@@ -94,6 +94,7 @@ repos:
     rev: v2.3.2
     hooks:
     -   id: autopep8
+        args: [--global-config=.flake8]
```
dkrako: is this change needed? the config is defined in `tox.ini` and we don't have any file `.flake8`.

dkrako: discarding the changes didn't seem to make a difference so I'll leave this out

src/pymovements/gaze/gaze_dataframe.py
dkrako: let's improve on this edge case later on. I created #1193 for the default value and create a PR soon for the warning

dkrako: As the issue isn't that essential I created #1194 for inconsistent columns and won't create a PR

Issue #1021, opened 2025-03-12
implement local pre-commit hooks for adding datasets


dkrako:
## Description of the problem
When adding new datasets contributors currently need to make sure of the following:

- adding the dataset to one of the csv files in `docs/source/datasets`
- if it's a yaml definition, then adding the dataset to `src/pymovements/datasets/datasets.yaml`

For automatically adding yaml files to `datasets.yaml` #914 provides a nice private function `write_dataset_definitions_yaml()` in `src/pymovements/dataset/_utils/_yaml.py`.

It's important to note though, that this function is not targeted for end-users (not even implicitly) and therefore there's no reason to deploy it in a package.

## Description of a solution

Fortunately pre-commit allows specifying local pre-commit hooks that run local repository scripts: https://pre-commit.com/#repository-local-hooks.

Functions like the `write_dataset_definitions_yaml()` should be moved to a root-level directory of the repository, e.g. `bin`, `scripts` (or decide for a more fitting name).

## Context

This is a continuation of this [discussion](https://github.com/aeye-lab/pymovements/pull/914#discussion_r1985379151) in #914

> Let's recap this very standard CI/CD model:
> 
> ![img](https://camo.githubusercontent.com/b94c6d8e302bacd9b5841023c55ad1b9d42c3a64c00e824da8858ccf27e37fc8/68747470733a2f2f6167696c652d636f6163682d61636164656d792e64652f77702d636f6e74656e742f75706c6f6164732f323032312f30342f4465764f70732d506970656c696e652e706e67) (here acceptance tests are what we call functional tests in our pipeline) (as an unrelated side note: lots of tests in `dataset_test.py` are actually bad integration or functional tests and not unit tests)
> 
> All code that exists in our `src` directory goes through these stages, and after deployment it is targeted to be used during application time. Even something which is not intended for direct usage and more or less obfuscated from the user (e.g. a completely private `utils.checks` module) will be run at the application stage after deployment.
> 
> `write_dataset_definitions_yaml()` is something completely different: it is intended to be run before the commit and build stages, and especially before getting into any testing stages. There is not a single use case where this function is called at application stage (*). From CI/CD model perspective, this function is very similar to something that `pre-commit` does.
> 
[...]

PR #1147, opened 2025-05-10
ci: add github workflow to autogenerate sphinx dataset pages

dkrako:
## Description

This PR autogenerates sphinx datasets pages without requiring any dataset definition classes.

For now the datasets pages are very rudimentary, listing only the name of the dataset and a link to the dataset definition class.

## Implemented changes

- [x] create table of contents for datasets
- [x] create templates for dataset pages
- [x] write script to create rst files for each dataset
- [x] extend script to create a metadata file for each dataset

## How Has This Been Tested?

- [x] test that rst files are created
- [x] test that metadata files are created

#### related issues:
- #1021

## Checklist:


docs/source/datasets/datasets.yaml
dkrako, line 1: You may ask yourself why do we need an additional `datasets.yaml` as we already have one in `src/pymovements/datasets/`.

This file here is generated using `DatasetLibrary.names()` and will include all datasets, even those that are only implemented as python classes.

It is required for the `_templates/datasets.rst` template.

docs/source/_templates/datasets.rst
```diff
@@ -0,0 +1,9 @@
+{% for item in data %}
+.. toctree::
+
+    :hidden:
```
dkrako, line 4: Hides the dataset entries for now. This means they are not listed in any table of contents. The pages are still accessible via explicit URLs.

docs/source/datasets/datasets.yaml
dkrako, line 1: You may ask yourself: why isn't this implemented as a pre-commit hook like #1145?

This workflow is much more elaborate than the pre-commit hook. It needs to import pymovements to get all integrated datasets, regardless if they're implemented as python classes or yaml files.

Further, this workflow will be extended in the future, to automatically infer additional metadata. This means, datasets will be downloaded by scripts (or rather: dataset files stored in GitHub artifacts will be accessed).

This would be really overkill for a pre-commit hook that should terminate fast, so it's much better to run these as dedicated GitHub jobs.

.github/workflows/datasets.yml
dkrako, line 1: You may ask yourself: why isn't this implemented as a pre-commit hook like #1145?

This workflow is much more elaborate than the pre-commit hook. It needs to import pymovements to get all integrated datasets, regardless if they're implemented as python classes or yaml files.

Further, this workflow will be extended in the future, to automatically infer additional metadata. This means, datasets will be downloaded by scripts (or rather: dataset files stored in GitHub artifacts will be accessed).

This would be really overkill for a pre-commit hook that should terminate fast, so it's much better to run these as dedicated GitHub jobs.

PR #1150, opened 2025-05-14
chore!: restrict boolean arguments in `to_dict()` to keyword-only

dkrako:
## Description

Allowing the two boolean arguments to be passed as positional arguments is error-prone for switchups and can lead to problems with future enhancements and backwards compatiblity. The arguments should be restricted to keyword only until we have a stable interface.


SiQube (approved):
LGTM

PR #1168, opened 2025-05-27
ci: reduce number of caches on github

dkrako:
## Description

In case of lots of PRs we currently exceed the 10GB cache size pretty quickly.

This PR reduces the number of different caches to be more space efficient.

## Implemented changes

- [x] only save cache on main branch, PR branches will now only load the cache
- [x] don't use cache in setup-python instead cache both pip and tox environment in one go
- [x] hash not just `tox.ini` for cache key but also `pyproject.toml` and associated github workflow file


SiQube (approved):
Don't really know caching

PR #1169, opened 2025-05-27
build: restrict setuptools to <80.9.0

dkrako:
quick fix for making our main branch pass again. #1167 is related and the better fix, but will be handled afterwards.


dkrako:
closed in favor if #1167

PR #1171, opened 2025-05-27
ci: revert removing pybtex from test dependencies

dkrako:
The cached environments made the PR tests pass before merging. Then a new cache was build on main and they failed.


dkrako:
what a circlejerk, the cached dependencies (excluding pybtex) now made all the tests fail.
I cleared all caches. now all tests pass...

SiQube (approved):

PR #1211, opened 2025-08-18
docs: add scanpathplot to online documentation

dkrako:
## Description

scanpathplot was missing


dkrako:
> maybe we can automize this?

created #1228

PR #1221, opened 2025-08-18
docs: upgrade `Gaze.split()` documentation

jewelsg:
- added some example parameters
- improving the documentation for the split method for better understanding why it is useful for the large datasets.


src/pymovements/gaze/gaze.py
```diff
@@ -319,11 +319,23 @@ def split(self, by: Sequence[str] | None = None) -> list[Gaze]:
             will be split by unique combinations of values in all specified columns.
             If None, uses trial_columns. (default=None)
 
+            Examples:
```
dkrako: can you move this to a dedicated example section below the notes section?

like this:

```
Examples
--------
- `by=["trial"]` → split by trial; each trial becomes a separate Gaze object.
- `by=["subject_id", "trial"]` → split by subject and trial; one Gaze object per subject-trial pair.
```

src/pymovements/gaze/gaze.py
```diff
@@ -319,11 +319,23 @@ def split(self, by: Sequence[str] | None = None) -> list[Gaze]:
...
+
         Returns
         -------
         list[Gaze]
             A list of new Gaze instances, each containing a partition of the
             original data with all metadata and configurations preserved.
+
+        Notes
+        ------
+            - This method is particularly useful for large datasets or experiments where
```
SiQube: `experiments` is wrong here and can lead to confusion: within `pymovements`, `experiment` generally refers to specifications of `Experiment`, see [experiment](https://github.com/aeye-lab/pymovements/blob/main/src/pymovements/gaze/experiment.py)

dkrako: yes probably it's best to only write "...for large datasets where..."

src/pymovements/gaze/gaze.py
dkrako: additonally, it would be very nice to have a more complete code example. You could write something like this:

```
Examples
--------
First let's create a simple samples dataframe:

>>> import numpy as np
>>> import polars as pl
>>> import pymovements as pm
>>> samples = pl.from_dict({'x': range(100), 'y': range(100), 'trial': np.repeat([1, 2, 3, 4, 5], 20)})

Then let's initialize our `Gaze` object:

>>> gaze = pm.Gaze(samples=samples, trial_columns='trial')
>>> gaze

Now we can split the gaze by the trial column values into separate objects:

>>> gaze.split(by='trial')
```

You can further extend this example to have an additional column to split by.

dkrako (changes requested):
the example should be extended a bit more to be a full code exmaple

PR #1227, opened 2025-08-18
docs: fix print statement of disclaimer

SiQube:
changes disclaimer from

```console
INFO:pymovements.dataset.dataset:        You are downloading the pymovements Toy Dataset. Please be aware that pymovements does not

```

to

```console
INFO:pymovements.dataset.dataset:
        You are downloading the pymovements Toy Dataset. Please be aware that pymovements does not

```


dkrako (approved):
great! thanks a lot!

PR #1240, opened 2025-08-19
chore: gitignore data directory

jewelsg:
by mistake , the data folder can be uploaded with commit. So, to avoid this situation, gitignore file has been updated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ build/
 develop-eggs/
 dist/
 downloads/
+data/
 eggs/
 .eggs/
 lib/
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -25,6 +26,7 @@ share/python-wheels/
 .installed.cfg
 *.egg
 MANIFEST
+*.ipynb
```
cbueth: Are you sure `*.ipynb` Notebooks should also all be ignored? That would mean that all new notebooks, e.g. in `docs/source/tutorials/` would needed to be forcefully updated.

SiQube: please delete this

dkrako: haha thank you both for reviewing :)

dkrako:
closed in favor of #1245

PR #1243, opened 2025-08-19
feat: add fixations, saccades, blinks, microsaccades as class properties

bujkani:
## Description

This PR introduces convenient accessors for specific eye movement event types in the `Events` class, allowing users to directly access fixations, saccades, blinks, and microsaccades using property methods.

Previously, users needed to filter the general `.events` DataFrame using column expressions, e.g.:

```python
gaze.events.frame.filter(pl.col("name").str.starts_with("fixation"))
```

This PR simplifies that by providing:

```python
gaze.fixations
gaze.saccades
gaze.blinks
gaze.microsaccades
```

This change is implemented in a backward-compatible way. It does not introduce any additional dependencies.

## Implemented changes

- [x] Added `fixations`, `saccades`, `blinks`, and `microsaccades` read-only properties to the `Events` class
- [x] Implemented logic based on `.filter(pl.col("name").str.starts_with(...))` for each event accessor
- [x] Added unit tests for all new properties under `tests/unit/events/events_test.py`
- [x] Verified integration with existing `.events` workflows

## How Has This Been Tested?

- [x] Added tests using a helper `_edf()` to mock minimal event DataFrames
- [x] Verified that the returned results match expected filters by event name prefixes
- [x] Ran the full test suite locally (Python 3.13.3 with Pyenv) with all existing and new tests passing

### Test examples:
```python
edf = _edf(["fixation", "saccade", "blink", "microsaccade"])
assert edf.fixations["name"].to_list() == ["fixation"]
assert edf.saccades["name"].to_list() == ["saccade"]
assert edf.blinks["name"].to_list() == ["blink"]
assert edf.microsaccades["name"].to_list() == ["microsaccade"]
```

## Type of change

- [x] New functionality
- [x] Documentation update
- [ ] Bug fix
- [ ] Performance improvement
- [ ] Breaking change
- [ ] New dataset

## Checklist:

- [x] I have added tests that prove my feature works


src/pymovements/events/events.py
```diff
@@ -287,6 +287,70 @@ def event_property_columns(self) -> list[str]:
         event_property_columns -= set(self._additional_columns)
         return list(event_property_columns)
 
+    def _filter_by_prefix(self, prefix: str) -> pl.DataFrame:
```
SiQube, line 290: why does this need to be "private" -- I think it would be a great function in general for the user? familiarity with polars is low so this could be an excellent way of prevent users to learn polars syntax.

dkrako: good idea, we'll do this in a follow-up PR

PR #1245, opened 2025-08-19
chore: gitignore data directory

jewelsg:
by mistake , the data folder can be uploaded with commit. So, to avoid this situation, gitignore file has been updated.

***      *.ipynb condition has been removed not to ignore anymore.


dkrako (approved):
Thanks again!

Issue #686, opened 2024-02-22
add support for binocular eyelink parsing


dkrako:
## Description of the problem
The lines of binocular and monocular eyelink recordings vary, as they have more data fields for each eye tracker sample.

## Description of a solution
Currently only the regex for monocular samples is defined as `EYE_TRACKING_SAMPLE` in `utils.parsing`.
This should be renamed into `EYE_TRACKING_SAMPLE_MONOCULAR` and a new `EYE_TRACKING_SAMPLE_BINOCULAR` for binocular data should be created.

There is a regex implemented in #674 which checks sampling rate and which eye is tracked:
`r'SAMPLES\s+GAZE\s+(?P<tracked_eye>LEFT|RIGHT)\s+RATE\s+(?P<sampling_rate>[-]?\d*[.]\d*)\s+TRACKING\s+....`

In case of a binocular recording, the file will contain two lines for each tracked eye.

The metadatafield `tracked_eye` will currently only have `LEFT` or `RIGHT`, but in the binocular case it should be set to `BOTH` internally.

You should check early on which of the two regular expressions for the eye tracking samples should be used for the remaining file.

The parsing will have to be changed a bit then, by parsing header information first and only after having read the `SAMPLES\s+GAZE\s+...` line(s) the eye tracking sample would be checked too (`EYE_TRACKING_SAMPLE_MONOCULAR` or `EYE_TRACKING_SAMPLE_BINOCULAR`).

A simpler alternative is to always check for both samples, but I guess this would be much slower for one of the two cases (e.g. in case `EYE_TRACKING_SAMPLE_MONOCULAR` is matched first, binocular files would parse slower as the line has to be matched a second time with `EYE_TRACKING_SAMPLE_BINOCULAR`).

## Minimum acceptance criteria
- [ ] parse binocular eyelink data
- [ ] for binocular data replace the `tracked_eye` metadata with `BOTH`
- [ ] add eyelink binocular test example file to repository

dkrako:
These sample line formats are relevant and should all be supported eventually:

> From the User Manual of the EyeLink Portable Duo (Version 1.0.9), section 4.9.2 (p. 92):
> 
> > **SAMPLE LINE FORMATS**
> > 
> > * Monocular:
> >   `<time> <xp> <yp> <ps>`
> > * Monocular, with velocity
> >   `<time> <xp> <yp> <ps> <xv> <yv>`
> > * Monocular, with resolution
[...]

PR #1255, opened 2025-08-20
feat: parse binocular eyelink files

bernhardangele:
## Description
This pull request adds support for parsing both monocular and binocular EyeLink ASC files in parsing.py. The main changes include introducing separate regex patterns for monocular and binocular samples.
Which REGEX is used depends on the line SAMPLES (the RECCFG message states what was recorded, but the actual ASC file content can differ if the -r or -l flags were used in edf2asc).
Added a new metadatafield to distinguish between tracked and recorded eyes 

**Parsing improvements:**
* Added separate regex patterns for monocular (`EYE_TRACKING_SAMPLE_MONOCULAR`) and binocular (`EYE_TRACKING_SAMPLE_BINOCULAR`) EyeLink sample lines, allowing accurate parsing for both sample types.
* Added regex (`SAMPLES_CONFIG_REGEX`) to extract sample configuration details (e.g., which eyes are tracked) from the SAMPLES lines. Added a new function `_check_samples_config_key` that follows the same logic as `_check_reccfg_key`.
* Detect whether recording is binocular or monocular in `parse_eyelink`, and update the samples dictionary and parsing logic to handle the correct set of fields for each case. 
* Added a new metadata field `recorded_eye` (obtained from RECCFG) and `tracked_eye` (obtained from from SAMPLES line).

**Schema and data structure updates:**
* Extended schema overrides in `parse_eyelink` to include binocular fields (`x_left_pix`, `y_left_pix`, etc.) when appropriate.

## How Has This Been Tested?
* All tests in `_parsing_test.py` work.
* Added new test `test_parse_eyelink_binocular_simple` to confirm that the binocular data structure is read correctly.

## Type of change

- New functionality

## Context

Addresses  #686, but changes in `from_asc.py` are still required.


bernhardangele:
Now from_asc should be able to handle binocular files, although I am not sure if the format is exactly correct

src/pymovements/gaze/_utils/parsing.py
```diff
@@ -192,26 +218,52 @@ def get_pattern_keys(compiled_patterns: list[dict[str, Any]], pattern_key: str)
...
-    if BLINK_START_REGEX.match(line):
-        return 'blink'
+    """Check if the line contains the start of an event and return the event name.
+
+    Returns event names with eye suffixes: e.g. 'fixation_left' or 'saccade_right'.
+    """
+    if match := FIXATION_START_REGEX.match(line):
+        eye = match.group('eye').upper()
+        suffix = '_left' if eye == 'L' else '_right'
+        return f'fixation{suffix}'
```
dkrako: I reconsidered this a bit. Maybe it would be better to put this information into a separate events column named `eye`.

This way it will be much easier to process later on.

So I would propose to change the function signature to:

```python
def parse_eyelink_event_start(line: str) -> tuple[str, str] | None:
```

If an event was parsed, the returned value would then be a tuple of the event name (`fixation`, `saccade`, `blink`) and the respective eye ( `left`, `right`).

src/pymovements/gaze/_utils/parsing.py
```diff
@@ -192,26 +218,52 @@ def get_pattern_keys(compiled_patterns: list[dict[str, Any]], pattern_key: str)
...
+        suffix = '_left' if eye == 'L' else '_right'
+        return f'saccade{suffix}'
+    if match := BLINK_START_REGEX.match(line):
+        eye = match.group('eye').upper()
+        suffix = '_left' if eye == 'L' else '_right'
+        return f'blink{suffix}'
     return None
 
 
 def parse_eyelink_event_end(line: str) -> tuple[str, float, float] | None:
```
dkrako: This would then be changed to

```python
def parse_eyelink_event_end(line: str) -> tuple[str, str, float, float] | None:
```

Alternatively you could also return a dictionary, but as this is an internal function I think a tuple is perfectly fine.

dkrako (changes requested):

src/pymovements/gaze/_utils/parsing.py
bernhardangele: This should be addressed by the new commit

src/pymovements/gaze/_utils/parsing.py
```diff
@@ -303,13 +362,58 @@ def parse_eyelink(
...
     num_blink_samples = 0
+    blink_intervals: list[tuple[float, float]] = []
     blinking = False
 
+    # Detect if the file is binocular or monocular
+    # Collect ALL SAMPLES config lines (don't stop at the first match) so
+    # inconsistent values (e.g. different sampling rates across SAMPLES lines)
+    # can be detected later.
+    is_binocular = False
+    for line in lines:
```
dkrako, line 379: I think this is fine for this PR. In the future we should try to implement this more dynamically and prevent traversing the file twice. I'll write a follow-up issue on this: #1274

tests/unit/gaze/io/asc_test.py
```diff
@@ -530,30 +599,54 @@ def test_from_asc_raises_exception(kwargs, exception, message_prefix):
                     mount='Desktop',
                 ),
             ),
-            id='500hz_no_dummy',
+            id='monocular_500hz_no_dummy',
+        ),
+
+        pytest.param(
```
dkrako, line 605: this test fails as strangely both `screen` and `eyetracker` field differ.

the actual result is:

```
Experiment(screen=Screen(width_px=1921, height_px=1081, width_cm=None, height_cm=None, distance_cm=None, origin=None), eyetracker=EyeTracker(sampling_rate=1000.0, left=False, right=False, model='EyeLink Portable Duo', version='6.14', vendor='EyeLink', mount='Desktop'))
```

I would expect

```
Experiment(screen=Screen(width_px=1920, height_px=1080, width_cm=None, height_cm=None, distance_cm=None, origin=None), eyetracker=EyeTracker(sampling_rate=1000.0, left=True, right=True, model='EyeLink Portable Duo', version='6.14', vendor='EyeLink', mount='Desktop'))
```

 @bernhardangele can you investigate?

dkrako (changes requested):
I added some tests and refactored the `asc_test.py` file. It now contains some fixtures to create test files such that testing is a bit easier. The fixtures could probably be reused in `_parsing_test.py`, but that can be left for a future PR.

What's strange though, is that one of my tests didn't lead to a parsed expected experiment. I added a comment to the test in question. This should be resolved before merging.

src/pymovements/gaze/io.py
```diff
@@ -33,6 +34,11 @@
 from pymovements.gaze.gaze import Gaze
 
 
+# Note: column detection for ASC files now uses simple substring matching
```
dkrako: this comment isn't probably that relevant on the module level. You can move these lines to the relevant code around lines 480 - 500.

tests/unit/gaze/io/asc_test.py
bernhardangele: This sounds like it's related to the issue of the screen dimensions. The eyelink_binocular_example.asc I added for testing (like probably all the files created when using PyGaze) had width set to 1920 and height set to 1080, [contrary to the recommendation by SR Research](https://www.sr-research.com/support/thread-9129.html):
`MSG	1408659 GAZE_COORDS 0.00 0.00 1920.00 1080.00`

 The eyelink_monocular_example.asc has the correct specification: 

`MSG	2095865 DISPLAY_COORDS 0 0 1279 1023`

The issue is this in parsing.py:
```python
            elif match := RESOLUTION_REGEX.match(line):
            left, top, right, bottom = (float(coord) for coord in match.group('resolution').split())
            # GAZE_COORDS is always logged after RECCFG -> add it to the last recording_config
            recording_config[-1]['resolution'] = (right - left + 1, bottom - top + 1)
```
This adds the extra width and height pixels.

I guess the easiest solution would be to change the eyelink_binocular_example.asc, but I that won't solve the issue with inconsistent screen dimensions between Experiment Builder and PyGaze files.

dkrako: I see, so it's basically an issue with the pygaze output. I guess the `screen` is not an issue then. I would leave it as is. Is there a way to infer if a file was created with pygaze? 

What's left is the unexpected `eyetracker`:

the actual result is:

```
EyeTracker(sampling_rate=1000.0, left=False, right=False, model='EyeLink Portable Duo', version='6.14', vendor='EyeLink', mount='Desktop')
```

I would expect

```
EyeTracker(sampling_rate=1000.0, left=True, right=True, model='EyeLink Portable Duo', version='6.14', vendor='EyeLink', mount='Desktop')
```

dkrako: As the file is binocular, `left` and `right` should both be `True`.

dkrako: I addressed the issue with the resultion in #1286

dkrako: there was mismatch between space and tab in the samples config which lead to both `left` and `right` being `False`.

I fixed this.

src/pymovements/gaze/io.py
dkrako: done

bernhardangele:
> I just resolved some of the minor issues and finalized this PR to be ready for merge.
> 
> Thank you so much for this enhancement! @bernhardangele

Sorry I wasn't able to finish it! I will look at adjusting the documentation again to fit this enhancement when I get a chance (should I create a new issue?)

dkrako:
> > I just resolved some of the minor issues and finalized this PR to be ready for merge.
> > Thank you so much for this enhancement! @bernhardangele
> 
> Sorry I wasn't able to finish it! I will look at adjusting the documentation again to fit this enhancement when I get a chance (should I create a new issue?)

Ah don't worry about this, I'm glad I could help out with the details and they were not that obvious.

I just updated the docstring to hopefully reflect all the changes. Feel free to create a new issue/PR if you want to make further improvements.

PR #1247, opened 2025-08-19
feat: automatically set axis dimension in traceplot

shaitarAn:
## Description

Automatically set the data coordinates in traceplot based on the experiment.screen dimensions if they are specified.

## Implemented changes

- [x] Update traceplot.py to pass screen dims info to matplot set_xlim, set_ylim to set the data coordinates
- [x] raised error if experiment.screen.origin not 'upper left'
- [x] update figsize to be proportional to experiment.screen dims

## How Has This Been Tested?

- [x] Ran traceplot on a dataset with/without experiment data
- [x] Ran traceplot on a dataset with/without experiment.screen_height_px and experiment.screen_width_px
- [x] Ran traceplot on a dataset with/without experiment.screen.origin != 'upper left'

## Type of change

Remove irrelevant items:
- Performance improvement

## Context

#1230 is resolved partially because scanpathplot has been filed under a separate issue

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

PR #1256, opened 2025-08-20
depr: deprecate scanpathplot events argument

dkrako:
## :warning: Deprecation

The keyword argument `events` of the `scanpathplot()` method is deprecated. Please use `gaze` instead.
This argument will be removed in pymovements v0.28.0.

## Description

Having both an `events` and a `gaze` keyword argument in `scanpathplot()` is confusing, as `events` is now already included in `Gaze`. This PR deprecates the `events` keyword argument in favor or the `gaze` keyword argument.

This way it will be more straight forward to implement #1247 also for the `scanpathplot()`.

## Implemented changes

- [x] deprecate `events` argument


src/pymovements/plotting/traceplot.py
```diff
@@ -42,7 +42,7 @@
 def traceplot(
         gaze: Gaze,
         position_column: str = 'pixel',
-        cval: np.ndarray | None = None,  # pragma: no cover
+        cval: np.ndarray | None = None,
```
dkrako, line 45: this must've been a strange copypaste error from a few lines above. don't know how this happened.

SiQube (approved):

Issue #1230, opened 2025-08-19
automatically set axis dimension in pixel-data plots


dkrako:
## Description of the problem
Currently the user needs to manually specify the plot dimensions when using scanpathplot and traceplot.

## Description of a solution
As the information is already present in `Gaze.experiment.screen`, it would be nice to set the plot dimensions accordingly by default.

dkrako:
#1247 is being merged. I updated the `scanpathplot()` function in #1256 such that implementing the axis dimensions will be less confusing there. The `heatmap()` plot should also be enhanced with this.

Here's a checklist on remaining tasks:

- [x] `traceplot()` #1247
- [ ] `scanpathplot()`
- [ ] `heatmap()`

dkrako:
> Regarding heatmap(), it looks like the axis dims can be fixed by swapping the y_edges in the extent so that the coordinates match the stimulus. 

I'm not sure if I understand correctly. Do you mean changing this

https://github.com/aeye-lab/pymovements/blob/348acee650537da5be10f0f7bee2ebd370a13921/src/pymovements/plotting/heatmap.py#L140-L146

to this?

```python
    heatmap_plot = ax.imshow(
        heatmap_value,
        cmap=cmap,
        origin=origin,
        interpolation=interpolation,
        extent=[x_edges[0], x_edges[-1], y_edges[-1], y_edges[0]],
    )
```

I'm not sure if this will work correctly. Can you create two plots to show the difference?

Moreover as the `x_edges, y_edges` are inferred from the sample data in `Gaze`, it would be best to pass the actual dimensions from `Screen` if they are available.

> Also, changing the default origin to "upper" makes it consistent with traceplot()/scanpathplot(). Is there a reson why it's set to "lower"?

Do you mean the default value in `_draw_image_stimulus()`? You can change it to 'upper' if you like, it's probably set accidentally this way.

> Aspect ratio behaves a bit differently with imshow (it stretches to fill the axes). it seems to produce a correct aspect ratio already, or am I not looking in the right place?

It will only produce the correct axes if there's a stimulus image provided (with the correct resolution). Without the image this isn't be guaranteed, so you should just use the same function as you already do in #1295.

PR #1257, opened 2025-08-20
feat!: Add axes in- and output to plotting functions

### LINKED ISSUE #1231: feat: add axis in- and output to all plotting functions
## Description of the problem
The way our plotting functions terminate is inconsistent.

Some are calling `plt.closefig()`, some are not.

## Description of a solution

It would be nice to have an argument in all plotting funtions like `closefig` (default: `True`) .

This way users still have the freedom to further tinker with the plot (for example adding some overlays).

cbueth:
## :boom: Breaking Change

`pm.heatmap(...)` now returns not just a `plt.Figure`, but also the `plt.Axes` in a tuple. This behaviour is unified across all plotting functions.

## Description

All plotting functions allow the in- and output of `matplotlib` axes, making it more flexible to plot into any existing plot, or make further changes to a `pymovements` plot afterwards.

Resolves #1231

## Implemented changes

- [x] Adds the ability to pass and return `fig` and `ax` from all plotting functions.
- [x] Unified figure opening and closing. User can pass ax or let it be created.
- [x] No breaking changes

## How Has This Been Tested?

- [x] Plot opening/closing behavior
- [x] Cover all new code
- [x] All tutorial pages plots produce the same output as before

## Type of change

- New functionality

## Context

Resolves #1231

#### related issues:
- #1230

Looking at the related PR I assume no merge conlicts will arise, and programatically it also looks go to me.

## Checklist:


src/pymovements/plotting/_matplotlib.py
```diff
@@ -84,70 +89,149 @@
...
+    """
+    if savepath is not None:
+        fig.savefig(savepath)
+
+    if show:
+        if own_figure:
+            plt.show()
+        else:
+            warn(
+                f'{func_name}: "show=True" has no effect when plotting into an external Axes.',
```
dkrako: probably it's better to use if instead of when

src/pymovements/plotting/_matplotlib.py
```diff
@@ -84,70 +89,149 @@
...
+                UserWarning,
+                stacklevel=2,
+            )
+
+    if closefig is None:
+        do_close = own_figure
+    else:
+        if not own_figure and closefig:
+            warn(
+                f'{func_name}: "closefig=True" is ignored when an external Axes is provided.',
```
dkrako: probably it's better to use if instead of when

src/pymovements/plotting/heatmap.py
```diff
@@ -49,7 +51,10 @@ def heatmap(
         path_to_image_stimulus: str | Path | None = None,
         stimulus_origin: str = 'upper',
         alpha: float = 1.,
-) -> plt.Figure:
+        *,
+        ax: plt.Axes | None = None,
+        closefig: bool | None = None,
+) -> tuple[plt.Figure, plt.Axes]:
```
dkrako, line 57: It would be great if we could manage to avoid this breaking change. How could we achieve that?

dkrako: We would probably need an additional argument (like `return_only_figure: bool = True`), that would raise a deprecation warning if `True`, with the instruction to set it to `False`. Then after a few minor versions we would remove this altogether.

So the perfect process would be something like:
1. add a `return_only_figure` argument with a deprecated default value (`True`).
2. change the default value to `False` and deprecate the argument.
3. remove the argument and remove the union from the return typehint.

I think this would be very inconvenient for users, as they would need to fiddle with this argument at least two times.

As this is just a plotting function, I would also be ok with having this breaking change.

dkrako (changes requested):
Great job! By refactoring `_matplotlib.py` the module is now much more readable and reusable.

There's basically just the issue left if we want to introduce a breaking change in this PR.

Here I tend to think it's not worth the effort of deprecating in this case, as it's just a plotting function and the deprecation process in this case would probably be more inconvenient for users.

Let's think about it for a day how to proceed.

src/pymovements/plotting/heatmap.py
SiQube: I think breaking here would be fine, especially if we at the same time update the tutorials, i.e. before next release -- is this possible?

dkrako: alright, I'll merge this right after releasing `v0.23.1` today.

src/pymovements/plotting/_matplotlib.py
cbueth: Changed in 7ac1dd6.

src/pymovements/plotting/_matplotlib.py
cbueth: Changed in 7ac1dd6.

PR #1267, opened 2025-08-21
Feature/add load stimuli in dataset

saphjra:
## Implemented changes

- [ ] Change 1
- [ ] Change 2
- [ ] Change 3

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Type of change

Remove irrelevant items:
- Bug fix
- New functionality
- Performance improvement
- New dataset
- Breaking change
- Documentation update

## Context

Resolves #

#### related issues:
- #

#### required by:
- #

#### requires:
- [ ] #

#### conflicts with:
- #

## Checklist:


tests/unit/dataset/dataset_test.py
```diff
@@ -528,6 +603,49 @@ def test_load_correct_raw_gazes(gaze_dataset_configuration):
...
+
+@pytest.mark.parametrize(
+    'gaze_dataset_configuration',
+    ['ToyAOI'],
+    indirect=['gaze_dataset_configuration'],
+)
+@pytest.mark.parametrize(
+    'expected',
+    [
+        EXPECTED_AOI_MULTIPLEYE_STIMULI_TOY_X_1_TEXT_1_1,
```
SiQube, line 878: please add an additional test using the existing aoi file in `tests/files`

tests/unit/dataset/dataset_test.py
```diff
@@ -528,6 +603,49 @@ def test_load_correct_raw_gazes(gaze_dataset_configuration):
         )
 
 
+def test_stimuli_list_exists(gaze_dataset_configuration):
+    dataset = Dataset(**gaze_dataset_configuration['init_kwargs'])
+
+    assert isinstance(dataset.stimuli, list)
```
SiQube, line 847: this could also be an empty list

SiQube (changes requested):
really cool feature, some minor feedback (already)

src/pymovements/dataset/dataset.py
```diff
@@ -71,6 +71,7 @@ def __init__(
         self.events: list[Events] = []
         self.precomputed_events: list[PrecomputedEventDataFrame] = []
         self.precomputed_reading_measures: list[ReadingMeasures] = []
+        self.stimuli: list[TextStimulus] = []
```
dkrako: I think this is fine for this PR, but we should probably create a more general `Stimulus` base class. Moreover, we should think about `Stimulus` collections, as an individual `Stimulus` may be mapped to some trials in the experiment. I will create an issue regarding these.

src/pymovements/dataset/dataset.py
```diff
@@ -131,6 +133,8 @@ def load(
             :py:meth:`pymovements.Dataset.path`.
             This argument is used only for this single call and does not alter
             :py:meth:`pymovements.Dataset.preprocessed_rootpath`. (default: None)
+        stimuli_dirname: str | None
+            :py:meth:`pymovements.Dataset.stimuli_rootpath`. (default: None)
```
dkrako: probably missed to paste 

```
            One-time usage of an alternative directory name to save data relative to
            :py:meth:`pymovements.Dataset.path`.
            This argument is used only for this single call and does not alter
```

src/pymovements/dataset/dataset.py
```diff
@@ -168,6 +172,12 @@ def load(
             for loaded_gaze, loaded_events in zip(self.gaze, self.events):
                 loaded_gaze.events = loaded_events
 
+        # Load text stimuli files if present
+        if self.definition.resources.has_content('stimuli'):
+            self.load_text_stimuli()
+            # stimuli_dirname=stimuli_dirname, # TODO custom dir name
+            # extension=extension,
```
dkrako: passing the extension won't be necessary, as it's more related to the gaze files.

src/pymovements/dataset/dataset_definition.py
```diff
@@ -203,6 +203,28 @@ class DatasetDefinition:
         transformations. If not specified, the constant eye-to-screen distance will be taken from
         the experiment definition. This column will be renamed to ``distance``. (default: None)
 
+    aoi_content_column: str | None
```
dkrako: For this PR I would leave these fields at this level, but they should probably be moved to the `ResourceDefinition` level before the next release. I need to think about how to integrate these best without crowding the `ResourceDefinition` with lots of fields.

Maybe we should create classes like `StimulusResourceDefinition`, `SamplesResourceDefinition`, `EventsResourceDefinition`, `LabelsResourceDefinition`. These would then inherit from `ResourceDefinition` but include the more specific fields associated with these content types.

Alternatively, #1270 paved the way for having `load_kwargs` in the `DatasetDefinition`. Maybe we could use these.

I'll need to think about this a bit more and write up an issue.

src/pymovements/stimulus/text.py
```diff
@@ -150,85 +150,90 @@ def get_aoi(
...
+            width_column: str | None = None,
+            height_column: str | None = None,
+            end_x_column: str | None = None,
+            end_y_column: str | None = None,
+            page_column: str | None = None,
+            custom_read_kwargs: dict[str, Any] | None = None,
+    ) -> TextStimulus:
+        """Load text stimulus from file.
 
-def from_file(
```
dkrako, line 205: I would probably like to deprecate this function instead of removing it to prevent a breaking change.
But I'll do this in a separate PR after merging this. For your just revert the removal and reuse `TextStimulus.from_file()` in the old function.

EDIT: I will merge the separate deprecation PRs before merging this PR here

src/pymovements/dataset/dataset_definition.py
dkrako: I created #1294 to implement the `load_kwargs` field. This way it won't be necessary to adjust neither `DatasetDefinition` nor `ResourceDefinition`.

src/pymovements/dataset/dataset_files.py
```diff
@@ -560,7 +560,7 @@ def load_precomputed_event_file(
     return PrecomputedEventDataFrame(data=precomputed_event_df)
 
 
-def load_text_stimuli_files(
+def load_stimuli_files(
```
dkrako, line 629: I'll create a PR for an `@experimental` property that raises an `ExperimentalWarning` if this function is called

src/pymovements/dataset/dataset.py
dkrako: I changed it to `self.stimuli: list[ImageStimulus | TextStimulus]` for this PR

src/pymovements/dataset/dataset_definition.py
dkrako: as #1294 is merged now, I would remove these explicit fields from `DatasetDefinition`

src/pymovements/__init__.py
```diff
@@ -89,6 +87,9 @@
     'plotting',
     'stimulus',
     'synthetic',
+
+    'exceptions',
+
```
cbueth: Is there a specific reason for moving this, or is it an artifact from previous commits. I'd suggest undoing it.

src/pymovements/dataset/dataset.py
```diff
@@ -431,6 +443,27 @@ def load_event_files(
...
+    def load_stimuli(self) -> None:
+        """Load text stimuli.
+
+        This method checks that the file information for stimuli is available,
+        then loads each text stimulus file listed in ``Dataset.fileinfo['stimuli']`` using
+        the dataset definition and path settings. The resulting list of
+        stimulus objects is assigned to ``Dataset.stimuli``.
+
+        Supported file extensions:
+        - CSV-like: .csv, .tsv, .txt
```
cbueth, line 459:
```suggestion
        Supported file extensions:

        - CSV-like: .csv, .tsv, .txt
```

Needs empty line to be a bullet point list (with on eitem though). See:

https://pymovements--1267.org.readthedocs.build/en/1267/reference/api/pymovements.Dataset.load_stimuli.html

docs/source/reference/stimulus.rst
```diff
@@ -12,14 +12,3 @@ Stimulus
...
-.. rubric:: Input / Output
-
-.. currentmodule:: pymovements.stimulus.text
-
-.. autosummary::
-    :toctree: api
-    :nosignatures:
-    :template: function.rst
-
-    from_file
```
cbueth, line 25: This API Reference section is removes because `from_file` will be deprecated later, but it should not be used anymore, right?

src/pymovements/stimulus/image.py
```diff
@@ -55,6 +55,45 @@ def show(self, stimulus_id: int, origin: str = 'upper') -> None:
...
+        ----------
+        path:  str | Path
+            Path to directory with image stimulus files.
+        filename_pattern:  str
+            Pattern of the image stimulus file names.
+
+        Returns
+        -------
+        ImageStimulus
+            Returns an ImmageStimulus initialized with all matched image stimulus files.
```
cbueth:
```suggestion
            Returns an ImageStimulus initialized with all matched image stimulus files.
```

src/pymovements/stimulus/image.py
```diff
@@ -69,10 +108,7 @@ def from_file(image_path: str | Path) -> ImageStimulus:
...
         Returns the image stimulus file.
     """
-    if isinstance(image_path, str):
-        image_path = Path(image_path)
-
-    return ImageStimulus(images=[image_path])
+    return ImageStimulus.from_file(path=image_path)
 
 
 def from_files(path: str | Path, filename_format: str) -> ImageStimulus:
```
cbueth, line 91: Sometimes there is `filename_pattern` and sometimes `filename_format`. Are these standing for the same thing? If yes, maybe it would be nice to use one of both names everywhere, also in `dataset_test.py` and `dataset_definition.py`.

src/pymovements/stimulus/text.py
```diff
@@ -170,6 +170,88 @@ def get_aoi(
...
+            aoi_column: str,
+            start_x_column: str,
+            start_y_column: str,
+            width_column: str | None = None,
+            height_column: str | None = None,
+            end_x_column: str | None = None,
+            end_y_column: str | None = None,
+            page_column: str | None = None,
+            trial_column: str | None = None,
+            read_csv_kwargs: dict[str, Any] | None = None,
```
cbueth, line 186: Suggestion: renaming `read_csv_kwargs` to `custom_read_kwargs` if thay are the same format and used here like such.

```suggestion
            custom_read_kwargs: dict[str, Any] | None = None,
```

src/pymovements/stimulus/image.py
```diff
@@ -55,6 +55,45 @@ def show(self, stimulus_id: int, origin: str = 'upper') -> None:
...
+        -------
+        ImageStimulus
+            Returns an ImmageStimulus initialized with all matched image stimulus files.
+        """
+        filenames = get_filepaths(path, regex=curly_to_regex(filename_pattern))
+        image_stimuli = []
+        for filename in filenames:
+            image_stimuli.append(filename)
+
+        return ImageStimulus(image_stimuli)
```
cbueth: The iteration is not needed, I suppose this can be done with `list(filenames)` or a list comprehension. Furthermore, we might want to validate file extensions and raise if no files have been found.

```suggestion
        filenames = get_filepaths(path, regex=curly_to_regex(filename_pattern))
        if not filenames:
            raise ValueError(f'No files found matching pattern "{filename_pattern}" in {path}')
        # Validation
        valid_extensions = {'.jpg', '.jpeg', '.png', '.bmp', '.tiff'} # other types?
        for filename in filenames:
            if filename.suffix.lower() not in valid_extensions:
                raise ValueError(f'Unsupported image format: {filename.suffix}')
        return ImageStimulus(list(filenames))
```

src/pymovements/stimulus/image.py
```diff
@@ -55,6 +55,45 @@ def show(self, stimulus_id: int, origin: str = 'upper') -> None:
...
+        ----------
+        path:  str | Path
+            Path to image file to be read.
+
+        Returns
+        -------
+        ImageStimulus
+            Returns an ImageStimulus initialized with the image stimulus file.
+        """
+        return ImageStimulus(images=[Path(path)])
```
cbueth, line 72: I'd suggest file extension validation here, too. Maybe add a CONST at the top of this file for allowed extensions. This exists in TextStimulus already, with the proper error handling.

src/pymovements/stimulus/text.py
```diff
@@ -170,6 +170,88 @@ def get_aoi(
...
+        """
+        if isinstance(path, str):
+            path = Path(path)
+        if read_csv_kwargs is None:
+            read_csv_kwargs = {}
+
+        valid_extensions = {'.csv', '.tsv', '.txt', '.ias'}
+        if path.suffix in valid_extensions:
+            stimulus_df = pl.read_csv(path, **read_csv_kwargs)
+            stimulus_df = stimulus_df.fill_null(' ')
```
cbueth: Checking that the file exists is sufficiently covered by `polars.read_csv`? I was expecting somethink like this:

```suggestion
        if isinstance(path, str):
            path = Path(path)
        if not path.exists():
            raise FileNotFoundError(f'Stimulus file not found: {path}')
        if read_csv_kwargs is None:
            read_csv_kwargs = {}

        valid_extensions = {'.csv', '.tsv', '.txt', '.ias'}
        if path.suffix in valid_extensions:
            stimulus_df = pl.read_csv(path, **read_csv_kwargs)
            stimulus_df = stimulus_df.fill_null(' ')
```
```suggestion
        if isinstance(path, str):
            path = Path(path)
        if read_csv_kwargs is None:
            read_csv_kwargs = {}

        valid_extensions = {'.csv', '.tsv', '.txt', '.ias'}
        if path.suffix in valid_extensions:
            stimulus_df = pl.read_csv(path, **read_csv_kwargs)
            stimulus_df = stimulus_df.fill_null(' ')
```

src/pymovements/dataset/dataset_files.py
```diff
@@ -116,10 +118,13 @@ def scan_dataset(
             resource_dirpath = paths.precomputed_events
         elif content_type == 'precomputed_reading_measures':
             resource_dirpath = paths.precomputed_reading_measures
+        elif content_type == 'stimulus':
```
dkrako: an alternative to `stimulus` could be `TextStimulus`/`ImageStimulus`. This way specifying the load function wouldn't be necessary. I think I might actually prefer it that way.

src/pymovements/dataset/dataset_files.py
```diff
@@ -621,6 +626,71 @@ def load_precomputed_event_file(
     return PrecomputedEventDataFrame(data=precomputed_event_df)
 
 
+def load_stimuli_files(
```
cbueth, line 629: Looking for tests, I did not find this function tested directly alone. Maybe consider adding a dedicated test. (I'd like a test case with the empty list :)

cbueth (changes requested):
Thank's to everyone who worked on this PR, and @dkrako for wrapping it up. There are several changes, well structured, documented and most important tests covered. You can find my comments and suggestions in the separate comments.

I see there is already quite much under the "Future work and comment on warnings" section that is important. Especially having some notebook / examples and this feature in action would be great. Furthermore, I am hesitant to ask for an integration test for this one.

src/pymovements/stimulus/image.py
dkrako: `DatasetDefinition.filename_format` is deprecated in favor of `ResourceDefinition.filename_pattern`. I didn't want to change the signature of the old function but I wanted to use the new naming for the new one.

cbueth: Sure, that's valid.

src/pymovements/__init__.py
dkrako: reverted the change in [`f065a51`](https://github.com/pymovements/pymovements/pull/1267/commits/f065a512df3e48224d5ed93f8a075f5356df6783)

cbueth: thanks!

src/pymovements/stimulus/text.py
dkrako: Using `read_csv_kwargs` is in line with `gaze.io.from_csv()`. `custom_read_kwargs` was deprecated in `DatasetDefinition` as it lead to confusions with `ResourceDefinition.load_kwargs`. `read_csv_kwargs` was chosen to make it more obvious that these kwargs are forwarded to `polars.read_csv()`.

src/pymovements/stimulus/text.py
dkrako: polars already raises a `FileNotFoundError` in case the file does not exist. I added an additional message to provide some more context in [`947195a`](https://github.com/pymovements/pymovements/pull/1267/commits/947195a4a31f3a362bf03e5dbfd569e73d7dd124)

src/pymovements/stimulus/text.py
cbueth: Good choice then!

dkrako: Thanks :)

src/pymovements/dataset/dataset_files.py
dkrako: implemented in [`96506f2`](https://github.com/pymovements/pymovements/pull/1267/commits/96506f2cd0e05a20f51fde6a0130dd6e9bc1fe17)

docs/source/reference/stimulus.rst
dkrako: I reverted that change, should be removed when the function itself is removed.

src/pymovements/stimulus/image.py
dkrako: Thanks for the comment, it made me reconsider some of the code. I changed the following:

- remove `ImageStimulus.from_files()`: currently the `ImageStimulus` class is inconsistent with `TextStimulus` as it may contain a list of stimuli. I think we will rather implement a `Stimuli` class in the future that serves as a collection with keys for each stimuli. These can then be mapped to trials. The multi-image approach in `ImageStimulus` will therefore be deprecated. [`f7b2e27`](https://github.com/pymovements/pymovements/pull/1267/commits/f7b2e2728eba30c3b6d165e0da209161d440e98e)

- remove the filename extension validation from `TextStimulus.from_csv()`: I think this approach is just too restrictive and would require to manually add all the filename extensions (e.g., there might be another csv-like eye tracker format like `ias` which would then require extending the list manually. Further, in case of a binary file that is wrongly named `something.csv` we would get an unhelpful polars error. Instead, I went for the more robust approach of catching the exceptions when loading the files. [`8d2cf51`](https://github.com/pymovements/pymovements/pull/1267/commits/8d2cf511df43019853b2c30718fd324f43043dde)

Further, the `ImageStimulus` will need further rework, because it should actually contain the stimulus data, not just the path to the file. Besides in `ImageStimulus.show()` the actual stimulus data is currently not loaded at all. If the loading would actually happen in `ImageStimulus.from_file()`, then the exceptions could be raised there. Reworking that is more or less out of scope of this PR though (as it requires changes to `ImageStimulus.images`).

src/pymovements/stimulus/image.py
dkrako: instead of filename validation I went for catching the `PIL.UnidentifiedImageError` exception as I think that is a more robust approach  [`f7b2e27`](https://github.com/pymovements/pymovements/pull/1267/commits/f7b2e2728eba30c3b6d165e0da209161d440e98e)

tests/unit/dataset/dataset_test.py
dkrako: added in [`312d82b`](https://github.com/pymovements/pymovements/pull/1267/commits/312d82b944e43835a7d3223ecd2d975565dfb965)

src/pymovements/dataset/dataset_files.py
dkrako: added tests in [`4d66ad4`](https://github.com/pymovements/pymovements/pull/1267/commits/4d66ad461ed9dc3e745304e2b3104a035003a27b)

On a more general note I usually prefer not to explicitly test non-user facing functionality as this often leads to more work when refactoring. As the whole `dataset_files_test.py` does this currently, I added these for now.

Moreover, I think it might be worth refactoring the load functions into a user-facing `DatasetFile.load() -> Gaze | Events | ReadingMeasures | Stimulus` method in the future. The added tests will still be of value then. But that's something we can discuss after v0.30 when `load_gaze_file()` won't depend on `DatasetDefinition` anymore.

src/pymovements/stimulus/image.py
cbueth: >     * remove `ImageStimulus.from_files()`: currently the `ImageStimulus` class is inconsistent with `TextStimulus` as it may contain a list of stimuli. I think we will rather implement a `Stimuli` class in the future that serves as a collection with keys for each stimuli. [...]

A `Stimuli` (plural) /`Stimulus` class sounds reasonable to me if it simplified the usage for the user.

>     * remove the filename extension validation from `TextStimulus.from_csv()`: [...]

Not validating here is a valid choice, the one further down is a good tradeoff.

dkrako (approved):

PR #1287, opened 2025-09-17
feat: add equality comparison methods for `Events` and `Gaze`

dkrako:
## Description

Implement equality comparison methods for `Events` and `Gaze`.

## Implemented changes

- [ ] implement `Events.__eq__()`
- [ ] implement `Gaze.__eq__()`

## How Has This Been Tested?

- [ ] added parametrized test function to `events_test.py`
- [ ] added parametrized test function to `gaze_test.py`

### DIFF
--- a/src/pymovements/events/events.py
+++ b/src/pymovements/events/events.py
@@ -522,6 +522,12 @@ class Events:
         aoi_df = pl.concat(aois)
         self.frame = pl.concat([self.frame, aoi_df], how='horizontal')
 
+    def __eq__(self, other: Events) -> bool:
+        """Check equality between this and another :py:cls:`~pymovements.events.Events` object."""
+        frames_equal = self.frame.equals(other.frame)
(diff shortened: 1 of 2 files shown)


SiQube (changes requested):
can we check float vs int for e.g. `x` in samples? i.e. `x=20` vs `x=20.` should this be equal?

dkrako:
> can we check float vs int for e.g. `x` in samples? i.e. `x=20` vs `x=20.` should this be equal?

So actually I though for polars dataframes, if the type of the column is different, [`equals()`](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.equals.html) returns `False`. But apparently it returns `True` if the values are equal.

I added a test to make this more explicit and ti track changes that might come in the future.

dkrako:
Ah I didn't answer your last question: I think it's fine to have `20.0 == 20`, as this is true for builtin python variables.

PR #1288, opened 2025-09-18
feat!: add `as_dict` argument to `split()` methods

### LINKED ISSUE #1279: `Gaze.split()` drops gaze if there are no events
## Current Behavior

Calling `gaze.split()` when `gaze.events` is empty results in an empty list being returned.

The problem is the use of `zip` here: https://github.com/aeye-lab/pymovements/blob/e61b668afad4c3c88525a17626cea0ac008a10cd/src/pymovements/gaze/gaze.py#L355

Also, if only a subset of trials have no events, the `Gaze` and `Event` objects in the returned list will be misaligned.

## Expected Behavior

`Gaze` should be split and returned even if there are no events. Samples and events should be aligned even if some trials have no events.

### Steps to Reproduce

Example file: [gaze.csv](https://github.com/user-attachments/files/22179116/gaze.csv)

To reproduce empty return value:

```py
import pymovements as pm

gaze = pm.gaze.from_csv(
    "gaze.csv",
    pixel_columns=["pixel_x", "pixel_y"],
    trial_columns=["trial"],
)
gaze.split()  # empty
```

To reproduce misaligned events:

```py
import pymovements as pm
import polars as pl

gaze = pm.gaze.from_csv(
    "gaze.csv",
    pixel_columns=["pixel_x", "pixel_y"],
    trial_columns=["trial"],
)
# Add event for trial2 only
gaze.events.frame.extend(
    pl.DataFrame(
        {
            "trial": ["trial2"],
            "name": ["fixation"],
            "onset": [0],
            "offset": [1000],
            "duration": [1000],
        }
    )
)
gaze.split()  # gaze for trial1 and event for trial2; gaze for trial2 missing
```

## Checklist
 
dkrako:
## :boom: Breaking Change

`Gaze.split()` now raises an exception if the columns specified in `by` are not found in `Gaze.events`.
Previously, empty `events` were returned in `Gaze.split()` leading to potential silent errors.

If you want to ignore `events` during splitting, please call `your_gaze.events = None` beforehand.

## Description

Add a new argument `as_dict` to `Gaze.split()` and `Events.split()`. If `True`, these methods will return a dictionary instead of a list. The keys of the dictionary are tuples of the unique values specified in `by`.

## Example

```python
>>> events = Events(onsets=[20, 30], offsets=[24, 40], name=['blink', 'fixation'], trials=[1, 2])
>>> events.split(as_dict=True)  # splits by self.trial_columns by default
{
 (1,): shape: (1, 5)
  ┌───────┬───────┬───────┬────────┬──────────┐
  │ trial ┆ name  ┆ onset ┆ offset ┆ duration │
  │ ---   ┆ ---   ┆ ---   ┆ ---    ┆ ---      │
  │ i64   ┆ str   ┆ i64   ┆ i64    ┆ i64      │
  ╞═══════╪═══════╪═══════╪════════╪══════════╡
  │ 1     ┆ blink ┆ 20    ┆ 24     ┆ 4        │
  └───────┴───────┴───────┴────────┴──────────┘,
 (2,): shape: (1, 5)
  ┌───────┬──────────┬───────┬────────┬──────────┐
  │ trial ┆ name     ┆ onset ┆ offset ┆ duration │
  │ ---   ┆ ---      ┆ ---   ┆ ---    ┆ ---      │
  │ i64   ┆ str      ┆ i64   ┆ i64    ┆ i64      │
  ╞═══════╪══════════╪═══════╪════════╪══════════╡
  │ 2     ┆ fixation ┆ 30    ┆ 40     ┆ 10       │
  └───────┴──────────┴───────┴────────┴──────────┘
}
```

## How Has This Been Tested?

I tried to test this exhaustively, as the method had already bugfixes and I wanted to make sure to handle all edge cases.

- [x] Test `Events.split()` with one and two trial columns
- [x] Test `Gaze.split()` with samples without events
- [x] Test `Gaze.split()` with samples with experiment
- [x] Test `Gaze.split()` with samples with trial_columns
- [x] Test `Gaze.split()` with samples with events
- [x] Test `Gaze.split()` with samples missing `by` columns
- [x] Test `Gaze.split()` with events missing `by` columns

## Context

This PR was motivated by #1279 to make sure all samples are mapped to the correct events in `Gaze.split()`.

Resolves #1279

#### required by:
- #1221

## Checklist:


tests/unit/events/events_test.py
```diff
@@ -623,27 +623,6 @@ def test_add_trial_column_raises_exception(events, kwargs, exception, message):
     assert message == excinfo.value.args[0]
 
 
-def test_split():
```
dkrako, line 626: this here is just the exact same code as in the test function below. probably a mistake during a merge.

src/pymovements/gaze/gaze.py
```diff
@@ -319,41 +324,54 @@ def split(self, by: Sequence[str] | None = None) -> list[Gaze]:
...
 
         # Convert single string to list for consistent handling
         by = [by] if isinstance(by, str) else by
-        samples_list = self.samples.partition_by(by=by)
 
-        # Check if all columns in 'by' are in events columns
-        events_list = (
-            self.events.split(by)
-            if all(col in self.events.columns for col in by)
-            else [pm.Events()] * len(samples_list)
```
dkrako, line 343: instead of this silent error, now a `polars.exceptions.ColumnNotFoundError` is raised

If a user wants to ignore the events, they can simply call `gaze.events = None` beforehand

tests/unit/gaze/gaze_test.py
```diff
@@ -591,6 +950,26 @@ def test_gaze_dataframe_split_default_no_trial_columns():
         gaze.split()
 
 
+def test_gaze_drop_event_properties(make_gaze_with_events):
```
dkrako, line 1074: moved these tests below as they were between split()  tests

src/pymovements/dataset/dataset.py
```diff
@@ -303,7 +303,8 @@ def split_gaze_data(
         all_fileinfo_rows = []
 
         for frame, fileinfo_row in zip(self.gaze, fileinfo_dicts):
-            split_frames = frame.split(by=by)
+            split_frames = frame.split(by=by, as_dict=False)
+            assert isinstance(split_frames, list)  # or else mypy complains
```
SiQube: below would be more explicit and would prevent assert statement

```diff
diff --git a/src/pymovements/dataset/dataset.py b/src/pymovements/dataset/dataset.py
index ba50231..7c4f662 100644
--- a/src/pymovements/dataset/dataset.py
+++ b/src/pymovements/dataset/dataset.py
@@ -304,7 +304,6 @@ class Dataset:
 
         for frame, fileinfo_row in zip(self.gaze, fileinfo_dicts):
             split_frames = frame.split(by=by, as_dict=False)
-            assert isinstance(split_frames, list)  # or else mypy complains
             all_gaze_frames.extend(split_frames)
             all_fileinfo_rows.extend([fileinfo_row] * len(split_frames))
 
diff --git a/src/pymovements/gaze/gaze.py b/src/pymovements/gaze/gaze.py
index 714089d..8801d09 100644
--- a/src/pymovements/gaze/gaze.py
+++ b/src/pymovements/gaze/gaze.py
@@ -27,6 +27,8 @@ from collections.abc import Sequence
 from copy import deepcopy
 from pathlib import Path
 from typing import Any
+from typing import Literal
+from typing import overload
 
 import numpy as np
 import polars as pl
@@ -309,11 +311,15 @@ class Gaze:
         else:
             raise ValueError(f"unsupported method '{function}'")
 
+    @overload
+    def split(self, by: Sequence[str] | None = None, *, as_dict: Literal[False]) -> list[Gaze]: ...
+    @overload
+    def split(self, by: Sequence[str] | None = None, *, as_dict: Literal[True]) -> dict[tuple[Any, ...], Gaze]: ...
     def split(
             self,
             by: Sequence[str] | None = None,
             *,
-            as_dict: bool = False,
+            as_dict: Literal[True] | Literal[False] = False,
     ) -> list[Gaze] | dict[tuple[Any, ...], Gaze]:
         """Split a single Gaze object into multiple Gaze objects based on specified column(s).
```

src/pymovements/events/events.py
```diff
@@ -423,7 +423,12 @@ def copy(self) -> Events:
         """
         return self.clone()
 
-    def split(self, by: Sequence[str] | None = None) -> list[Events]:
+    def split(
+            self,
+            by: Sequence[str] | None = None,
+            *,
+            as_dict: bool = False,
+    ) -> list[Events] | dict[tuple[Any, ...], Events]:
```
SiQube, line 445: use `overload` and `as_dict: Literal[True] | Literal[False] = False` for typing `split` function

src/pymovements/gaze/gaze.py
```diff
@@ -309,7 +309,12 @@ def apply(
         else:
             raise ValueError(f"unsupported method '{function}'")
 
-    def split(self, by: Sequence[str] | None = None) -> list[Gaze]:
+    def split(
+            self,
+            by: Sequence[str] | None = None,
+            *,
+            as_dict: bool = False,
```
SiQube, line 332: see comment above

src/pymovements/gaze/gaze.py
```diff
@@ -319,41 +324,54 @@ def split(self, by: Sequence[str] | None = None) -> list[Gaze]:
...
-        # Check if all columns in 'by' are in events columns
-        events_list = (
-            self.events.split(by)
-            if all(col in self.events.columns for col in by)
-            else [pm.Events()] * len(samples_list)
-        )
+        if not self.samples.is_empty():
+            # We use as_dict=True here to make sure to map samples to the correct events.
+            grouped_samples = self.samples.partition_by(by=by, as_dict=True)
+            assert isinstance(grouped_samples, dict)  # or else mypy complains below
```
SiQube: see comment above

src/pymovements/gaze/gaze.py
```diff
@@ -319,41 +324,54 @@ def split(self, by: Sequence[str] | None = None) -> list[Gaze]:
...
+        else:
+            grouped_samples = {}
 
-        return [
-            Gaze(
-                samples=samples,
+        if self.events:
+            # We use as_dict=True here to make sure to map events to the correct samples.
+            grouped_events = self.events.split(by=by, as_dict=True)
+            assert isinstance(grouped_events, dict)  # or else mypy complains below
```
SiQube: see above

src/pymovements/gaze/gaze.py
```diff
@@ -319,41 +324,54 @@ def split(self, by: Sequence[str] | None = None) -> list[Gaze]:
...
+            assert isinstance(grouped_samples, dict)  # or else mypy complains below
+        else:
+            grouped_samples = {}
 
-        return [
-            Gaze(
-                samples=samples,
+        if self.events:
+            # We use as_dict=True here to make sure to map events to the correct samples.
+            grouped_events = self.events.split(by=by, as_dict=True)
```
SiQube, line 372: what happens in the case that `by` is in samples but not in events?

tests/unit/events/events_test.py
```diff
@@ -720,10 +699,167 @@ def test_split_default_no_trial_columns():
...
+                        'offset': [1340],
+                    }),
+                ),
+            },
+            id='two_columns',
+        ),
+        pytest.param(
+            Events(
+                pl.from_dict({
+                    'trial_id': [0, 1, 1, 2],
```
SiQube, line 827: what happens if we have `null` values?

SiQube (changes requested):
nice change -- some feedback/questions

src/pymovements/dataset/dataset.py
dkrako: Excellent! I never bothered to fully understand the `@overload` decorator and was mostly confused by its use. This exemplifies it perfectly! I'll apply the changes

dkrako: furtunately mypy understood that `bool` is equivalent to `Literal[True] | Literal[False]` so I didn't need to adjust the base method signature (which would make the docs a bit messy).

dkrako: done

src/pymovements/events/events.py
dkrako: done

src/pymovements/gaze/gaze.py
dkrako: done

src/pymovements/gaze/gaze.py
dkrako: a `polars.exceptions.ColumnNotFoundError` is raised. This is actually why this PR is a breaking change. I think this is the safest approach. If a user wants to ignore the events for splitting, they can simply assign `gaze.events = None` beforehand.

Here's the respective test: 

https://github.com/aeye-lab/pymovements/blob/963c179bf6fe054448d9d97b6cd8e99fe66baedc/tests/unit/gaze/gaze_test.py#L784-L801

dkrako: In the future we should probably add a similar class to `pymovements.exceptions` and enclose these calls with `try: ... except pl.exceptions.ColumnNotFoundError:` but I think this is a bit out of scope of this PR.

src/pymovements/gaze/gaze.py
dkrako: done

src/pymovements/gaze/gaze.py
dkrako: done

tests/unit/events/events_test.py
dkrako: Good catch! added tests in [`e9c69b2`](https://github.com/aeye-lab/pymovements/pull/1288/commits/e9c69b2958176baf924ce3abf20ba4eadd28efda)

Needed to adjust the sorting function to treat `None` values as negative infinity and add `None` as a valid type for trials in `Events.__init__()`.

src/pymovements/events/events.py
```diff
@@ -423,7 +425,21 @@ def copy(self) -> Events:
         """
         return self.clone()
 
-    def split(self, by: Sequence[str] | None = None) -> list[Events]:
+    @overload
+    def split(self, by: Sequence[str] | None = None, *, as_dict: Literal[False]) -> list[Events]:
+        ...
+
```
dkrako, line 433: the newlines between the overloaded signatures are automatically added by `autopep8`

tests/unit/events/events_test.py
dkrako: I additionally added a test where null values appear in a trial column of type string and as expected the sorting function didn't work again. I removed the sorting in [`853704d`](https://github.com/aeye-lab/pymovements/pull/1288/commits/853704d9601b64a9fd2e6c43d63de00d518ba209) as this just isn't worth the complexity.

dkrako:
somehow two tests are failing now

dkrako:
Finally. This took a while, but now the splits are sorted correctly and helpful error messages are raised in case of weird column dtypes.

src/pymovements/events/events.py
SiQube: because of line length

SiQube (approved):

PR #1295, opened 2025-09-25
feat: set screen axes and aspect ratio in plotting functions

### LINKED ISSUE #1230: automatically set axis dimension in pixel-data plots

shaitarAn:
## Description

This PR introduces a shared helper function `_set_screen_axes()` to consistently set plot axes based on screen dimensions defined in `gaze.experiment.screen`.

The change ensures that `traceplot`, `scanpathplot`, and `heatmap` correctly:

- Set xlim and ylim according to the screen resolution
- Flip the y-axis for `origin='upper left'` (to match screen coordinates)
- Enforce equal aspect ratio (square pixels)

## Implemented changes

In plotting/

- [x] Added a helper function `_set_screen_axes()` to `_matplotlib.py`
- [x]  Implemented `_set_screen_axes()` in `traceplot.py`, `scanpathplot.py`, and `heatmap.py`.
- [x] Added tests to `traceplot_test.py`, `scanpathplot_test.py`, and `heatmap_test.py`

## How Has This Been Tested?

- [x] unit tests for axes limits, aspect ratio, invalid origin ValueError, and Screen.width and/or Screen.height being None
- [x] All plotting-related unit tests pass.
- [x] Verified visually with `ToyDataset`

## Type of change

Remove irrelevant items:
- Bug fix
- New functionality

## Context

Resolves #1230 

#### related issues:
- #1247 

## Checklist:

### DIFF
--- a/tests/fixtures/file_fixtures.py
+++ b/tests/fixtures/file_fixtures.py
@@ -18,6 +18,8 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 """Provide fixtures to securely make files from examples in ``tests/files``."""
+from __future__ import annotations
+
 import shutil
 from collections.abc import Callable
 from pathlib import Path
@@ -56,3 +58,55 @@ def fixture_make_example_file(testfiles_dirpath: Path, tmp_path: Path) -> Callab
         shutil.copy2(source_filepath, target_filepath)
         return target_filepath
     return _make_example_file
+
+
+@pytest.fixture(name='make_text_file', scope='function')
+def fixture_make_text_file(tmp_path: Path) -> Callable[[str | Path, str, str, str], Path]:
+    """Make a custom file with self-written header and body.
+
+    Parameters
+    ----------
+    tmp_path : Path
+        Temporary directory where files are copied to. Existing fixture.
+
+    Returns
+    -------
+    Callable[[str | Path, str, str, str], Path]
+        Function that takes a filename, the file header and body and the encoding of the file.
+        Returns the path to the created file.
+        Header and body are simply concatenated into a single string and written to the file.
+        The file is saved into a temporary directory.
+
+    """
+    def _ensure_rel_filename(filename: str | Path) -> Path:
+        """Ensure filename is a relative path.
+
+        Accepts str or Path. Rejects other types with TypeError.
+        Rejects absolute paths, drive letters (Windows),
+        and user home shortcuts '~' with ValueError.
+        Returns a Path relative to current directory (to be joined with tmp_path).
+        """
+        if not isinstance(filename, (str, Path)):
+            raise TypeError(f"filename must be a str or Path, got {type(filename).__name__}")
+
+        if isinstance(filename, str) and filename.startswith('~'):
+            raise ValueError("filename must be a relative path; '~' (home) is not allowed")
+
+        p = Path(filename)
+        # On Windows, p.drive captures drive letters; p.anchor is non-empty for absolute paths
+        if p.is_absolute() or p.anchor or getattr(p, 'drive', ''):
+            raise ValueError('filename must be a relative path without drive or root')
+        return p
+
+    def _make_text_file(
+            filename: str | Path, header: str = '', body: str = '\n', encoding: str = 'utf-8',
+    ) -> Path:
+        rel = _ensure_rel_filename(filename)
+        content = header + body
+        filepath = tmp_path / rel
+        # assure parent directory exists
+        filepath.parent.mkdir(parents=True, exist_ok=True)
+        # write contents to a file
+        filepath.write_text(content, encoding=encoding)
+        return filepath
+    return _make_text_file
--- a/tests/fixtures/file_fixtures_test.py
+++ b/tests/fixtures/file_fixtures_test.py
@@ -19,6 +19,7 @@
 # SOFTWARE.
 """Test file fixtures."""
 import filecmp
+from pathlib import Path
 
 import pytest
 
@@ -43,3 +44,59 @@ def test_make_example_file_returns_copy(filename, make_example_file, testfiles_d
 
     assert fixture_filepath != testfiles_filepath  # different filepath
     assert filecmp.cmp(fixture_filepath, testfiles_filepath)  # same content
+
+
+def test_make_text_file_accepts_relative_path_object(make_text_file):
+    p = make_text_file(Path('nested') / 'custom.txt', header='H', body='B')
+    assert p.name == 'custom.txt'
+    assert p.parent.name == 'nested'
+    print(f"dir in parent: {p.parent}")
+    assert p.exists()
+    assert p.read_text(encoding='utf-8') == 'HB'
+
+
+def test_make_text_file_rejects_non_pathlike(make_text_file):
+    with pytest.raises(TypeError, match='filename must be a str or Path, got int'):
+        make_text_file(123, header='h', body='b')
+
+
+def test_make_text_file_writes_concatenated_content(make_text_file):
+    header = 'HEADER LINE\nSECOND\n'
+    body = 'BODY LINE\nEND'
+    p = make_text_file('custom.txt', header=header, body=body)
+    assert p.name == 'custom.txt'
+    assert p.exists()
+    assert p.read_text(encoding='utf-8') == header + body
+
+
+def test_make_text_file_defaults(make_text_file):
+    p = make_text_file('default.txt')
+    assert p.exists()
+    # default header is '' and default body is '\n'
+    assert p.read_text(encoding='utf-8') == '\n'
+
+
+def test_make_text_file_non_utf8_encoding(make_text_file):
+    header = 'über header\n(seit 2017 gibt es ein großgeschriebenes ß: NĀMLICH ẞ 🤷\n'
+    body = 'Le type naïf de la fête de l\'été'
+    p = make_text_file('utf16.txt', header=header, body=body, encoding='utf-16')
+    assert p.read_text(encoding='utf-16') == header + body
+
+
+def test_make_text_file_overwrites_existing(make_text_file):
+    p1 = make_text_file('same.txt', header='A', body='B')
+    p2 = make_text_file('same.txt', header='C', body='D')
+    assert p2 == p1
+    assert p1.read_text(encoding='utf-8') == 'CD'
+
+
+def test_make_text_file_rejects_absolute_paths(make_text_file):
+    with pytest.raises(ValueError, match='relative path'):
+        make_text_file(Path('/absolute.txt'), header='h', body='b')
+    with pytest.raises(ValueError, match='relative path'):
+        make_text_file('/absolute.txt', header='h', body='b')
+
+
+def test_make_text_file_rejects_tilde_home(make_text_file):
+    with pytest.raises(ValueError, match=r"~\' \(home\) is not allowed|relative path"):
+        make_text_file('~/.secret.txt', header='h', body='b')
--- a/tests/unit/dataset/_utils/_archives_test.py
+++ b/tests/unit/dataset/_utils/_archives_test.py
@@ -78,13 +78,11 @@ def test_detect_file_type_no_suffixes():
         'xz_compressed_zip_archive',
     ],
 )
-def fixture_archive(request, tmp_path):
-    rootpath = tmp_path
+def fixture_archive(request, make_text_file):
     compression, extension = request.param
 
-    # write tmp filepath
-    test_filepath = rootpath / 'test.file'
-    test_filepath.write_text('test')
+    test_filepath = make_text_file(filename='test.file', body='test')
+    rootpath = test_filepath.parent
 
     single_child_directory = 'singlechild'
     top_level_directory = 'toplevel'
@@ -195,13 +193,12 @@ def fixture_compressed_file(request, tmp_path):
         'xz_compressed_unsupported_archive',
     ],
 )
-def fixture_unsupported_archive(request, tmp_path):
-    rootpath = tmp_path
+def fixture_unsupported_archive(request, make_text_file):
     compression, extension = request.param
 
-    # write tmp filepath
-    filepath = rootpath / 'test.file'
-    filepath.write_text('test')
+    filepath = make_text_file(filename='test.file', body='test')
+    rootpath = filepath.parent
+
     archive_path = rootpath / f'test.{extension}.{compression}'
     comp_type = _ZIP_COMPRESSION_MAP[f'.{compression}']
     with zipfile.ZipFile(archive_path, 'w', compression=comp_type) as zip_open:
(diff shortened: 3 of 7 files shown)


src/pymovements/plotting/traceplot.py
```diff
@@ -153,6 +140,8 @@ def traceplot(
         ax=ax,
     )
 
+    _apply_screen_axes(ax, gaze, func_name='traceplot')
```
dkrako: this would be placed better after `_draw_line_data()`, because that function might change the axes again

src/pymovements/plotting/_matplotlib.py
```diff
@@ -314,3 +313,41 @@ def _draw_line_data(
     line_collection.set_linewidth(2)
     line = ax.add_collection(line_collection)
     return line
+
+
+def _apply_screen_axes(
```
dkrako: rename this to `_set_screen_axes()`

src/pymovements/plotting/_matplotlib.py
```diff
@@ -314,3 +313,41 @@ def _draw_line_data(
...
+    ax: plt.Axes,
+    gaze: object | None,
```
dkrako: instead of `gaze`, use `screen: Screen` as the parameter. This makes it more obvious what the function does.

You would then call the function like this:

```python
if gaze is not None and gaze.experiment is not None:
    _set_screen_axes(ax, gaze.experiment.screen, func_name='...')
```

dkrako (changes requested):
Great! Thanks for the addition. I just have some minor change requests.

Additionally, can you also add this to `heatmap`?

src/pymovements/plotting/heatmap.py
```diff
@@ -165,14 +168,15 @@ def heatmap(
     # Convert heatmap values from sample count to seconds
     heatmap_value /= gaze.experiment.sampling_rate
 
-    if origin == 'upper':
-        extent = [x_edges[0], x_edges[-1], y_edges[0], y_edges[-1]]
-    else:
-        extent = [x_edges[0], x_edges[-1], y_edges[-1], y_edges[0]]
+    extent = [x_edges[0], x_edges[-1], y_edges[0], y_edges[-1]]
```
dkrako: I'm a bit confused why you changed this? #1301 already updated the default value, there's no need in changing this

src/pymovements/plotting/heatmap.py
```diff
@@ -165,14 +168,15 @@ def heatmap(
...
 
     # If add_stimulus is requested, we still reuse/create fig/ax via prepare_figure and then draw
     fig, ax, own_figure = prepare_figure(ax, figsize, func_name='heatmap')
 
+    # # Apply screen-based axis limits and aspect ratio
+    if gaze is not None and gaze.experiment is not None:  # pragma: no cover
```
dkrako: instead of using `#pragma: no cover` please write a test that covers both the `True` and `False` case.

src/pymovements/plotting/scanpathplot.py
```diff
@@ -178,6 +179,9 @@ def scanpathplot(
         ax=ax,
     )
 
+    if gaze is not None and gaze.experiment is not None:
+        _set_screen_axes(ax, gaze.experiment.screen, func_name='scanpathplot')
```
dkrako: please move this further below, shortly before `ax.set_title()`.

setting the screen axes should be done **after** any plots are made (e.g. `Circle`, `_draw_line_data`). Otherwise these new plot calls may overwrite what is set in `_set_screen_axes`.

src/pymovements/plotting/heatmap.py
```diff
@@ -120,6 +121,8 @@ def heatmap(
     ValueError
         If the experiment property of the Gaze is None
     """
+    _ = origin  # silence unused-argument warning
```
dkrako: why? this makes the argument nonfunctional

src/pymovements/plotting/conftest.py
```diff
@@ -0,0 +1,36 @@
...
+from __future__ import annotations
+
+from collections.abc import Generator
+
+import matplotlib.pyplot as plt
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def close_figures_after_test() -> Generator[None]:  # pragma: no cover
```
dkrako: can you remove the `# pragma: no cover` here?

tests/unit/plotting/conftest.py
dkrako, line 1: this file really shouldn't be included in the `src` directory as `pytest` is not a user dependency

src/pymovements/plotting/_matplotlib.py
```diff
@@ -314,3 +315,49 @@ def _draw_line_data(
...
+    if screen.origin != 'upper left':
+        raise ValueError(
+            f'{func_name}: screen origin must be "upper left", got "{screen.origin}".',
+        )
+
+    if (
+        screen.width_px is None or screen.height_px is None
+        or screen.width_px <= 0 or screen.height_px <= 0
+    ):
+        raise ValueError(
```
dkrako: this would raise a `ValueError` if `screen == Screen()` even though `screen is None` doesn't result in an exception. if any of them is `None` you should probably just `return` without setting axes.

src/pymovements/plotting/heatmap.py
shaitarAn: i think it's better to omit the condition entirely because the branch gaze=None is unreachable due to the earlier check:
    if not gaze.experiment:
        raise ValueError(
            'Experiment property of Gaze is None. '
            'Gaze must be associated with an experiment.',
        )
   assert gaze.experiment.sampling_rate is not None

src/pymovements/plotting/heatmap.py
shaitarAn: #1301 only resolved half of the issue. The heatmap was still flipped. I kept the wrong version of extent. It should be extent = [x_edges[0], x_edges[-1], y_edges[-1], y_edges[0]]. I'm changing that.

tests/unit/plotting/conftest.py
```diff
@@ -0,0 +1,36 @@
...
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+"""Pytest configuration for plotting tests.
+
+Defines global fixtures to ensure Matplotlib figures are closed after each test.
+"""
```
dkrako: having this file would be quite beneficial but it would probably be best to create a separate PR for this.

we can there further discuss how to handle existing test cases: should we remove the `closefig` everywhere or are there tests that should also test if a figure was closed if desired?

The testing of the plotting functions is currently in a very poor state so having this fixture would be great. The work is probably just out of scope of this PR and valids a proper PR.

Also, please move the file to `tests/unit/plotting`, as this should not be part of the package source.

src/pymovements/plotting/heatmap.py
dkrako: you're right, didn't see that the function is that restrictive.

tests/unit/plotting/traceplot_test.py
```diff
@@ -202,26 +202,3 @@ def test_traceplot_exceptions(gaze, kwargs, exception, monkeypatch):
 def test_traceplot_no_experiment(gaze_no_exp):
     # Should not raise any exception
     pm.plotting.traceplot(gaze_no_exp, show=False)
-
-
-@pytest.mark.parametrize(
```
dkrako, line 207: these tests are removed now, this is probably a mistake right?

src/pymovements/plotting/heatmap.py
dkrako: I'm really unsure if this is correct. The extent is needed for loading the image correctly.

Can you provide plots that this works as expected? 

these four plots are needed to assess this correctly: 
- `origin='upper'`, `stimulus_origin='upper`
- `origin='upper'`, `stimulus_origin='lower`
- `origin='lower'`, `stimulus_origin='upper`
- `origin='lower'`, `stimulus_origin='lower`

tests/unit/plotting/traceplot_test.py
```diff
@@ -202,26 +202,3 @@ def test_traceplot_exceptions(gaze, kwargs, exception, monkeypatch):
...
-        (1280, None),   # height None
-        (None, None),   # both None
-    ],
-)
-def test_traceplot_screen_dims(gaze, width, height):
-    # Set screen dimensions
-    gaze.experiment.screen.width_px = width
-    gaze.experiment.screen.height_px = height
-
-    pm.plotting.traceplot(gaze=gaze, show=False)
```
dkrako, line 221: Add a check here that the axes limits are set correctly.

Then please add this kind of test also to `heatmap_test.py` and `scanpathplot_test.py`

tests/unit/plotting/figure_utils_test.py
```diff
@@ -156,3 +158,67 @@ def test_prepare_figure_external_ax_figsize_none_no_warning():
...
+@pytest.mark.parametrize('origin', ['lower left', 'center'])
+def test_set_screen_axes_invalid_origin(origin):
+    fig, ax = plt.subplots()
+    screen = Screen(width_px=800, height_px=600, origin=origin)
+    with pytest.raises(ValueError, match='screen origin must be "upper left"'):
+        _set_screen_axes(ax, screen, func_name='dummyplot')
+    plt.close(fig)
+
+
+class DummyScreen(Screen):
```
dkrako: I don't really understand why this class is necessary. Couldn't you just use `Screen` in the tests below?

dkrako (changes requested):

src/pymovements/plotting/heatmap.py
dkrako: It would probably be best to keep these changes out of this PR and create a separate one and add these plots to the description to visualize what the effect of these changes are.

tests/unit/plotting/figure_utils_test.py
shaitarAn: I used `DummyScreen` to bypass the validation in Screen.__post_init__, so I could test how _set_screen_axes() handles invalid or incomplete screen dimensions without the constructor raising errors first. Maybe it's unnecessary if all `Screen` instances reaching hte plotting stage are always valid.

dkrako: Ah I see now further down below there's the test for negative values. You're right that the check in `_set_screen_axes()` for negative screen values isn't really necessary then, as we can safely assume that the `Screen` is always valid.

Anyway, always good to think about the edge cases. :)

src/pymovements/plotting/_matplotlib.py
```diff
@@ -314,3 +315,53 @@ def _draw_line_data(
...
+
+    # If screen has no pixel info, skip silently
+    if screen.width_px is None or screen.height_px is None:
+        return
+
+    if (
+        screen.width_px is None or screen.height_px is None
+        or screen.width_px <= 0 or screen.height_px <= 0
+    ):
+        raise ValueError(
```
dkrako: as discussed, this value error is not necessary as we can assume the `Screen` to be valid.

tests/unit/plotting/scanpathplot_test.py
```diff
@@ -236,3 +236,59 @@ def test_scanpathplot_events_is_deprecated(gaze):
...
+
+def test_set_screen_axes_valid(gaze):
+    fig, ax = scanpathplot(
+        gaze=gaze,
+        show=False,
+    )
+    assert ax.get_xlim() == (0, gaze.experiment.screen.width_px)
+    assert ax.get_ylim() == (gaze.experiment.screen.height_px, 0)
+    assert ax.get_aspect() == 1
+    plt.close(fig)
```
dkrako:
```suggestion
```

since #1309 was merged you don't need to manually close the figure anymore.

tests/unit/plotting/scanpathplot_test.py
```diff
@@ -236,3 +236,59 @@ def test_scanpathplot_events_is_deprecated(gaze):
...
+
+    # Axes limits should be finite numbers, not NaN/None
+    xlim, ylim = ax.get_xlim(), ax.get_ylim()
+    assert np.isfinite(xlim).all()
+    assert np.isfinite(ylim).all()
+
+    # Aspect ratio should not be 'equal' (not forced by _set_screen_axes)
+    assert ax.get_aspect() != 'equal'
+
+    plt.close(fig)
```
dkrako:
```suggestion
```

tests/unit/plotting/traceplot_test.py
```diff
@@ -204,24 +204,49 @@ def test_traceplot_no_experiment(gaze_no_exp):
...
-        pm.plotting.traceplot(gaze=gaze, show=False)
+    # Axes limits should be finite numbers, not NaN/None
+    xlim, ylim = ax.get_xlim(), ax.get_ylim()
+    assert np.isfinite(xlim).all()
+    assert np.isfinite(ylim).all()
+
+    # Aspect ratio should not be 'equal' (not forced by _set_screen_axes)
+    assert ax.get_aspect() != 'equal'
+
+    plt.close(fig)
```
dkrako:
```suggestion
```

tests/unit/plotting/traceplot_test.py
```diff
@@ -204,24 +204,49 @@ def test_traceplot_no_experiment(gaze_no_exp):
...
 
+def test_set_screen_axes_valid(gaze):
+    fig, ax = pm.plotting.traceplot(
+        gaze=gaze,
+        show=False,
+    )
+    assert ax.get_xlim() == (0, gaze.experiment.screen.width_px)
+    assert ax.get_ylim() == (gaze.experiment.screen.height_px, 0)
+    assert ax.get_aspect() == 1
+    plt.close(fig)
```
dkrako:
```suggestion
```

tests/unit/plotting/traceplot_test.py
```diff
@@ -204,24 +204,49 @@ def test_traceplot_no_experiment(gaze_no_exp):
...
 )
-def test_traceplot_screen_dims(gaze, width, height):
-    # Set screen dimensions
+def test_set_screen_axes_none_dimensions_returns(width, height, gaze):
+    """Should not raise or override axes when screen dimensions are None."""
     gaze.experiment.screen.width_px = width
     gaze.experiment.screen.height_px = height
 
-    pm.plotting.traceplot(gaze=gaze, show=False)
+    fig, ax = plt.subplots()
```
dkrako:
```suggestion
    assert ax.get_aspect() != 'equal'
```

As you check the aspect of the axis later on, you should also check its value before calling the function.

PR #1298, opened 2025-10-01
Feat/dataset events property

### LINKED ISSUE #1110: implement `Dataset.events` as a property
## Description of the problem
Currently `Datsaset.events` is simply implemented as a list of `EventDataFrame`.

This leads to headaches when having to keep `Dataset.gaze` and `Dataset.events` consistent.

## Description of a solution

`Dataset.events` should be implemented as a property that returns a list of `GazeDataFrame.events` (from `Dataset.gaze`).
This way it is guaranteed that 

- [ ] pass events to `GazeDataFrame` on loading files
- [ ] implement getter property for `Dataset.events`
- [ ] implement setter property for `Dataset.events`
- [ ] This for loop introduced in #922 won't be necessary anymore and should be removed accordingly:
https://github.com/aeye-lab/pymovements/blob/dc99209e774155bd148dafcee159fac632dd583f/src/pymovements/dataset/dataset.py#L161-L162

## Context
 
- This comment in [#921](https://github.com/aeye-lab/pymovements/issues/921#issuecomment-2652323952) proposed getting rid of `EventDataFrame` completely

shaitarAn:
## Description

This PR refactors events as a property of Dataset keeping it in sync with gaze.events. Additionally, tests that assumed independence between Dataset.events and gaze.events have been adapted to reflect the new consistent behavior.

## Implemented changes

- [x] Refactored Dataset.events to be a property synchronized with gaze.events.
- [x] Implemented getter and setter property for Dataset.events
- [x] Updated clear_events() to reset events directly on each Gaze object instead of overwriting the dataset’s event list.
- [x] Updated dataset_test.py::test_clear_events to align with the new design, adding dummy gazes when needed for synchronization.

## How Has This Been Tested?

- [x] Verified that modified test_clear_events now passes under the new setup.
- [ ] Confirmed that other dataset-related tests run successfully.

## Type of change

Remove irrelevant items:
- New functionality
- Breaking change (potentially, tests and code relying on old Dataset.events behavior may need adjustment).

## Context

Resolves #1110

## Checklist:


src/pymovements/dataset/dataset.py
```diff
@@ -165,11 +164,34 @@ def load(
...
-            for loaded_gaze, loaded_events in zip(self.gaze, self.events):
-                loaded_gaze.events = loaded_events
 
         return self
 
+    @property
+    def events(self) -> list[Events]:
+        """Return events for all Gaze objects in the Dataset.
+
+        This harmonizes Dataset.events with Gaze.events
```
dkrako: harmonizes is probably a bit vague. Instead, describe that it is a mapping between `Dataset.gaze` and how they correspond.

Idea: This maps bidirectionally to `Dataset.gaze`, with `Dataset.events[i] is Dataset.gaze[i].events`.

Probably the wording can be improved but you'll get the gist.

src/pymovements/dataset/dataset.py
```diff
@@ -165,11 +164,34 @@ def load(
...
+        """
+        return [gaze.events for gaze in self.gaze]
+
+    @events.setter
+    def events(self, events_list: list[Events]) -> None:
+        """Assign events to each Gaze object in the Dataset.
+
+        Parameters
```
dkrako, line 196: here you could add something similar as proposed above

src/pymovements/dataset/dataset.py
```diff
@@ -165,11 +164,34 @@ def load(
...
+        ----------
+        events_list : list[Events]
+            Must have the same length as Dataset.gaze.
```
dkrako: add a section like

```rst
Raises
------
ValueError
    If length of input `events_list` does not match the length of `Dataset.gaze`.
```

src/pymovements/dataset/dataset.py
```diff
@@ -165,11 +164,34 @@ def load(
                 events_dirname=events_dirname,
                 extension=extension,
             )
-            for loaded_gaze, loaded_events in zip(self.gaze, self.events):
```
dkrako, line 168: great that we can now get rid of this

src/pymovements/dataset/dataset.py
```diff
@@ -886,8 +906,8 @@ def clear_events(self) -> Dataset:
         if len(self.events) == 0:
             return self
 
-        for file_id, _ in enumerate(self.events):
-            self.events[file_id] = Events()
+        for gaze in self.gaze:
+            gaze.events = Events()
```
dkrako, line 937: Nice!

tests/unit/dataset/dataset_test.py
```diff
@@ -1185,21 +1188,33 @@ def test_detect_events_raises_column_not_found_error(
                 Events(name='event', onsets=[0], offsets=[99]),
             ],
             [Events(), Events()],
+            2,
             id='two_instance_filled_df_gets_cleared_to_two_empty_dfs',
         ),
     ],
 )
-def test_clear_events(events_init, events_expected, tmp_path):
+def test_clear_events(events_init, events_expected, tmp_path, num_gazes):
```
dkrako: instead of adding `num_gazes` as a parameter, you can just use `num_gazes = len(events_expected)` in the function body

tests/unit/dataset/dataset_test.py
```diff
@@ -2144,3 +2159,35 @@ def test_drop_event_property(gaze_dataset_configuration):
...
+            pixel_columns=['pixel_x', 'pixel_y'],
+            time_column='time',
+            time_unit='ms',
+        ),
+    )
+
+    assert isinstance(dataset.events, list)
+    assert len(dataset.events) == 1
+    assert isinstance(dataset.events[0], Events)
+    assert dataset.events[0].frame.is_empty()
```
dkrako: also add the assert `assert dataset.events[0] is dataset.gaze[0].events`

tests/unit/dataset/dataset_test.py
```diff
@@ -2144,3 +2159,35 @@ def test_drop_event_property(gaze_dataset_configuration):
...
+        Gaze(
+            pl.DataFrame({'time': [], 'pixel_x': [], 'pixel_y': []}),
+            pixel_columns=['pixel_x', 'pixel_y'],
+            time_column='time',
+            time_unit='ms',
+        ),
+    )
+
+    assert isinstance(dataset.events, list)
+    assert len(dataset.events) == 1
```
dkrako: also add a test with `n=3`

dkrako (changes requested):
clicked the wrong button

shaitarAn:
> Great that you are taking care of this issue! There are a few minor change requests I would like to make.
> 
> Apart from this, probably a few more tests would be great that cover the setter.
> 
> Please add the following tests:
> 
> * [ ]  assign an `Events` object to `Dataset.gaze[1].events` and check that only `Dataset.events[1]` refers to this object, not the others in the list (with `len(Dataset.gaze= == 3`).
> * [ ]  vice versa, assign to `Dataset.events[1]` and check `Dataset.gaze[1].events`

Should these tests assert object identity (is) between dataset.events[i] and dataset.gaze[i].events, or is it sufficient to check equality (==)? Currently, the identity check does not work because the getter returns a list. Should i change the getter behaviour or only check for equality in the test?

shaitarAn:
> > Great that you are taking care of this issue! There are a few minor change requests I would like to make.
> > Apart from this, probably a few more tests would be great that cover the setter.
> > Please add the following tests:
> > 
> > * [ ]   assign an `Events` object to `Dataset.gaze[1].events` and check that only `Dataset.events[1]` refers to this object, not the others in the list (with `len(Dataset.gaze= == 3`).
> > * [ ]   vice versa, assign to `Dataset.events[1]` and check `Dataset.gaze[1].events`
> 
> Currently, assigning a new Events object to Dataset.events[1] and then checking the equality with Dataset.gaze[1].events does not work because the Dataset.events getter returns a list of gaze.events. Should i change the getter behaviour?

so yeah, I guess while the new fix maps dataset.events[i] to dataset.gaze[i].events, they are not bi-directionally in sync

tests/unit/dataset/dataset_test.py
shaitarAn: done

src/pymovements/dataset/dataset.py
shaitarAn: done

src/pymovements/dataset/dataset.py
shaitarAn: As you can see from my comments bellow, I realized that this mapping is not truly bidirectional. Assigning a new Events object to an element of Dataset.events does not update the corresponding gaze.events object.

tests/unit/dataset/dataset_test.py
shaitarAn: done

tests/unit/dataset/dataset_test.py
shaitarAn: done

src/pymovements/dataset/dataset.py
dkrako: Ah I see. So this should work:

```python
new_events = Events()
dataset.gaze[1].events = new_events
assert dataset.events[1] is dataset.gaze[1].events
```

but this won't:
```python
new_events = Events()
dataset.events[1] = new_events
assert dataset.events[1] is dataset.gaze[1].events
```

The problem here is, `dataset.events[1] = new_events` first calls the getter which returns the list of events, so we don't have the means to warn the user that this operation doesn't do anything.

Now I see two options to handle this:

1. return an immutable tuple of events ( `def events(self) -> tuple[Events, ...]:`)  instead of `list[Events]`.
    This way `dataset.events[1] = Events()` would return a `TypeError: 'tuple' object does not support item assignment`.
    Calling methods of an indivudual events object would then be reflected on both sides. e.g., `dataset.events[1].drop('column')` would lead to the correct result:
     `assert 'column'  not in dataset.gaze[1].events.columns`
2. return a custom class that functions as a true bi-directional mapping. basically you would create a subclass of `list`, and override the assignment operator and possible other updating functions. This would be really complex.

I think implementing this custom class is not worth the complexity, so I would be in favor to just return a tuple instead of a list.

tests/unit/dataset/dataset_test.py
```diff
@@ -2144,3 +2157,70 @@ def test_drop_event_property(gaze_dataset_configuration):
...
+    dataset.events[1] = ev
+    assert dataset.gaze[0].events is not ev
+    assert dataset.events[0] is not ev
+    assert dataset.gaze[2].events is not ev
+    assert dataset.events[2] is not ev
+    # these do not work because of how getter/setter is implemented
+    # assert dataset.gaze[1].events is ev
+    # assert dataset.events[1] is ev
+
+    ev2 = Events(pl.DataFrame({'onset': [2], 'offset': [3], 'name': ['saccade']}))
```
dkrako: this is probably worth a separate test case.
it tests if changes in `dataset.gaze[i]` are also reflected in `dataset.events[i]`

tests/unit/dataset/dataset_test.py
```diff
@@ -2144,3 +2157,70 @@ def test_drop_event_property(gaze_dataset_configuration):
...
+            ),
+        )
+
+    ev = Events(pl.DataFrame({'onset': [0], 'offset': [1], 'name': ['fixation']}))
+    dataset.events[1] = ev
+    assert dataset.gaze[0].events is not ev
+    assert dataset.events[0] is not ev
+    assert dataset.gaze[2].events is not ev
+    assert dataset.events[2] is not ev
+    # these do not work because of how getter/setter is implemented
```
dkrako: yes, the issue here is, that `dataset.events[1] = ev` actually calls the getter of `dataset.events` and then calls the setter on the returned list.

to test the setter, you would need to do something like in `test_events_setter_raises_on_length_mismatch`, just with matching number of events (`n=3`). then you can check for each entry `assert dataset.gaze[i].events is dataset.events[i]`

dkrako (changes requested):
I think it's best to try to implement this as a tuple for now.
It might be that some tests will break, I hope not but I'm not 100% sure.

src/pymovements/dataset/dataset.py
```diff
@@ -773,7 +812,8 @@ def detect(
                 df=gaze.events.frame,
                 fileinfo=fileinfo_row,
             )
-            self.events[file_id] = gaze.events
+            self.gaze[file_id].events = gaze.events
```
dkrako: this line can be just removed now as it doesn't do anything

src/pymovements/dataset/dataset.py
```diff
@@ -963,7 +1003,7 @@ def save_events(
             If extension is not in list of valid extensions.
         """
         dataset_files.save_events(
-            events=self.events,
+            events=list(self.events),
```
dkrako: instead of casting to `list` please change the signature of the `dataset_files` helper function to

```python
def save_events(
        events: Sequence[Events],
        fileinfo: pl.DataFrame,
        paths: DatasetPaths,
        events_dirname: str | None = None,
        verbose: int = 1,
        extension: str = 'feather',
) -> None:
```

dkrako (approved):
Thanks a lot for your initiative and handling all the change requests!

I updated some parts of the docstring to my liking and now it's ready to merge.

PR #1299, opened 2025-10-01
depr: deprecate `**kwargs` in `gaze.io` functions in favor of explicit arguments

dkrako:
## Description

Please include a clear and concise description of the change and which issue is fixed.
Please also include excepted improvements, relevant motivation and context.
List any dependencies that are required for this change.

## Implemented changes

Insert a description of the changes implemented in the pull request.

- [ ] Change 1
- [ ] Change 2
- [ ] Change 3

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes.
Provide instructions so we can reproduce.
Please also list any relevant details for your test configuration

- [ ] Test A
- [ ] Test B

## Type of change

Remove irrelevant items:
- Bug fix
- New functionality
- Performance improvement
- New dataset
- Breaking change
- Documentation update

## Context

Resolves #

#### related issues:
- #

#### required by:
- #

#### requires:
- [ ] #

#### conflicts with:
- #

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] I have checked my code and corrected any misspellings

PR #1297, opened 2025-09-30
test: add fixture to make example files

dkrako:
## Description

Please include a clear and concise description of the change and which issue is fixed.
Please also include excepted improvements, relevant motivation and context.
List any dependencies that are required for this change.

## Implemented changes

Insert a description of the changes implemented in the pull request.

- [ ] Change 1
- [ ] Change 2
- [ ] Change 3

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes.
Provide instructions so we can reproduce.
Please also list any relevant details for your test configuration

- [ ] Test A
- [ ] Test B

## Type of change

Remove irrelevant items:
- Bug fix
- New functionality
- Performance improvement
- New dataset
- Breaking change
- Documentation update

## Context

Resolves #

#### related issues:
- #

#### required by:
- #

#### requires:
- [ ] #

#### conflicts with:
- #

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] I have checked my code and corrected any misspellings

PR #1300, opened 2025-10-02
test: implement fixture to assert deprecation is removed as scheduled

dkrako:
## Description

This PR implements a fixture that can be used to assert that a deprecated function is removed as scheduled.

## Implemented changes

- [x] implement `assert_deprecation_is_removed` fixture in `tests/fixtures/deprecation_fixtures.py`
- [x] reuse `assert_deprecation_is_removed` fixture in existing tests

## How Has This Been Tested?

- [x] test that `assert_deprecation_is_removed` raises `AssertionError` on non-matching warning message
- [x] test that `assert_deprecation_is_removed` raises `AssertionError` on wrong scheduled version
- [x] test that `assert_deprecation_is_removed` raises `AssertionError` on current version later than scheduled
- [x] test that `assert_deprecation_is_removed` does not raise `AssertionError` if current version earlier than scheduled

#### requires:
- [x] #1293 
- [x] #1297 
- [x] #1299

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ The complete reference of the package can be found here:
 
 We welcome any sort of contribution to pymovements!
 
-For a detailed guide, please refer to our [CONTRIBUTING.md](CONTRIBUTING.md) first.
+For a detailed guide, please refer to our [CONTRIBUTING.md](https://github.com/pymovements/pymovements/blob/main/CONTRIBUTING.md) first.
 
 If you have any questions, please [open an issue](
 https://github.com/pymovements/pymovements/issues/new/choose) or write us at
--- a/src/pymovements/gaze/_utils/parsing.py
+++ b/src/pymovements/gaze/_utils/parsing.py
@@ -25,6 +25,7 @@ import datetime
 import re
 import warnings
 from collections import defaultdict
+from collections.abc import Sequence
 from pathlib import Path
 from typing import Any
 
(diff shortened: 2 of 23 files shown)


tests/fixtures/deprecation_fixtures.py
```diff
@@ -0,0 +1,84 @@
...
+    """Assert that function deprecation is removed as scheduled."""
+    regex = re.compile(r'.*will be removed in v(?P<version>[0-9]*[.][0-9]*[.][0-9]*)[.)].*')
+
+    def _assert_deprecation_is_removed(
+            *,
+            function_name: str,
+            warning_message: str,
+            scheduled_version: str,
+            current_version: str,
+    ) -> None:
```
cbueth, line 41:
```suggestion
    def _assert_deprecation_is_removed(
            *,
            function_name: str,
            warning_message: str,
            scheduled_version: str,
            current_version: str = __version__,
    ) -> None:
```

Consider using the default variable for `current_version`, this reduces the import in each deprecation test.

cbueth:
I checked out https://github.com/pymovements/pymovements/pull/1300/commits/cd8585bc2366b3d6fabe36d8d039efa0d6801697 and tested it.

Tox builds the package an runs it from the virtualenv, currently running from the root `.` as test target, so pytest goes into `src/` (`ERROR collecting src/pymovements/utils/paths.py`) looking for tests. This is problematic when accessing `from pymovements import __version__`, pulling it  from the venv and not the source `src/` itself.
`src/pymovements/_scripts/write_dataset_docpages.py` is collected from source, but the module with the same name is already imported from site‑packages `.tox/py312/lib/python3.12/site‑packages/pymovements/_scripts/write_dataset_docpages.py`. Are there any tests in `src/` we need?

If no: To prevent pytest to search in `src/`, the command target can be limited to the `tests/` folder: `{posargs:tests}`. See changes in ec586c31.

Setting `testpaths = ['tests', 'src']` to `testpaths = ['tests']` did not do the trick for me.

Alternatively:

- Adding `--import-mode=importlib` to `pyproject.toml` fixes it:

```toml
[tool.pytest.ini_options]
addopts = "-ra -l --tb=long --doctest-modules --import-mode=importlib"
```

- Using the editable mode for the `[testenv]` could work.

I think the first solution is the most elegant.

dkrako:
>  Are there any tests in src/ we need?

Yes `src` needs to be tested to verify all doctest examples.

I adjusted the import mode and now it works. Thank you so much for digging into this!

PR #1306, opened 2025-10-07
ci: exclude pre-commit-ci bot in release-drafter

dkrako:
## Description

I don't understand why, but the bots are never excluded in the release draft.
I removed the trailing `[bot]` and reverted the `dependabot` exclude, as updating package dependencies is useful information in the changelog.

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -32,8 +32,7 @@ exclude-labels:
   - skip changelog
 
 exclude-contributors:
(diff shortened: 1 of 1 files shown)


dkrako:
here's the doc on the configuration option:
https://github.com/marketplace/actions/release-drafter#exclude-contributors

As this will only affect the draft release message and I don't know how to test this PR without being on the main branch, I would need to merge this first to test this.

PR #1261, opened 2025-08-20
feat: filter for fixations in scanpathplot and add tutorial

sasogr:
### Plotting fixations using scatterplot.

- Modified argument list (gaze.events removed - it is extracted from the gaze argument, event_name argument added to specify the name of the fixation)
- Tutorial for proper use of scatterplot

PR #1314, opened 2025-10-14
docs: create a unified plotting tutorial

shaitarAn:
## Description

This PR restructures the plotting tutorials, to make them more unified and clear. 

## Implemented changes

- [x] removed `tutorials/heatmap.ipynb`
- [x] removed `tutorials/plot-main-sequence.ipynb`
- [x] created `tutorials/plotting.ipynb` which includes plotting instructions for all plots: main sequence, traceplot, scanpath plot, heatmap.
- [x] merged `scanpathplot` tutorial and removed it #1261 

## Type of change

- Documentation update

## Checklist:

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,30 +20,15 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - macos-13
           - macos-latest
           - ubuntu-latest
           - windows-latest
         tox_env:
-          - py39
           - py310
           - py311
           - py312
           - py313
-        exclude:
-          - os: macos-latest
-            tox_env: py39
-          - os: macos-13
-            tox_env: py310
-          - os: macos-13
-            tox_env: py311
-          - os: macos-13
-            tox_env: py312
-          - os: macos-13
-            tox_env: py313
         include:
-          - tox_env: py39
-            python: "3.9"
           - tox_env: py310
             python: "3.10"
           - tox_env: py311
@@ -110,7 +95,7 @@ jobs:
         uses: actions/cache/restore@v4
         with:
           key:
-            pylint-ubuntu-latest-py39-${{ hashFiles('tox.ini', 'pyproject.toml', '.github/workflows/tests') }}
+            pylint-ubuntu-latest-py310-${{ hashFiles('tox.ini', 'pyproject.toml', '.github/workflows/tests') }}
           path: |
             ~/.cache
             ~/.tox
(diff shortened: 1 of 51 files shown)


cbueth:
Now it builds!
The notebook summarizes the current plotting capabilities well. Notes:

- I suggest putting the plotting tutorial as the third in order of tutorials, right now it is last, but could be more prominent.
- Between the headings/after „Detecting events and computing properties“ and the two following subsections, a few words about the motivation/ one sentence describing each saccades and fixations would help much.
- There is a broken reference. Could be:
```markdown
{py:class}`~pymovements.datasets.ToyDataset`
# or
{ref}`toydataset` # works only if there is a target with this name
# what I think you tried
{doc}`ToyDataset <../../reference/api/pymovements.datasets.ToyDataset.html>`
```
might work? While the docstings use reStructuredText format, the notebooks are parsed as Markdown. It can be quite confusing https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html#reference-roles.

cbueth:
@theDebbister noticed that the Scanpathplot and Traceplot are having a non-equal aspect ratio.
I will open a seprate issue for this.

dkrako:
There's an error with the event names in the new scanpathplot tutorial. that's why the plots are empty.

cbueth:
> It would probably be cleaner to close each figure with `plt.close(fig)`. @cbueth what do you think?

Closing the figures would be the clean way, but not really an issue in this notebook, in my opinion.

What should be changed for sure is using
```python
fig, ax = pm.plotting.heatmap(
```
instead of 
```python
figure = pm.plotting.heatmap(
```
to better prime the user's expectations of the output.

Heatmap and other plotting functions offer `show` and `closefig`. As we use `show=True` the figure is plotted in the notebook. But `closefig` is `None` by default. The notebook also shows the plot, if `show=False` and if `pm.plotting.heatmap(...)` is the last line in a cell, when not assigning it to anything, as is uses the `repr`. But if `show=True` and `closefig=True`, then plots do not appear, unless they are `pm.plotting.heatmap(...)` on the last line anyways (all except `heatmap`).

The more broad question is, what should the user expect. I think it is fine to keep the logic of `finalize_figure` and defaults like so. If an advanced user wants to edit the figure further, they probably want to use `show=False`, but these users will pass an `ax` anyways. And in that case `show` and `closefig` are ignored, as per #1257.

https://github.com/pymovements/pymovements/blob/cacfe47504a704bfe5113362bbfab703a5804b10/src/pymovements/plotting/_matplotlib.py#L183

dkrako (approved):
Awesome, thansk for taking care of all the details!

PR #1317, opened 2025-10-15
ci: add codecov configuration file

dkrako:
I'm still clueless why codecov doesn't report the project coverage anymore. This might help as I think it was setup this way on the codecov project page before the migration (as global YAML on the webconfig). Setting it explicitly in this repository might fix the issue.


cbueth:
While the repository config has precedence over the global config, does the wished setting need a repo config, or could it be solved via the global one? https://docs.codecov.com/docs/codecov-yaml
It reads the config file should be on `git/hg`. I have not used codecov that detailled before.

Issue #1056, opened 2025-03-25
add task column to dataset table


dkrako:
## Description of the problem
In our [dataset table](https://pymovements.readthedocs.io/en/stable/datasets/index.html) we currently do not provide the task that participants are given in the experiment of each dataset.

## Description of a solution

Add a column with the participants' tasks of each dataset.

Issue #38, opened 2022-10-14
Add event detection metrics


dkrako:
## Description of the problem
We provide a range of event detection methods and want the user to be able to evaluate them.
There is a paper which has done this and provides python source code for evaluation.
https://link.springer.com/article/10.3758/s13428-021-01763-7
https://github.com/r-zemblys/EM-event-detection-evaluation

## Description of a solution
- [ ] create table with all metrics
- [ ] implement metrics
- [ ] add test cases and make them pass
- [ ] add documentation page with description of metrics

## Minimum acceptance criteria
- [ ] all metrics of source repository implemented with passing tests
- [ ] event detection metric module documentation page

PR #1321, opened 2025-10-20
docs: alternative dataset table columns for reduced width

### LINKED ISSUE #1320: datasets table too wide
## Current Behavior
The dataset table is too wide to be shown on a standard HD width. The rightmost columns is occluded for about ⅔.

## Expected Behavior
All columns should be visible without horizontal scrolling, especially when viewing from a desktop viewport.

No horizontal scrolling to see all columns.

### Steps to Reproduce
See https://pymovements.readthedocs.io/en/v0.24.0/datasets/

## Context
Relates to

- #1056 
- #38
 
## Checklist

cbueth:
## Description

The tables are  just a bit too wide for the viewport with the wide/desktop view and will never fit the screen: #1320.
Somehow they should fir the screen without horizontal scrolling. This can be achieved in several ways.

I was not able to rotate the columns labels (Samples and Events) to save space and thought merging the first two columns would work well. But sadly I did not find a solution to add a column break for better readability between name and citation.
One suggestion has been implemented, but is open to discussion.

## Implemented changes

United the first two columns and reduced width.

Furthermore, slightly reduced the width of the example data table.

## Suggested changes

image

I do agree that this is not looking perfect, but it suffices the minimal criteria, not scrolling horizontally.

## Before

image

## Type of change

Remove irrelevant items:
- Documentation update

## Context

Resolves #1320 

#### related issues:
- #1056
- #38

#### conflicts with:
- #?

## Checklist:

### DIFF
--- a/tests/unit/utils/plotting_test.py
+++ b/tests/unit/utils/plotting_test.py
@@ -20,7 +20,7 @@
 """Test pymovements plotting utils."""
 import re
 
-import matplotlib.pyplot
+import matplotlib.pyplot as plt
 import numpy as np
 import pytest
 
@@ -38,7 +38,7 @@ from pymovements.utils.plotting import setup_matplotlib
     ],
 )
 def axes_fixture(request):
-    fig = matplotlib.pyplot.figure(figsize=request.param)
+    fig = plt.figure(figsize=request.param)
     yield fig.gca()
 
 
@@ -58,17 +58,20 @@ def axes_fixture(request):
 )
 def test_setup_matplotlib(kwargs):
     setup_matplotlib(**kwargs)
+    plt.close()
 
 
 @pytest.mark.filterwarnings('ignore::DeprecationWarning')
 def test_draw_image_stimulus(axes, make_example_file):
     filepath = make_example_file('pexels-zoorg-1000498.jpg')
     draw_image_stimulus(image_stimulus=filepath, ax=axes)
+    plt.close()
 
 
 @pytest.mark.filterwarnings('ignore::DeprecationWarning')
 def test_draw_line_data(axes):
     draw_line_data(x_signal=np.array([0.0, 0.0]), y_signal=np.array([0.0, 0.0]), ax=axes)
+    plt.close()
 
 
 @pytest.mark.parametrize(
@@ -100,6 +103,7 @@ def test_draw_line_data(axes):
 def test_plotting_function_deprecated(plotting_function, kwargs):
     with pytest.raises(DeprecationWarning):
         plotting_function(**kwargs)
+    plt.close()
 
 
 @pytest.mark.parametrize(
@@ -131,6 +135,7 @@ def test_plotting_function_deprecated(plotting_function, kwargs):
 def test_plotting_function_removed(plotting_function, kwargs):
     with pytest.raises(DeprecationWarning) as info:
         plotting_function(**kwargs)
+    plt.close()
 
     regex = re.compile(r'.*will be removed in v(?P<version>[0-9]*[.][0-9]*[.][0-9]*)[.)].*')
 


docs/source/datasets/index.rst
```diff
@@ -35,7 +35,7 @@ Download your dataset and load it into memory with these few lines of code:
 
 .. csv-table::
    :file: example_datasets.csv
-   :widths: 65, 5, 5, 5, 5, 5, 5, 5
+   :widths: 45, 5, 5, 5, 5, 5, 5, 5
```
dkrako, line 38: the widths should sum up to 100 (they are percentages) so it seems 65 already was the correct value

docs/source/datasets/public_datasets.csv
```diff
@@ -1,32 +1,32 @@
...
-:py:class:`~pymovements.datasets.MECOL2W2`,:cite:p:`MECOL2W2`,,X,,660,1,Various Hz,Text (documents)
-:py:class:`~pymovements.datasets.OneStop`,:cite:p:`OneStop`,,X,X,360,1,1000 Hz,Text (documents)
-:py:class:`~pymovements.datasets.PoTeC`,:cite:p:`PoTeC`,X,,,75,1,1000 Hz,Text (documents)
-:py:class:`~pymovements.datasets.PotsdamBingeRemotePVT`,:cite:p:`PotsdamBingePVT`,X,,,57,2,1000 Hz,Dots (PVT test)
-:py:class:`~pymovements.datasets.PotsdamBingeWearablePVT`,:cite:p:`PotsdamBingePVT`,X,,,57,2,200 Hz (upsampled to 1000 Hz),Dots (PVT test)
-:py:class:`~pymovements.datasets.Provo`,:cite:p:`Provo`,,X,,470,1,1000 Hz,Text (documents)
-:py:class:`~pymovements.datasets.SBSAT`,:cite:p:`SB-SAT`,X,X,,95,1,1000 Hz,Text (documents)
-:py:class:`~pymovements.datasets.UCL`,:cite:p:`UCL`,,X,X,43,1,500 Hz,Text (sentences)
+Dataset,Samples,Events,Reading Measures,# Subjects,# Sessions per Subject,Sampling Rate,Stimulus Type
+":py:class:`~pymovements.datasets.BSC`, :cite:p:`BSC`",,X,,60,1,500 Hz,Text (sentences)
```
dkrako: I would prefer to leave out the comma between class and citation

docs/source/datasets/index.rst
cbueth: I decreased it, as the table seemed quite bloated.
For the other datasets `30, 35, 5, 5, 5, 5, 5, 5, 5` before also summed to 100, which I changed to  `45, 5, 5, 5, 5, 5, 5, 5` summing to 80.

docs/source/datasets/public_datasets.csv
cbueth: Ok!

cbueth:
> This discussion here could help with adjusting the width by setting up a custom css stylesheet: [pydata/pydata-sphinx-theme#190](https://github.com/pydata/pydata-sphinx-theme/pull/190)
> 
> Maybe this approach would be more future proof?

To my impression this would style all tables, but not single cells, correct? I do not think this is what we want, as then the header row gets quite tall for wider tables.
> 
> EDIT: Here's also a page on styling: https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/styling.html
> 
> Probably the `$breakpoint-page-width` would be the variable to update.


Does that mean making the usable page width wider?

dkrako:
> Does that mean making the usable page width wider?

I'm not sure but I hope so

docs/source/datasets/index.rst
dkrako: Hmm if I compare your PR with the current docs, the example dataset table has the exact same dimensions:

https://pymovements--1321.org.readthedocs.build/en/1321/datasets/index.html#example-datasets
https://pymovements.readthedocs.io/en/stable/datasets/index.html#example-datasets

So it doesn't seem that the change had any effect, which is unexpected and leads me to thinking that not summing to 100 leads to unspecified behavior.

The [documentation on the csv-table directive](https://docutils.sourceforge.io/docs/ref/rst/directives.html#csv-table) just specifies the widths as a *list of relative column widths* but doesn't really specify to what it is relative to (table width? relative to other column widths?).

All in all I really think the values should sum up to 100.

docs/source/datasets/public_datasets.csv
dkrako: or alternatively: can you try a newline inbetween class and citation? maybe this would look a bit less convoluted

docs/source/datasets/index.rst
cbueth: They are relative to each other, so the table will always be at the divs width. So if we have `2, 1` the left column would be double as wide as the right one. WIth to the one exception if the cell content is just too wide and cannot be broken to multiple lines to fit the width.

Specifically `[PotsdamBingeWearablePVT](https://pymovements.readthedocs.io/en/stable/reference/api/pymovements.datasets.PotsdamBingeWearablePVT.html#pymovements.datasets.PotsdamBingeWearablePVT)` cannot be broken, so the first column will always be at least this wide.

docs/source/datasets/public_datasets.csv
cbueth: The commas have been removed in 1c23bf3.

docs/source/datasets/index.rst
dkrako: Yes that's what I thought after rereading the csv-table documentation, but then I still don't understand why changing the widths had no effect on the example table. The first column should be reduced but it isn't so something doesn't add up.

docs/source/datasets/public_datasets.csv
dkrako: ah sorry my comment was too late

docs/source/datasets/index.rst
cbueth: True, I compared them on my side and it is unchanged. It could be that with `45, 5, 5, 5, 5, 5, 5, 5` the widths of the small columns `5` is too small, so they all have slightly different widths, as the minimum of what their content allows, see 'Events' being the smallest column, but also just as wide as the word.

When changing to `20, 10, 10, 10, 10, 10, 10, 10`, the columns seem to be equally wide, with the exception of maybe the last column.

image

dkrako: Yes that could actually be the issue. `20, 10, 10, 10, 10, 10, 10, 10` seems fine

PR #1309, opened 2025-10-09
test: add fixture to autoclose all figures on test teardown

shaitarAn:
## Description

This PR adds a `conftest.py` file to `tests/unit/plotting` to hold universla fixtures like closing figures.

## Implemented changes

- [x] added `conftest.py` file to `tests/unit/plotting` 

## Type of change

Remove irrelevant items:

- New functionality

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

PR #1327, opened 2025-10-21
fix: do not raise error if plotting empty data

dkrako:
## Description

If input data for plotting functions was either empty or was equal on a single dimension, setting up the axes limits didn't work properly in `_setup_axes_and_colormap()`. This PR handles these edge cases.

## Implemented changes

- [x] only set axes limits if more than one element in data
- [x] only set axes limits if min value is not equal to max value
- [x] only set color bar values if data is not empty

## How Has This Been Tested?

- [x] parametrized existing fixtures such that different data lengths are produced

#### requires:
- [x] #1309

#### required by:
- #1261

## Checklist:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -53,7 +53,7 @@ repos:
         args: [--use-current-year]
         types: [python]
 -   repo: https://github.com/jsh9/pydoclint
-    rev: 0.7.3
+    rev: 0.7.6
     hooks:
     - id: pydoclint
       args: ["--config=pyproject.toml"]


tests/unit/plotting/scanpathplot_test.py
```diff
@@ -35,23 +35,68 @@
...
                 'location': [(1, 2), (2, 3)],
             },
-        ),
-    ).clone()
+        )
+    elif request.param == '3_events':
+        events = pl.DataFrame(
+            data={
+                'trial': [1, 1, 1],
+                'name': ['fixation', 'saccade', 'foo'],
```
cbueth, line 84: 'name' can be any string here?

dkrako: yes, we also support custom event names in `Events`.

src/pymovements/plotting/_matplotlib.py
```diff
@@ -238,8 +238,8 @@ def _setup_axes_and_colormap(
         ax.imshow(img, origin=stimulus_origin, extent=None)
     else:
         if n > 0:  # autoset axes limits if there is at least one data point
-            x_min, x_max = np.nanmax(x_signal) - np.nanmin(x_signal)
-            y_min, y_max = np.nanmax(y_signal) - np.nanmin(y_signal)
+            x_min, x_max = np.nanmax(x_signal), np.nanmin(x_signal)
+            y_min, y_max = np.nanmax(y_signal), np.nanmin(y_signal)
```
cbueth: `x_min` is actually the `nanmax` of `x_signal`, which sounds unintuitive to me, is this by design?

cbueth (changes requested):
Thank you for adding the comments, I do think it's easier to understand now.

src/pymovements/plotting/_matplotlib.py
dkrako: good catch

cbueth: Jetzt stimmts :)

cbueth (approved):

PR #1337, opened 2025-10-22
docs: use MyST to cross-reference tutorials to sphinx pages

cbueth:
**TL;DR**: If you got sent here and want to know more about the MyST reference roles, look here https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html#reference-roles. If you want to know about the ecosystem, read this comment https://github.com/pymovements/pymovements/pull/1337#discussion_r2459758876.

## Description

To prevent broken links, this PR adds the [MyST-NB](https://myst-nb.readthedocs.io/en/v1.3.0/index.html) extension to enable cross-referencing in the tutorial notebooks.

This change is useful, as it allows the future usage of MyST, e.g.
- references roles https://mystmd.org/guide/cross-references#reference-roles
- admonitions https://mystmd.org/guide/admonitions
- tables https://mystmd.org/guide/tables, as referenced in #1321
- among others ... https://mystmd.org/

## Implemented changes

Insert a description of the changes implemented in the pull request.

- [x] Add dependency `myst-nb`
- [x] Add sphinx extension `myst_nb`
- [x] Add references to the datasets, events and plotting reference pages.
- [x] Use `{ref}` directive in notebooks, as per #1336. See https://mystmd.org/guide/cross-references#reference-roles
- [x] Add further directives in notebooks.
- [x] .gitignore `docs/jupyter_execute`
- [x] Activated intersphinx to easily link to functionality of dependent packages, e.g. `polars.read_csv`. This also links python intrinsic types in the API reference.

## How Has This Been Tested?

Local building of pages. All new links have been tested.

## Type of change

Remove irrelevant items:
- New functionality
 
## Context

Resolves #

#### related issues:
- #1312
- #1336
- #1261 

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

PR #1344, opened 2025-10-28
docs: activate nitpicky sphinx option

cbueth:
## Description

When using cross references in sphinx, it can happen that the reference does not exist due to various reasons, eg typos or code changes, breaking links. To speed up development and prevent dead links, sphinx has the option `nitpicky` which should be activated.
This activation also brings many warnings that can be silenced, as sphinx has various places cross references are tried to be inferred, but are not necessarily needed.

## Implemented changes

- [x] Activate `nitpicky` option
- [x] For all new warnings: Ignore reasonabele errors or fix references.
- [ ] Assure if any of the ignored references should rather not be ignored and actively referenced

## How Has This Been Tested?

- [x] Local `tox -e docs` run

## Type of change

Remove irrelevant items:
- Documentation update

#### related issues:
- #1337

#### requires:
- [ ] #1337

## Checklist:

  - The doc build serves as test

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -107,7 +107,7 @@ copybutton_prompt_is_regexp = True
 copybutton_line_continuation_character = '\\'
 copybutton_here_doc_delimiter = 'EOT'
 
-# -- General cross-references ---------------------------------------------------
+# -- Options for cross-references ---------------------------------------------------
 
 # Help Napoleon resolve common short type names in docstrings
 napoleon_type_aliases = {


cbueth:
Initially, activating `nitpicky` mode resulted in 845 warnings. These have been grouped and ignored.
For reviewing:
- Please read though 17c7ce9683d2224506a8be2c8f031baa49a36db2 to get an idea where sphinx tries and fails to inference cross references. If any of these should rather have working cross references, check out the branch, uncomment the ignore regex, `tox -e docs` and see where the references appear in the docs. The high number is not especially large, sphinx just likes to infer references anywhere possible.
- Then consider how to change the references to activate them. Even if we keep all of the ignord messages as in 17c7ce9683d2224506a8be2c8f031baa49a36db2, any new reference that is not being found will result in a warning and thus CI/docs fail. Same is the case for references that turn out invalid when changing internals. Ergo: keeping all ignores is also a valid in my opinion.

dkrako (approved):
Thanks a lot for your initiative! I think it's fine to have all these ignores for this initial PR but I would like to get rid of all the ignores eventually.

PR #1162, opened 2025-05-26
depr: deprecate DatasetDefinition.filename_format in favor of Resource.filename_pattern

dkrako:
…e.filename_pattern## Description

Please include a clear and concise description of the change and which issue is fixed.
Please also include excepted improvements, relevant motivation and context.
List any dependencies that are required for this change.

## Implemented changes

Insert a description of the changes implemented in the pull request.

- [ ] Change 1
- [ ] Change 2
- [ ] Change 3

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes.
Provide instructions so we can reproduce.
Please also list any relevant details for your test configuration

- [ ] Test A
- [ ] Test B

## Type of change

Remove irrelevant items:
- Bug fix
- New functionality
- Performance improvement
- New dataset
- Breaking change
- Documentation update

## Context

Resolves #

#### related issues:
- #

#### required by:
- #

#### requires:
- [ ] #

#### conflicts with:
- #

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] I have checked my code and corrected any misspellings

PR #1347, opened 2025-10-29
depr: deprecate `filename_format` as `DatasetDefinition` init parameter

dkrako:
## :warning: Deprecation

`DatasetDefinition` init parameter `filename_format` was deprecated in v0.24.1. Please use `ResourceDefinition.filename_pattern` instead.
`DatasetDefinition.filename_format` will be removed in pymovements v0.28.0

## Description

Before this change, passing `filename_format` to `DatasetDefinition.__init__()` would not raise a `DeprecationWarning` if `resources` was of type `ResourceDefinitions`. I updated some of the logic such that the warning is now raised in any case.

## Implemented changes

- [x] improve property setters to also support legacy definitions that have not specified resources
- [x] simplify resource initiatialization by reusing property setters

## How Has This Been Tested?

- [x] tested deprecation warnings during initialization
- [x] test that values are set correctly even if no resources specified

## Context

Related issues: #1162

## Checklist:


src/pymovements/dataset/dataset_definition.py
```diff
@@ -568,41 +613,21 @@ def has_resources(self) -> _HasResourcesIndexer:
...
             resources: ResourceDefinitions | ResourcesLike | None,
-            filename_format: dict[str, str] | None,
-            filename_format_schema_overrides: dict[str, dict[str, type]] | None,
     ) -> ResourceDefinitions:
         """Initialize ``ResourceDefinitions`` instance if necessary."""
         if isinstance(resources, ResourceDefinitions):
             return resources
 
         if resources is None:
-            # some legacy definitions may have defined filename format but no resources.
```
dkrako, line 580: this is the reason why no deprecation warning was raised, as it circumvented that warning.

now this logic was moved to the deprecated filename_format setter.

src/pymovements/dataset/dataset_definition.py
```diff
@@ -568,41 +599,21 @@ def has_resources(self) -> _HasResourcesIndexer:
...
-                return ResourceDefinitions()
-
-        # this calls deprecated methods and will be removed in the future.
-        if isinstance(resources, dict):
-            if filename_format:
-                for content_type in filename_format:
-                    for resource_dict in resources[content_type]:
-                        resource_dict['filename_pattern'] = filename_format[content_type]
-            if filename_format_schema_overrides:
-                for content_type in filename_format_schema_overrides:
```
dkrako, line 597: this too is moved to the setter

tests/unit/dataset/dataset_definition_test.py
```diff
@@ -834,9 +846,32 @@ def test_dataset_to_dict_exclude_none(dataset_definition, exclude_none, expected
...
+            '0.28.0',
+            id='resources_dict_legacy',
+        ),
+        pytest.param(
+            {
+                'resources': [{'content': 'gaze'}],
+                'filename_format': {'gaze': '{subject}.csv'},
+            },
+            '0.28.0',
+            id='filename_format',
```
dkrako, line 862: these were not raised before

src/pymovements/dataset/dataset_definition.py
cbueth: You mean if resources were None, no warning was emitted per-se?

cbueth (approved):
Good work, looks elegant.

src/pymovements/dataset/dataset_definition.py
dkrako: Before this would only raise a `DeprecationWarning` for calling `ResourceDefinitions.from_dict` in the next if section.

My previous comment here is a bit misleading, because the real culprit was further down to `if isinstance(resources, Sequence):`

This definition wouldn't have raise a warning and wouldn't set filename format correctly:

```python
DatasetDefinition(
    resources=[Resource(content='gaze')],
    filename_format={'gaze': 'abc.csv'},
)
```

PR #1348, opened 2025-10-29
chore: move messages to Gaze

dkrako:
## Description

After some discussion and consideration I think it's best to have the `messages` field at the `Gaze` level and not the `Experiment` level.

The `Experiment` class holds all static experiment related data (i.e. data does not change during a recording session), whereas the `Gaze` class holds all data related to a particular session. It is a monolithic class that keeps data of a single session consistent across various representations. That's why `Gaze` is a better place to hold the messages.

Further I think we should restrict our use of monolithic classes to `Dataset` and `Gaze`.

## Implemented changes

- [x] moved message init from Experiment to Gaze
- [x] adjusted Gaze.__str__()
- [x] removed warning for already existing messages in from_asc
- [x] adjusted from_asc

## How Has This Been Tested?

- [x] migrated all existing tests if possible
- [x] removed test for warning
- [x] removed test for to_dict()

### DIFF
--- a/src/pymovements/gaze/gaze.py
+++ b/src/pymovements/gaze/gaze.py
@@ -1887,10 +1887,8 @@ class Gaze:
     def __str__(self) -> str:
         """Return string representation of Gaze.
 
-        The messages field expresses:
-        - 'messages=None' if no messages are provided.
-        - 'messages=<N> rows' if a messages DataFrame is provided,
-          where N is the number of rows.
+        If :py:attr:`~.Gaze.messages` is not ``None``, includes ``messages=<N> rows``,
+        where ``N`` is the number of rows.
         """
         fields = []
 
@@ -1908,10 +1906,8 @@ class Gaze:
     def __repr__(self) -> str:
         """Return string representation of Gaze.
 
-        The messages field expresses:
-        - 'messages=None' if no messages are provided.
-        - 'messages=<N> rows' if a messages DataFrame is provided,
-          where N is the number of rows.
+        If :py:attr:`~.Gaze.messages` is not ``None``, includes ``messages=<N> rows``,
+        where ``N`` is the number of rows.
         """
         return self.__str__()
 


tests/unit/gaze/gaze_messages_test.py
```diff
@@ -0,0 +1,110 @@
+# Copyright (c) 2024-2025 The pymovements Project Authors
```
SiQube, line 1: tests should probably be in `gaze_test.py`

dkrako: I actually prefer to split this into an own test module, because `gaze_test.py` is already massive in size.

there will be message related preprocessing functions coming (messages to trials, messages to sample/events columns, etc.). testing all this functionality in a separate file seems a bit cleaner to me.

but I can make the change if you very strongly prefer to keep them in a single file

src/pymovements/gaze/gaze.py
```diff
@@ -1872,14 +1885,30 @@ def __eq__(self, other: Gaze) -> bool:
         return samples_equal and events_equal and experiment_equal and trial_columns_equal
 
     def __str__(self) -> str:
-        """Return string representation of Gaze."""
-        if self.experiment is None:
-            return self.samples.__str__()
+        """Return string representation of Gaze.
```
dkrako, line 1888: the `Gaze.__str__()` implementation is currently really lacking and should be improved in a separate PR.

cbueth: What changes do you have in mind?

tests/unit/gaze/gaze_messages_test.py
cbueth: I do think splitting it into another file is valid here.

src/pymovements/gaze/gaze.py
dkrako: the string should probably start with `Gaze(` and include more info on fields

PR #1359, opened 2025-11-05
fix: preserve all columns in `data` on `Events` init

### LINKED ISSUE #1349: `gaze.clone()` overwrites properties
When cloning an `Events` object that has event properties added via `add_event_properties()`, the properties are lost. The `clone()` method does not preserve additional event property columns.

https://github.com/pymovements/pymovements/blob/43bfff286d1b91a4c50c9ed124506a2cf018cd34/src/pymovements/events/events.py#L182-L183

how I fixed it locally
```python
if self.trial_columns is not None:
      # check what other columns exist in the dataframe, without minimal schema and trial cols
      other_cols = [
          col for col in self.frame.columns
          if col not in self.trial_columns and col not in self._minimal_schema
      ]
      self.frame = self.frame.select([*self.trial_columns, *self._minimal_schema.keys(), *other_cols])
```

## Current Behavior
- if I run gaze.clone(), on a gaze instance which already includes event properties, the properties are overwritten and are not cloned together with the rest of the gaze object

## Expected Behavior
The cloned Events object should preserve all columns, including event properties, in the same order as the original.

- `clone()` preserves all event property columns
- Column order is maintained
- Data integrity is preserved for all columns

### Steps to Reproduce

- Creating an Events instance with trial data
- Adding event properties via `add_event_properties()`
- Calling `clone()` on the Events object
- -> The cloned Events object is missing the additional property columns

## Context

In `Events.__init__()` at lines 182-183, when `trial_columns` is set, the code reorders columns using:

https://github.com/pymovements/pymovements/blob/43bfff286d1b91a4c50c9ed124506a2cf018cd34/src/pymovements/events/events.py#L182-L183

This selects only trial columns and minimal schema columns, silently dropping any additional columns added via `add_event_properties()`. Any workflow using `clone()` on Events with properties will lose data silently.

## Checklist

cbueth:
## Description

Fixes issue where `Events.clone()` loses event properties when `trial_columns` are set. The column selection logic now preserves additional event property columns alongside trial columns and minimal schema columns. Find detailed analysis in the #1349 issue description.

## Implemented changes

- [x] Modified `Events.__init__()` to preserve additional columns beyond minimal schema and trial columns
- [x] Changed column selection to: `[*trial_columns, *minimal_schema.keys(), *other_cols]`
- [x] Added logic to detect and preserve event property columns during initialization

## How Has This Been Tested?

- [x] Added regression test `test_clone_preserves_event_properties_with_trial_columns()` in `tests/unit/events/events_test.py`. Passes after fix, fails without it.
- [x] Verified fix preserves columns when cloning Events with trial data and added properties
- [x] Existing clone tests remain passing
## Type of change

- Bug fix - non-breaking change

## Context

Resolves #1349

#### related issues:
- #

#### required by:
- #

#### requires:
- [ ] #

#### conflicts with:
- #

## Checklist:


tests/unit/events/events_test.py
```diff
@@ -538,6 +538,44 @@ def test_clones_trial_columns():
...
+
+    # Verify trial_columns is set (this triggers the column reordering logic)
+    assert events.trial_columns is not None
+
+    # Add event properties as additional columns
+    event_properties = pl.DataFrame({
+        'name': ['fixation', 'saccade'],
+        'custom_property': [1.5, 2.5],
+    })
+    events.add_event_properties(event_properties, join_on='name')
```
dkrako: By initiatilizing your events this way:

```python
events = Events(
    data=pl.from_dict({
        'name': ['fixation', 'saccade', 'fixation'],
        'onsets': [100, 200, 300],
        'offsets': [150, 250, 350],
        'trial_id': [1, 1, 2],
        'custom_property': [1.5, 2.5, 1.5],
    },
    trial_columns='trial_id',
)
```

you omit test dependencies by not calling any unrelated pymovements methods.

tests/unit/events/events_test.py
```diff
@@ -538,6 +538,44 @@ def test_clones_trial_columns():
...
+    original_columns = events.columns
+    original_len = len(events.frame)
+
+    # Clone the events
+    cloned = events.clone()
+
+    # Verify all columns are preserved in the clone
+    assert cloned.columns == original_columns, \
+        f"Cloned columns {cloned.columns} do not match original {original_columns}"
+    assert len(cloned.frame) == original_len
```
dkrako: these asserts are redundant as already tested by `assert_frame_equal()`

tests/unit/events/events_test.py
```diff
@@ -538,6 +538,44 @@ def test_clones_trial_columns():
     assert events.trial_columns == events_copy.trial_columns
 
 
+def test_clone_preserves_event_properties():
+    """Test that clone() preserves event properties added via add_event_properties().
+
+    Regression test for issue #1349
```
dkrako: the way this test is written it is actually not a unit test, but a functional test (also called acceptance test), that tests the interplay between calling different units.

Although there are quite a lot of old tests in `tests/unit/` written in that way, I would like avoid that in the future and put them where they belong in `tests/functional` (or alternatively `tests/functional/regressions`, `tests/regressions`, ).

the test at hand however can also be just subsumed as a test parameter to the already existing test function `test_init_expected_df` by using the dataframe from my proposed `pl.from_dict()` codeblock

dkrako (commented):
Great fix, thank you so much for your initiative!

This PR fixes more than it originally claims, it fixes a bug in the `Events.__init__()` method that shows when calling `Events.clone()` with computed event properties.

I don't have any further comments regarding the `src` code, but I would prefer to have some additional `test_init_expected_df` parameters that test the init function, not just the `clone()`.

Regarding the already written test you can either just subsume it in a `test_init_expected_df` parameter or if you want to keep it, then move it to something like `tests/functional/gaze_regressions_test.py`.

tests/unit/events/events_test.py
dkrako: although the first one might not be redundant, as it checks the `Events.columns` field not the dataframe

tests/unit/events/events_test.py
cbueth: True, thanks for the hint!

src/pymovements/events/events.py
```diff
@@ -180,7 +180,15 @@ def __init__(
 
         # Ensure column order: trial columns, name, onset, offset.
         if self.trial_columns is not None:
-            self.frame = self.frame.select([*self.trial_columns, *self._minimal_schema.keys()])
+            minimal_cols = set(self._minimal_schema.keys())
+            trial_cols_set = set(self.trial_columns)
+            other_cols = [
+                col for col in self.frame.columns
+                if col not in trial_cols_set and col not in minimal_cols
```
dkrako: is there a reason why you don't just write this and leave out the variables above?

```suggestion
                if col not in [*self.trial_columns, *self._minimal_schema.keys()]
```

tests/unit/events/events_test.py
```diff
@@ -538,6 +580,20 @@ def test_clones_trial_columns():
     assert events.trial_columns == events_copy.trial_columns
 
 
+def test_clone_preserves_event_properties():
```
dkrako: can you also add a test here that tests for the bug in #1349?

https://github.com/pymovements/pymovements/blob/07d7fb0899c246c6519132602e23819fd9d21cdd/tests/unit/gaze/gaze_test.py#L384-L399

Probably best to parametrize `gaze` in this test function.

tests/unit/events/events_test.py
```diff
@@ -538,6 +580,20 @@ def test_clones_trial_columns():
...
+    """Test that clone() preserves event properties added via add_event_properties().
+
+    Regression test for issue #1349
+    """
+    events = Events(
+        data=pl.from_dict(
+            {'name': ['saccade'], 'onsets': [0], 'offsets': [1]},
```
dkrako: this test would have passed before and doesn't really test for regression against #1349.

what wouldn't have passed is if trial columns and additional non-trial columns were provided in `data`.

may be worth to just parametrize `test_clone()` some lines above.

src/pymovements/events/events.py
dkrako: or rather this would be even better IMO:

```suggestion
                if col not in self.trial_columns and col not in self._minimal_schema.keys()
```

cbueth: There is no specific reason for using variables, will do!

cbueth: Using `.keys()` is not needed. The comprehension of keys is demotivated by `pylint` with `[C0201(consider-iterating-dictionary)`. But it works without.

PR #1365, opened 2025-11-11
chore!: don't add fileinfo columns to loaded files

dkrako:
## :boom: Breaking Change

This PR removes the adding of fileinfo columns (parsed from the filename using `filename_pattern`) to `Gaze` and `Events`. These columns always consisted of a single constant value per `Gaze`/`Events` object and were not of any use.

## Implemented changes

- [x] remove adding fileinfo columns when loading dataset `Gaze` files
- [x] remove adding fileinfo columns when loading dataset `Events` files
- [x] remove adding fileinfo columns when detecting events

## How Has This Been Tested?

This PR doesn not add any new tests.

- [x] adjusted existing tests
- [x] removed irrelevant add_fileinfo tests

#### required by:
- #1371

#### requires:
- [x] #1377

## Checklist:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,7 +13,7 @@ repos:
     hooks:
       - id: add-trailing-comma
   - repo: https://github.com/asottile/pyupgrade
-    rev: v3.21.1
+    rev: v3.21.2
     hooks:
       - id: pyupgrade
         args: [ --py310-plus ]
@@ -59,7 +59,7 @@ repos:
         args: [ --use-current-year ]
         types: [ python ]
   - repo: https://github.com/jsh9/pydoclint
-    rev: 0.8.1
+    rev: 0.8.2
     hooks:
       - id: pydoclint
         args: [ "--config=pyproject.toml" ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,9 +72,9 @@ dev = [
   "pytest-benchmark>=5.1.0",
   "pytest-cov>=4.0.0",
   "setuptools>=61.0.0",
-  "types-tqdm>=4.64.0",
+  "types-tqdm>=4.67.0",
   "pre-commit>=3.0.0",
-  "pylint>=3.0.0"
+  "pylint>=4.0.0"
 ]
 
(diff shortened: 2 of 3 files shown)


cbueth (changes requested):
Thank you for the PR! I do think it looks good, just want you to consider a few thoughts few points:

- Is this drop in `save_events` still needed?

https://github.com/pymovements/pymovements/blob/9a191bfccd1ba7d923ab7ae65bb27318e2d4b98b/src/pymovements/dataset/dataset_files.py#L602-L604

- This PR should not get rid of the whole `fileinfo` dataframe, right? There are still references in docstrings, and documentation/notebooks that might be relevant.

- We might want to rename these in `io.py`, `fileinfo_key` -> `column_key` for consistency - or in #1371:

https://github.com/pymovements/pymovements/blob/f06ecf692cb8c4e482d18a9b9887f1d85ae15741/src/pymovements/gaze/io.py#L505-L508

  There are two more of these list comprehensions in the file.

- How about adding regression tests checking that `Gaze.samples.columns` and `Events.columns` do not contain any fileinfo-derived fields after loading and detection?

dkrako:
> Thank you for the PR! I do think it looks good, just want you to consider a few thoughts few points:
> 
>     * Is this drop in `save_events` still needed?

Good catch, removed the redundancy.

>     * This PR should not get rid of the whole `fileinfo` dataframe, right? There are still references in docstrings, and documentation/notebooks that might be relevant.

Yes, the `fileinfo` dataframe is still kept until we manage to add a good replacement. The references in the docs "fortunately" didn't document this adding of fileinfo columns to the gaze and events objects so no updates necessary there.

>     * We might want to rename these in `io.py`, `fileinfo_key` -> `column_key` for consistency - or in #1371:

Good catch, changed to `column_key`/`column_dtype`.

> There are two more of these list comprehensions in the file.
> 
>     * How about adding regression tests checking that `Gaze.samples.columns` and `Events.columns` do not contain any fileinfo-derived fields after loading and detection?

As this isn't a bugfix I don't think it's worth to add and keep the tests for all future versions.

PR #1366, opened 2025-11-11
chore!: drop support for Python 3.9

dkrako:
## Description

It's time for dropping support of Python 3.9.

It has reached [end-of-life](https://devguide.python.org/versions/) two weeks ago and today the macos-13 worker was [phased out](https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/), which means that we no longer can test Python3.9 on macosx.

This PR also raises all dependency minimum version bounds to versions that were released after Python 3.10 (2021-10-04)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -72,7 +72,7 @@ repos:
       - id: nbqa-isort
         args: [ "--float-to-top" ]
       - id: nbqa-pyupgrade
-        args: [ "--py39-plus" ]
+        args: [ "--py310-plus" ]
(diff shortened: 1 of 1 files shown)


pylintrc
```diff
@@ -40,10 +40,6 @@ load-plugins=
 # number of processors available to use.
 jobs=1
 
-# When enabled, pylint would attempt to guess common misconfiguration and emit
-# user-friendly hints instead of false-positive error messages.
-suggestion-mode=yes
```
dkrako, line 45: this option was removed from pylint

Issue #881, opened 2024-10-24
Plotting the main sequence fit and R2


ivanvajs1996:
Modifying the main_sequence_plot function so that it includes the linear fit line, with the line color option, and plotting the R2 parameter.

PR #1367, opened 2025-11-11
feat!: add linear fit line to saccade main sequence plot

shaitarAn:
## :boom: Breaking change

All arguments except for `events` were set to keyword only. If you passed any of these as a positional argument before, you will now have to specify the keyword argument.

Further, the following arguments have been renamed:
- `alpha` -> `marker_alpha`
- `color` -> `marker_color`

## Description

This PR continues the work started in #883 The original PR added a linear regression fit line and R² annotation to the `main_sequence_plot` function, but it had diverged from `main` and included outdated figure-handling logic.

This PR restores compatibility with the current plotting architecture (using `prepare_figure` , `finalize_figure` and axes ownership semantics), resolves merge conflicts, and updates event input handling to support both `Events` and `EventDataFrame` inputs.

## Implemented changes

- [x] Integrated the linear fit line and R² annotation from #883 
- [x] Resolved merge conflicts against the current main branch
- [x] implemented `prepare_figure` and `finalize_figure`
- [x] implemented changes suggested in #833 
      -  Added `fit` argument to enable/disable regression line drawing (default: `True`)
      - Added `fit_measure` argument to enable/disable goodness-of-fit annotation and to allow selecting the metric ("r2" or "s")
- [x] added unit tests

## How Has This Been Tested?

- [x] visualization in a notebook
- [x] unit tests for measure and fit options

## Type of change

Remove irrelevant items:
- Bug fix
- Performance improvement

## Context

Resolves Issue #881 
closes PR #883 

#### related issues:
- #

#### dependent on
- [x] #1366

## Checklist:


cbueth:
@shaitarAn as #1366 has been merged into main now, updating your branch hopefully resolves the minimum requirement pipeline.

pyproject.toml
```diff
@@ -41,6 +41,7 @@ dependencies = [
   "pyopenssl>=22.0.0",
   "pyreadr>=0.5.2",
   "pyyaml>=6.0.2",
+  "scikit-learn>=1.2",
```
dkrako, line 44: #1374 made the upper version bounds obsolte, I updated accordingly

src/pymovements/plotting/main_sequence_plot.py
```diff
@@ -55,10 +61,19 @@ def main_sequence_plot(
     ----------
     events: Events | EventDataFrame | None
         It must contain columns "peak_velocity" and "amplitude".
+    fit: bool
+        Draw a linear fit line if True. If False, no line is drawn.
+    measure: bool | Literal['r2', 's']
```
dkrako: rename this to `fit_measure` so that confusion with eye movement measures is avoided

src/pymovements/plotting/main_sequence_plot.py
```diff
@@ -55,10 +61,19 @@ def main_sequence_plot(
     ----------
     events: Events | EventDataFrame | None
         It must contain columns "peak_velocity" and "amplitude".
+    fit: bool
```
dkrako, line 72: please move all fit-related arguments (`fit`, `fit_color`, `fit_measure`) after the `marker` argument such that they are all grouped together.

src/pymovements/plotting/main_sequence_plot.py
```diff
@@ -163,6 +178,40 @@ def main_sequence_plot(
...
+            elif measure == 's':
+                s = np.sqrt(np.sum(residuals**2) / (len(residuals) - 2))
+                val = np.round(s, 3)
+                label = f"S = {val}"
+
+            else:
+                raise ValueError("measure must be one of: True, False, 'r2', 's'")
+
+            line_axes.text(
+                0.05, 0.8, label,
```
dkrako: I'm unsure if I would rather prefer the text box at the lower right but let's discuss this first in tomorrow's meeting.

docs/source/tutorials/plotting.ipynb
dkrako, line 481: A sentence here on the `fit` argument would be great

docs/source/tutorials/plotting.ipynb
```diff
@@ -492,6 +492,7 @@
     "for event_df in dataset.events[:3]:\n",
     "    pm.plotting.main_sequence_plot(\n",
     "        event_df,\n",
+    "        fit=True,\n",
     "        title='Main sequence plot for '\n",
     "        f'text {event_df[0, \"text_id\"]}, '\n",
```
dkrako: since merging #1365 `text_id` is not available anymore as a column. it will be added to a proper metadata field soon. you can just remove this line, as in this example `text_id` was always zero anyway.

src/pymovements/plotting/main_sequence_plot.py
```diff
@@ -163,6 +178,40 @@ def main_sequence_plot(
...
+    # --- Linear fit (only if requested) ---
+    if fit:
+        a, b = np.polyfit(amplitudes, peak_velocities, 1)
+
+        min_ampl, max_ampl = min(amplitudes), max(amplitudes)
+        line_x = [min_ampl, max_ampl]
+        line_y = [a * min_ampl + b, a * max_ampl + b]
+
+        line_axes = plt.gca() if own else ax
+        line_axes.plot(line_x, line_y, c=fit_color)
```
cbueth: The suggested solution from the meeting:
Rather use this
```suggestion
        line_axes.plot(line_x, line_y, c=fit_color, label=label)
```
with the `label` you define later and activate it with `ax.legend()` afterwards. Then you can remove the manual `line_axes.text(...)` :)

dkrako: how do we want to treat the case of `fit_measure == False`?

1. use `None` as the `label`: simplest approach, wouldn't create a legend box
2. use something like `"linear fit"` as label, would always create a legend box if `fit==True`, users would need to call `ax.get_legend().remove()` to remove the legend themselves (or add a `legend: bool = True` argument) 

I think (1.) is perfectly fine for this PR. (2.) might be valuable sometime in the future when we provide also non-linear fit methods but not that important currently.

src/pymovements/plotting/main_sequence_plot.py
```diff
@@ -55,10 +61,19 @@ def main_sequence_plot(
     ----------
     events: Events | EventDataFrame | None
         It must contain columns "peak_velocity" and "amplitude".
+    fit: bool
+        Draw a linear fit line if True. If False, no line is drawn.
+    measure: bool | Literal['r2', 's']
+        Annotate a goodness-of-fit statistic:
+        - True or 'r2': coefficient of determination (R²)
+        - 's': standard error of the regression (S)
```
dkrako: as some people might not know `S`, could you add the info here that it is the root-mean-squared-error (rmse)?

dkrako: ah sorry, `S` isn't exactly the `RMSE`, so just ignore my comment

src/pymovements/plotting/main_sequence_plot.py
cbueth: Good point, I'd prefer (2) and always having the legend, as the green datapoints should also get a label, in my opinion.

dkrako: Maybe we can do this in a separate PR then, as I think (2) would need a `legend: bool = True` argument because calling `ax.get_legend().remove()` would be somewhat annoying.

So I would be in favor of implementing (1) in this PR, and (2) in a follow-up that adds the `legend` argument to the function.

dkrako (approved):
Excellent that you also checked the existence of the fit line in the plot. I made some minor changes and I think we're ready to merge.

PR #1368, opened 2025-11-12
docs: remove message note from CITATION.cff

dkrako:
The message is not displayed by GitHub and on zenodo this just looks strange:

image


cbueth:
We should choose a fitting test for the message field, if it is required, as you said. Looking into some random repositories, it occurs to me many have kept exactly this line. I do agree that this note seems a bit displaced on the Zenodo page, and makes more sense in the file itself.

PR #1374, opened 2025-11-13
ci: make nightly builds and remove upper version bounds

dkrako:
## Description

Initially we added upper version bounds to make sure that breaking changes and deprecations in dependencies do not lead to failing tests in unrelated PRs. However, upper version bounds can be very annoying for users and require constant file updates on our side (luckily dependabot takes care of this).

This PR removes all upper version bounds and adds a nightly build schedule.

In case a build fails, maintainers will get notified. In case a fix is not that quick, we can then simply add an upper version bound for a specific dependency until we made the respective changes.

## Implemented changes

- [x] add nightly build schedule at 3:30 UTC
- [x] remove all upper version bounds

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,12 +55,13 @@ docs/source/tutorials/data
 docs/jupyter_execute
 
 # Testing
-.tox
 .coverage
 .coverage.*
+.DS_Store
+.tox
 benchmark-data.json
 coverage.xml
 htmlcov/
+tests/events
 tests/preprocessed/
 toy_dataset/
-.DS_Store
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -31,33 +31,33 @@ classifiers = [
...
-  "matplotlib>=3.8.0,<3.11",
-  "fastexcel>=0.13.0,<1.0",
-  "numpy>=1.22.4,<2.3",
-  "pandas>=2.1.4,<2.4",
+  "deprecated>=1.2.18",
+  "matplotlib>=3.8.0",
+  "fastexcel>=0.13.0",
+  "numpy>=1.22.4",
+  "pandas>=2.1.4",
   "polars>=1.31.0,<1.32",
```
dkrako, line 39: this upper version bound is actually necessary since #1199

dkrako: created #1382 for this

PR #1391, opened 2025-11-21
chore: migrate to `polars>=1.34`

### LINKED ISSUE #1382: add support for polars >=1.32
## Current Behavior
The dependabot PR #1199 failed a few months ago and we (or at least I) forgot about that later on.

Luckily #1374 exposed this when removing upper version bounds.

## Minimum acceptance critera
- [ ] passing tests with polars >=1.32

## Context

Here's the errors that I got from #1374 (the outputs from #1199 aare not available any more): [job-logs.txt](https://github.com/user-attachments/files/23631630/job-logs.txt)

cbueth:
## Description

Polars was set to `<1.32` before #1199, but never released. We do not want to keep the upper limit of the package version, to not block users.

## Implemented changes

- [x] Remove upper `polars` limit
- [x] Test unit tests, summarised in https://github.com/pymovements/pymovements/issues/1382#issuecomment-3557485000, fixing them all.
- [x] Modify functional tests and cast numeric columns before interpolation in `resample()` in eed975d.

## How Has This Been Tested?

- [x] Add coverage, pass all unit tests locally.

## Type of change

Remove irrelevant items:
- Internal

## Context

Resolves #1382

#### related issues:
- #1199

#### required by:
- #

#### requires:
- [ ] #

#### conflicts with:
- #

## Checklist:

### DIFF
--- a/tests/unit/gaze/io/csv_test.py
+++ b/tests/unit/gaze/io/csv_test.py
@@ -272,9 +272,7 @@ def test_from_asc_parameter_is_deprecated(
     )
 
 
-pytest.mark.filterwarnings('ignore:Gaze contains samples but no components could be inferred.')
-
-
+@pytest.mark.filterwarnings('ignore:Gaze contains samples but no components could be inferred.')
 def test_from_csv_decimal_overrides_with_precision_and_scale(tmp_path):
     p = tmp_path / 'mini.csv'
     p.write_text('time,pupil\n0,1.23\n1,4.56\n')


src/pymovements/gaze/io.py
```diff
@@ -288,9 +288,19 @@ def from_csv(
             ])
 
     if column_schema_overrides is not None:
+        # Normalise Decimal dtype without precision/scale for newer Polars versions
+        normalised_overrides: dict[str, Any] = {}
+        for fileinfo_key, fileinfo_dtype in column_schema_overrides.items():
+            # Handle bare pl.Decimal without precision/scale
+            if fileinfo_dtype is pl.Decimal:  # type: ignore[comparison-overlap]
```
dkrako: would this type error be fixed if the type of `column_schema_overrides` would be extended to `dict[str, type | pl.DataType] | None`?

tests/unit/gaze/transforms/deg2pix_test.py
```diff
@@ -518,7 +518,10 @@ def test_deg2pix_returns(kwargs, series, expected_df, distance_as_column):
     result_df = df.select(
         pm.gaze.transforms.deg2pix(**kwargs),
     )
-    assert_frame_equal(result_df, expected_df.to_frame(), atol=1e-4)
+    try:
+        assert_frame_equal(result_df, expected_df.to_frame(), abs_tol=1e-4)
+    except TypeError:
```
dkrako: this is because the `atol` argument was renamed in `1.32.3`, right? Strange that it wasn't considered a breaking change. maybe because it's just the testing module?

Alternatively we could also set polars to `>=1.32.3` but it's also fine how you did it.

src/pymovements/gaze/io.py
```diff
@@ -288,9 +288,19 @@ def from_csv(
...
+                # Choose a conservative default; adjust if needed by callers
```
dkrako: what is meant by `adjust if needed by callers`?

src/pymovements/gaze/transforms.py
```diff
@@ -803,13 +810,31 @@ def resample(
...
 
-    # Replace pre-existing null values with NaN as they should not be interpolated
+    # Helper: determine numeric columns among the target columns
+    def _is_numeric_dtype(dtype: pl.DataType) -> bool:
+        base_dtype = dtype.inner if isinstance(dtype, pl.List) else dtype
+        return (
+            base_dtype in {
+                pl.Int8, pl.Int16, pl.Int32, pl.Int64,
+                pl.UInt8, pl.UInt16, pl.UInt32, pl.UInt64,
+                pl.Float32, pl.Float64,
```
dkrako: all of these types are subclasses of [`pl.NumericType`](https://github.com/pola-rs/polars/blob/ca54d5c8ba19264f8ea2cc3ee36536895fa095e1/py-polars/src/polars/datatypes/classes.py#L286). It would be more future-proof to just check is `issubclass(base_dtype, pl.NumericType)`.

src/pymovements/gaze/transforms.py
```diff
@@ -803,13 +810,31 @@ def resample(
...
+    # interpolated. Ensure we cast to Float64 so the UDF output dtype matches the declaration.
+    if numeric_columns:
         samples = _apply_on_columns(
             samples,
-            columns=columns,
-            transformation=lambda series: series.fill_null(np.nan),
+            columns=numeric_columns,
+            transformation=lambda series: series.cast(pl.Float64).fill_null(np.nan),
             n_components=n_components,
+            return_dtype=pl.Float64,
```
dkrako, line 834: what happens if an int column is resampled that way? would it be casted to float on `resample()`?

src/pymovements/gaze/transforms.py
```diff
@@ -972,11 +999,22 @@ def smooth(
                 **pad_kwargs,
             )
 
+            # Wrap to ignore any unexpected keyword arguments Polars may pass
+            def _wrapped_pad(x: np.ndarray, **_: Any) -> np.ndarray:
+                return pad_func(x)  # pragma: no cover
```
dkrako: I'm confused. why isn't this covered despite being used in `.map_batches()` below?

src/pymovements/gaze/transforms.py
```diff
@@ -711,9 +711,16 @@ def savitzky_golay(
         cval=constant_value,
     )
 
+    # Wrap function to ignore unexpected keyword arguments possibly passed by Polars
```
dkrako: I'm a bit confused on the issue with these unexpected keyword arguments. Are they passed by polars or by pymovements?

src/pymovements/gaze/io.py
cbueth: You mean the initial error without the new additions?

src/pymovements/gaze/io.py
cbueth: This is relevant if someone is using a `pl.Decimal` (no `()`) without specifing the precision and scale. Then, the default I choose `pl.Decimal(38, 10)` is the decimal with the highest precision (I figured), see in the current implementation https://github.com/pola-rs/polars/blob/py-1.35.2/py-polars/src/polars/datatypes/classes.py#L417-L460.
If we want to work with something else by default, then this would need to be changed Maybe `pl.Decimal()` would also default to this. I am not sure if we need this high precision by default. Let me know what you think.

src/pymovements/gaze/transforms.py
cbueth: Nice find, sure! I was looking for this, but only found `Series.is_numeric()` and did not find it useful here.

src/pymovements/gaze/transforms.py
cbueth: I figured that this is already happening, as in https://github.com/pymovements/pymovements/blob/fde24acef50fe168f5f0b9917444fb88312f40bc/tests/unit/gaze/transforms/resample_test.py#L278-L286, but I see this is just if the type is given.
This would need a specific test, but I do assume that it would be casted, as 
```
base_dtype in {
                pl.Int8, pl.Int16, pl.Int32, pl.Int64,
```
is probably `true`.

I suppose that ints should not be casted if not specifically given in the schema?

src/pymovements/gaze/transforms.py
cbueth: For some reason testing with coverage it was not marked. I assumed coverage is not able to follow through the batched mapper.

tests/unit/gaze/transforms/deg2pix_test.py
cbueth: Right! I thought as it's just in few spots, adding this differentiation is fine. Otherwise, we could add backwards compatibility ourselves.

src/pymovements/gaze/transforms.py
cbueth: This relates to the Polars UDF Return Type Inference Errors mentioned in https://github.com/pymovements/pymovements/issues/1382#issuecomment-3557485000. It helped with it, as it seemed the mapping passes extra kwargs that were causing some issues. But I removed it in 417d0512 and tests run fine.

tests/unit/gaze/_utils/_parsing_eyelink_test.py
```diff
@@ -256,8 +256,13 @@ def test_parse_eyelink(make_text_file):
         messages=True,
     )
 
-    assert_frame_equal(gaze_df, EXPECTED_GAZE_DF, check_column_order=False, rtol=0)
-    assert_frame_equal(event_df, EXPECTED_EVENT_DF, check_column_order=False, rtol=0)
+    try:
+        assert_frame_equal(gaze_df, EXPECTED_GAZE_DF, check_column_order=False, rel_tol=0)
```
dkrako: I would very much prefer to set the minimum boundary to `polars>=1.32.3` (or `1.33.0`) instead and remove the try/except

src/pymovements/gaze/io.py
dkrako: This is the current type: `column_schema_overrides: dict[str, type] | None`

mypy complains about this if-clause not being possible to reach if we conform to this typing.

I thought this would be resolved by updating the type to `dict[str, type | pl.DataType] | None`.

But actually only this is True: `isinstance(pl.Decimal(), pl.DataType)`, not `isinstance(pl.Decimal, pl.DataType)`. So updating the type won't change anything.

The more I think about this the more I would like to remove these implicit instantiations of `pl.Decimal`, as having an uninstantiated `pl.Decimal` in the passed schema doesn't really conform to the typing of the argument.

Let's discuss this in today's meeting.

src/pymovements/gaze/transforms.py
```diff
@@ -803,13 +806,31 @@ def resample(
     # Sort columns by datetime
     samples = samples.sort('datetime')
 
-    # Replace pre-existing null values with NaN as they should not be interpolated
+    # Helper: determine numeric columns among the target columns
+    def _is_numeric_dtype(dtype: pl.DataType) -> bool:
+        base_dtype = dtype.inner if isinstance(dtype, pl.List) else dtype
```
dkrako: this would break in case of an array. Probably it's best to apply the function recursively if there exists the inner attribute:

```python
if hasattr(dtype, 'inner'):
    return _is_numeric_dtype(dtype.inner)
```

src/pymovements/gaze/transforms.py
dkrako: I see I think it's fine for this PR as it really doesn't matter much

src/pymovements/gaze/transforms.py
dkrako: I'm still a bit confused why ignoring the unexpected keyword arguments is needed. How can this happen?

Will this still be required if we update the polars minimum version?

dkrako (changes requested):
Thank you so much for your initiative. I have some minor comments:

- [x] change the polars dependency to "polars>=1.32.3" and remove the try/except blocks from the tests.

Further I would prefer these changes but I would like to discuss with you first:

- [x] remove instantiation of `pl.Decimal`
- [ ] remove wrapped pad that ignores keyword aguments

tests/unit/gaze/transforms/deg2pix_test.py
cbueth: ToDo: raise min version and remove if-else with `atol`/`abs_tol`

src/pymovements/gaze/transforms.py
cbueth: With having the newest polars version, like this
https://github.com/pymovements/pymovements/blob/da8d64afcb508ddc1719b1286bdc96b9c321a509/src/pymovements/gaze/transforms.py#L970-L985

we get ```polars.exceptions.InvalidOperationError: UDF called without return type, but was not able to infer the output type. This used to be allowed but lead to unpredictable results. To fix this problem, either provide a return datatype or execute the UDF in an eager context (e.g. in `map_columns`).``` for the ` tests/unit/gaze/transforms/smooth_test.py::test_smooth_returns`. Maybe we should rather use https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.map_columns.html.

What I decided to do is to use a `return_dtype` to https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.map_batches.html. And if the function does not ignore extra kwargs, then `pad` will get this keyword argument `return_dtype` destined for `map_batches` as well, resulting in a `ValueError: unsupported keyword arguments for mode 'reflect': {'return_dtype'}`. This is the reason why a wrapper is needed there. I removed the use of `partial` now.

Do you think it is possible to reduce the huge expressions to use [`polars.DataFrame.map_columns`](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.map_columns.html)?

tests/unit/gaze/_utils/_parsing_eyelink_test.py
cbueth: Done in 2836a5d.

src/pymovements/gaze/transforms.py
cbueth: Solved in 
https://github.com/pymovements/pymovements/blob/52b64e0f91c97e6588203ac8a7f8a093c356706e/src/pymovements/gaze/transforms.py#L812-L825

tests/unit/gaze/transforms/deg2pix_test.py
dkrako: Done in [`367231f`](https://github.com/pymovements/pymovements/pull/1391/commits/367231f4ec7c1ea4a7af73cdceb6565ce580ee36)

dkrako (approved):
Thanks again for taking care!

PR #1393, opened 2025-11-24
test: centralise test figure management

### LINKED ISSUE #1330: use non-interactive mode in plotting tests
## Description of the problem
Currently we use 

This problem already lead to this dirty workaround here in source code:

https://github.com/pymovements/pymovements/blob/39fd87cd460c1193c794505381c00b8587b4be41/src/pymovements/plotting/scanpathplot.py#L40-L44

That codeblock should be removed accordingly.

## Description of a solution

@cbueth has already proposed an approach in #1309:

```python
import matplotlib
import pytest

# Plotting backend
matplotlib.use('Agg')  # Non-interactive plotting

@pytest.fixture(scope='session', autouse=True)
def setup_matplotlib():
    import matplotlib.pyplot as plt
    plt.ioff()  # Turn off interactive mode
```

- [ ] get rid of workaround codeblock in `scanpathplot.py`
- [ ] use non-interactive plotting backend for all plotting tests

cbueth:
## Description

The figure management has been centralised for tests. Before, a unified autoclose has been implemented in #1309.
Plotting backend code for tests still remained in source. This PR consolidates this into a fixture, pulling the autoclose to the same spot.

A design decision remains open to discussion: Should `fixture_close_figures_after_test` be autoused, or should we add it manually to each plotting function? Contra: then we could just use `plt.close()` anyway. Pro: Otherwise it will also be called for each function that is not related to plotting. But: Having the autoclose in the plotting submodule, as before, it should only `autouse=True` should only apply to these. Can one confirm this ([Autouse Doc](https://docs.pytest.org/en/stable/how-to/fixtures.html#autouse-fixtures-fixtures-you-don-t-have-to-request))?

## Implemented changes

- [x] Move autoclose fixture into dedicated fixture file, inside the fixtures folder.
- [x] Make matplotlib use non-interactive backend in this plotting fixture file.
- [x] Remove test backend code from source.
- [x] Remove remaining `.close()` from plotting functions, where not relevant for testing.

## How Has This Been Tested?

- [x] Test different combinations and solutions manually.
- [x] Session level `verify_no_figures_remain` fixture verifies all figures have been closed.

There is no test file `plotting_fixtures_test`, because `fixture_close_figures_after_test()` cannot be tested with a test using this function, as it is activated only after a test finishes (after `yield`). Thus, we have `verify_no_figures_remain`.

## Type of change

Remove irrelevant items:
- Internal

## Context

Resolves #1330

#### related issues:
- #1309

#### required by:
- #

#### requires:
- [ ] #

#### conflicts with:
- #

## Checklist:


src/pymovements/plotting/scanpathplot.py
```diff
@@ -39,12 +38,6 @@
 from pymovements.plotting._matplotlib import finalize_figure
 from pymovements.plotting._matplotlib import LinearSegmentedColormapType
 
-# This is really a dirty workaround to use the Agg backend if runnning pytest.
-# This is needed as Windows workers on GitHub fail randomly with other backends.
-# Unfortunately the Agg module cannot show plots in jupyter notebooks.
-if 'pytest' in sys.modules:  # pragma: no cover
```
dkrako, line 45: great, finally this is removed! 🎉

tests/fixtures/plotting_fixtures.py
```diff
@@ -17,17 +17,26 @@
...
-
+"""Provide fixtures for non-interactive plotting tests."""
 from collections.abc import Generator
 
+import matplotlib
 import matplotlib.pyplot as plt
 import pytest
 
+# Plotting backend
+matplotlib.use('Agg')  # Non-interactive plotting
```
dkrako: This would be a bit cleaner as a fixture. Maybe in the future we might add more logic to setup test plotting. Using a `@pytest.fixture(scope='session', autouse=True)` will make sure that it'll only run once.

tests/fixtures/plotting_fixtures.py
```diff
@@ -17,17 +17,26 @@
...
     """Automatically close all Matplotlib figures after each test."""
     yield
     plt.close('all')
+
+
+@pytest.fixture(scope='session', autouse=True)
+def verify_no_figures_remain():
+    """Verify no figures remain open after all tests."""
+    yield
+    assert len(plt.get_fignums()) == 0, 'Some figures were not closed!'
```
dkrako, line 45: nice!

tests/fixtures/plotting_fixtures.py
cbueth: Sure! See 6895a6c1.

tests/fixtures/plotting_fixtures.py
cbueth: I figured this is the best way, as we cannot really test `fixture_close_figures_after_test`.

PR #1410, opened 2025-12-10
docs: add link to dataset tutorial on datasets page

saeub:
Maybe this would help non-pymovements-users find the instructions on how to add their own datset?

### DIFF
--- a/src/pymovements/dataset/dataset.py
+++ b/src/pymovements/dataset/dataset.py
@@ -34,6 +34,7 @@ from pymovements._utils._html import repr_html
 from pymovements.dataset import dataset_download
 from pymovements.dataset import dataset_files
 from pymovements.dataset.dataset_definition import DatasetDefinition
+from pymovements.dataset.dataset_files import DatasetFile
 from pymovements.dataset.dataset_library import DatasetLibrary
 from pymovements.dataset.dataset_paths import DatasetPaths
 from pymovements.events import Events
@@ -67,6 +68,7 @@ class Dataset:
             path: str | Path | DatasetPaths,
     ):
         self.fileinfo: pl.DataFrame = pl.DataFrame()
+        self._files: list[DatasetFile] = []
         self.gaze: list[Gaze] = []
         self.precomputed_events: list[PrecomputedEventDataFrame] = []
         self.precomputed_reading_measures: list[ReadingMeasures] = []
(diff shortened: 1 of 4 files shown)


docs/source/datasets/index.rst
```diff
@@ -29,6 +29,22 @@ Download your dataset and load it into memory with these few lines of code:
...
 
+.. grid::
+    :gutter: 5
+
+    .. grid-item-card::
+
+        **Is your dataset missing?**
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+        Check out this tutorial on how to add your dataset to the library:
```
dkrako:
```suggestion
        Check out this guide on how to contribute datasets to the library:
```

docs/source/datasets/index.rst
```diff
@@ -29,6 +29,22 @@ Download your dataset and load it into memory with these few lines of code:
    :widths: 45, 5, 5, 5, 5, 5, 5, 5
    :header-rows: 1
 
+.. grid::
+    :gutter: 5
+
+    .. grid-item-card::
+
+        **Is your dataset missing?**
```
dkrako:
```suggestion
        **Are you missing a dataset?**
```

dkrako (changes requested):
Great idea! I would be in favor of putting the box more prominently right before the Public Datasets table.

docs/source/datasets/index.rst
```diff
@@ -24,6 +24,22 @@ Download your dataset and load it into memory with these few lines of code:
...
+        **Are you missing a dataset?**
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+        Check out this guide on how to contribute datasets to the library:
+
+        .. button-link:: ../tutorials/adding-public-dataset.html
+            :color: primary
+            :shadow:
+
+            Adding a Public Dataset to the Library
```
cbueth: There should be a way to replace this relative link with a header target. I think it would need the target in place of the relative url and an additional [`:ref-type: ref`](https://sphinx-design.readthedocs.io/en/latest/badges_buttons.html#button-link-and-button-ref-options) in the botton option.

This card uses the sphinx-design package, see the button doc here:
https://sphinx-design.readthedocs.io/en/latest/badges_buttons.html

```suggestion
        .. button-link:: name_of_the_desired_target
            :color: primary
            :shadow:
            :ref-type: ref

            Adding a Public Dataset to the Library
```

saeub:
Thanks @dkrako and @cbueth, I think I fixed everything now.

docs/source/datasets/index.rst
saeub: Turns out the `:ref-type: ref` is not necessary, but the `button-link` has to be turned into a `button-ref`.

PR #1412, opened 2025-12-10
docs: create contributing guide landing page

dkrako:
## Description

Create new landing page that lists all contributing guides.

Most content is copied from the `CONTRIBUTING.md`. I guess some of the sections could be improved or extended in the future but that's mostly out of scope of this PR.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -2,7 +2,7 @@ name: benchmark
 
 on:
   pull_request:
-    branches: [main]
+    branches: [ main ]
   workflow_call:
   workflow_dispatch:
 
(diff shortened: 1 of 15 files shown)


docs/source/contributing/code-of-conduct.rst
```diff
@@ -0,0 +1,3 @@
+.. _code-of-conduct:
+
+.. mdinclude:: ../../../CODE_OF_CONDUCT.md
```
cbueth, line 3: There is a small displaying problem when including this file:

image

The `&nbsp;` is not digested well from the raw, see

https://github.com/pymovements/pymovements/blob/119ba45676a5aa0cbbf75f413e77255b67a670c8/CODE_OF_CONDUCT.md?plain=1#L43

So replacing it with a normal space would be the dirty solution.

Issue #1042, opened 2025-03-14
support recomputing event properties


dkrako:
## Description of the problem
If a user calls `compute_event_property()` for a property that already exists in the event data frame, a confusing error is raised (see #1037).

## Description of a solution

Instead of raising an exception we could also support recalculating the event property.

## Context

Related issues:
- #1041
- #1043

PR #1419, opened 2025-12-16
chore!: unify measures module

### LINKED ISSUE #1400: get rid of circular dependency in `events.processing`
## Description of the problem

The current implementation of processing events is not optimal as it creates a circular dependency: `EventGazeProcessor` depends on `Gaze` which in turn depends on `EventGazeProcessor`.

This is the current signature of its processing method:

https://github.com/pymovements/pymovements/blob/d13ee90ed96d3d4d5a1fa48d9768301d5ca1255e/src/pymovements/events/processing.py#L128-L134

## Description of a solution

If all the arguments that require `pymovements` objects (`Gaze`, `Events`) would be replaced with arguments that require a `polars.DataFrame`, the circular dependency would be resolved:

```python
    def process(
            self,
            events: pl.DataFrame,
            samples: pl.DataFrame,
            groupby: str | list[str] | None = None,
            name: str | None = None,
    ) -> pl.DataFrame:
```

Removing the dependencies on the `pymovements` classes shouldn't be much of an issue as the function only uses `Events.frame` and `Gaze.samples`.

Moreover, the `EventGazeProcessor` should then also be renamed to `EventSampleProcessor` for a clearer naming.

Getting rid of this circular dependency would also help with testing, as `polars.DataFrames` can be used directly without the additional need of instantiating `pymovements` objects.

As I want to get rid of this circular dependency as soon as possible, and as the current `EventGazeProcessor` is probably only used internally by `Gaze.compute_event_properties()`, I would like to make this change without deprecating first.

As this then would be a breaking change anyway, I would propose moving all processing from `events` to `measure`:
- `events.processing` -> `measure.event_processing`
- `events.properties` -> `measure.event_measures`

Later on I would like to create `Gaze.measure_events()` and deprecate `Gaze.compute_event_properties()`.

- [ ] replace `pymovements` dependencies in `EventGazeProcessor` with `polars.DataFrame`
- [ ] rename `EventGazeProcessor` to `EventSampleProcessor`

## Optional acceptance criteria
- [ ] move `events.processing` to `measure.event_processing` 
- [ ] move `events.properties` to `measure.event_measures`
- [ ] deprecate `Gaze.compute_event_properties()` in favor of `Gaze.measure_events()`

## Context

- resolves #559

dkrako:
## Description

This PR restructures and unifies all measure functionality across the pymovements package into a single `measure` module with three submodules: `samples`, `events`, `reading`.

These characteristics should hold true for all measures in these submodules:

- on a general level, measures take an input (might be one or multiple series) and return a single value (might be grouped and formated as a dataframe)
- the actual implementations do not use any pymovements dataclasses (Gaze, Events, Stimulus) and purely rely on polars dataframes. this way we create a layered architecture and get rid of cyclical imports.

The individual submodules differ in terms of their input data:

- `samples`: these measures are applied on a series of samples and return a single value. before, we only had the `null_ratio` available as an explicit sample measure. This PR moves all event sample measures that were previously called event properties and defined in `events.properties` to `measures.samples` as they are basically the same thing: they are a measure that is applied on a series of samples and return a single value.

- `events`: here we have to discern between two operational types of event measures: measures that are applied to the samples of an event (e.g., `amplitude` or `location`) and measures that are applied to already existing measures. The latter were formerly named event properties and currently the only implemented measure in this category is `duration`. The measured that were defined in `events.properties` were all moved to `measure.samples`. Applying sample measures to events is achieved via the `EventSampleProcessor` (formerly named `EventGazeProcessor`).

- `reading`: this submodule currently doesn't provide any reading measures at all, just a `ReadingMeasures` class that's nothing more than a polars dataframe. I think we can discuss if we still need that class as the concept is a different here: here the reading measures relate to the output of the reading measure functions. I think it's fine to just return polars data frames by measure methods so I would probably get rid of that class soon.

If we also include aoi measures on a more general level than just reading measures, I would propose to create a `measure.interest` submodule for that in the future.

## Implemented changes

- [x] rename `InvalidProperty` exception to `UnknownMeasure`
- [x] moved `events/processing.py` to `measure/events/processing.py`
- [x] changed `EventProcessor` to not have dependency on `Events`
- [x] rewrote `EventGazeProcessor` as `EventSampleProcessor` without dependencies on `Gaze` and `Events`
- [x] use `measures` instead of `event_properties` in `EventProcessor` and `EventSampleProcessor`
- [x] remove circular dependencies in `gaze/gaze.py` and `events/processing.py`
- [x] moved `measure/library.py` and `measure/measures.py` to `measure/samples` submodule
- [x] moved `reading_measures` to `measure/reading`
- [x] add name alias to all measures
- [x] warn on column overwrite in `compute_event_properties()` instead of raising exception

## How Has This Been Tested?

- [x] restructured and adjusted existing tests
- [x] test that `compute_event_properties()` warns on column overwrite

## Context

Resolves #1400 #1042 #559

#### conflicts with:
- #1420

## Checklist:

### DIFF
--- a/src/pymovements/gaze/gaze.py
+++ b/src/pymovements/gaze/gaze.py
@@ -1151,12 +1151,13 @@ class Gaze:
 
         join_on = identifiers + ['name', 'onset', 'offset']
         column_intersection = set(self.events.columns) & set(results.columns)
-        if column_intersection != set(join_on):
+        overwrite_columns = list(column_intersection - set(join_on))
+        if overwrite_columns:
             warn(
                 'The following columns already exist in event and will be overwritten: '
-                f"{list(column_intersection - set(join_on))}",
+                f"{overwrite_columns}",
             )
-
+            self.events.drop(overwrite_columns)
         if results.height:
             self.events.add_event_properties(results, join_on=join_on)
 
--- a/tests/unit/gaze/gaze_test.py
+++ b/tests/unit/gaze/gaze_test.py
@@ -1118,6 +1118,41 @@ def test_gaze_compute_event_properties_no_events():
 
 
 @pytest.mark.parametrize(
+    ('existing_amplitude', 'expected_amplitude'),
+    [
+        pytest.param(0.0, np.sqrt(32), id='overwrite_zero'),
+        pytest.param(123.0, np.sqrt(32), id='overwrite_nonzero'),
+    ],
+)
+def test_gaze_compute_event_properties_overwrites_column(existing_amplitude, expected_amplitude):
+    gaze = Gaze(
+        samples=pl.DataFrame({
+            'time': [0, 1, 2, 3, 4],
+            'position': [[0, 0], [1, 1], [2, 2], [3, 3], [4, 4]],
+        }),
+        events=Events(
+            pl.DataFrame({
+                'name': ['fixation'],
+                'onset': [0],
+                'offset': [4],
+                'amplitude': [existing_amplitude],
+            }),
+        ),
+    )
+
+    expected_events = gaze.events.frame.with_columns(pl.lit(expected_amplitude).alias('amplitude'))
+
+    with pytest.warns(
+            UserWarning,
+            match='The following columns already exist in event and will be overwritten: '
+                  r'\[\'amplitude\'\]',
+    ):
+        gaze.compute_event_properties('amplitude')
+
+    assert_frame_equal(gaze.events.frame, expected_events, check_column_order=False)
+
+
+@pytest.mark.parametrize(
     ('gaze', 'attribute'),
     [
         pytest.param(


src/pymovements/events/processing.py
dkrako, line 1: this is moved to `measure/events/processing.py`

src/pymovements/measure/events/measures.py
dkrako, line 1: content from `events/properties.py`

src/pymovements/measure/events/measures.py
```diff
@@ -0,0 +1,59 @@
...
+
+    The duration is defined as the difference between offset time and onset time.
+
+    Returns
+    -------
+    pl.Expr
+        The duration of the event.
+    """
+    result = pl.col('offset') - pl.col('onset')
+    return result.alias('duration')
```
dkrako, line 59: all measures now use an alias according to their name

src/pymovements/measure/events/processing.py
```diff
@@ -0,0 +1,258 @@
...
+        for measure_name, _ in measures_with_kwargs:
+            if measure_name not in SampleMeasureLibrary.measures:
+                known_measures = list(SampleMeasureLibrary.measures.keys())
+                raise UnknownMeasure(
+                    measure_name=measure_name, known_measures=known_measures,
+                )
+
+        self.measures: list[pl.Expr] = [
+            # initialize measure functions to create polars expressions.
+            SampleMeasureLibrary.get(measure_name)(**measure_kwargs)
```
dkrako, line 126: the measure expressions are now constructed at initialization.

src/pymovements/measure/measures.py
dkrako, line 1: moved to `measure/samples/measures.py`

tests/unit/events/properties_test.py
dkrako, line 1: content moved to `tests/unit/measure/events/measures` and `tests/unit/measure/samples/measures`

src/pymovements/gaze/gaze.py
```diff
@@ -1111,36 +1114,37 @@ def compute_event_properties(
...
         )
+
         join_on = identifiers + ['name', 'onset', 'offset']
-        self.events.add_event_properties(new_properties, join_on=join_on)
+        column_intersection = set(self.events.columns) & set(results.columns)
+        if column_intersection != set(join_on):
+            warn(
+                'The following columns already exist in event and will be overwritten: '
+                f"{list(column_intersection - set(join_on))}",
+            )
```
cbueth: Reading this, I suppose the join does not overwrite the existing columns, but adds a suffix to the right column with identical name, happening in
https://github.com/pymovements/pymovements/blob/d92469b78bfe135666c4e3da05ca4c72d7088252/src/pymovements/events/events.py#L244

```suggestion
        overwrite_columns = list(column_intersection - set(join_on))
        if overwrite_columns:
            warn(
                'The following columns already exist in event and will be overwritten: '
                f"{overwrite_columns}",
            )
            self.events.drop(overwrite_columns)
```

As regression test, you might want to add to `gaze_test.py` something along the lines of:

```python
@pytest.mark.parametrize(
    ('existing_amplitude', 'expected_amplitude'),
    [
        pytest.param(0.0, np.sqrt(32), id='overwrite_zero'),
        pytest.param(123.0, np.sqrt(32), id='overwrite_nonzero'),
    ],
)
def test_gaze_compute_event_properties_overwrite(existing_amplitude, expected_amplitude):
    gaze = Gaze(
        samples=pl.DataFrame({
            'time': [0, 1, 2, 3, 4],
            'position': [[0, 0], [1, 1], [2, 2], [3, 3], [4, 4]],
        }),
        events=Events(pl.DataFrame({
            'name': ['fixation'],
            'onset': [0],
            'offset': [4],
            'amplitude': [existing_amplitude],
        })),
    )

    with pytest.warns(
        UserWarning,
        match=r'The following columns already exist in event and will be overwritten: \[\'amplitude\'\]',
    ):
        gaze.compute_event_properties('amplitude')

    assert np.isclose(gaze.events.frame['amplitude'][0], expected_amplitude)
    assert 'amplitude_right' not in gaze.events.columns
```

src/pymovements/measure/events/measures.py
cbueth: Confirmed 👍

src/pymovements/measure/events/processing.py
cbueth: Confirmed as well :+1:

src/pymovements/gaze/gaze.py
```diff
@@ -1111,36 +1114,37 @@ def compute_event_properties(
...
-        InvalidProperty
-            If ``property_name`` is not a valid property. See
-            :py:mod:`pymovements.events.event_properties` for an overview of supported properties.
+        UnknownMeasure
+            If ``event_properties`` includes an unknwon measure. See :ref:`sample-measures` and
+            :ref:`event-measures` for an overview of supported measures.
         RuntimeError
             If specified event name ``name`` is missing from ``events``.
         ValueError
             If the computed property already exists as a column in ``events``.
```
cbueth:
```suggestion
```

Might this be outdated, as `raise ValueError` has been deleted?

cbueth (changes requested):
The PR is an improvement to the package structure, improving on the modularity and resolving circular dependencies.

In the `Gaze.compute_event_properties()` method, existing columns were not actually being overwritten despite the warning (Polars join default behavior adds a suffix '_right', see [`polars.DataFrame.join`](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.join.html#polars.DataFrame.join)).

After this, the PR is ready to be accepted. Nice efforts on the restructuring and added tests!

src/pymovements/gaze/gaze.py
dkrako: you're right, good catch!

src/pymovements/gaze/gaze.py
dkrako: Thanks for the suggestion! implemented in [`0ba7183`](https://github.com/pymovements/pymovements/pull/1419/commits/0ba71837038a4ef70337546b8fd3b8d1e3c41bf6)

cbueth: Sweet, thanks!

PR #1420, opened 2025-12-16
feat: add sample measure `data_loss` with count, time, and ratio units

### LINKED ISSUE #1398: implement data loss ratio as a sample measure
## Description of the problem
Currently the data loss is only calculated inside the eyelink parsing. This has several shortcomings:

- conflates preprocessing with parsing
- needs reimplementation for each parser
- data loss is not available for csv files

## Description of a solution
It would be much better to implement the data loss ratio as a sample measure. Currently we have only a single one, the null ratio:

https://github.com/pymovements/pymovements/blob/b0212869857c7c7f3b66f6a82146b5eaf7fb5235/src/pymovements/measure/measures.py#L29-L60

The data loss should be implemented in a similar way and return a polars expression. To prevent circular dependencies the code should not rely on `Gaze` or `Events` and be on a more lower-level by using just polars functions at best.

The sample measures are then called by [`Gaze.measure_samples()`](https://github.com/pymovements/pymovements/blob/b0212869857c7c7f3b66f6a82146b5eaf7fb5235/src/pymovements/gaze/gaze.py#L1141-L1205)

Further there's the issue of the data loss unit. There are three options:
- ratio (relative to total duration)
- time unit (absolute duration)
- number of missing samples (absolute)

Although it would be great to provide both an absolute and relative measure, for a first iteration it would also be fine to provide just the absolute measure as the relative measure can just reuse the absolute measure logic (and be called `data_loss_ratio`).

We also need to work out how to provide the start and end recording times for calculating the data loss.
The current implementation in the eyelink parser uses the start and stop recording messages to calculate the total recording duration. I think this is the best approach in case these messages are available, as samples might be missing at the start or end of the recording.

At least for gaze sample csv files these message won't be available, so we would need to use the first and last timestamps as the start/stop recording timestamps.

I think it's best to implement the low-level function in such a way that it just gets the start and end timestep directly. Determining the start/end times and grouping by trials should then be done in `Gaze.measure_samples()` for now.

I would propose the following signature:

```python
def data_loss(column:str, start_time: float | None = None, end_time: float | None = None) -> pl.Expr:
```

Here `start_time` and `end_time` is optional: if these are `None`, just take the timestep of the first/last sample. (these arguments could also be left out in the first PR and add them later on when adding suport for recording timestamps.

- [ ] implement data loss as a function in `measure.measures` that returns a polars expression
- [ ] use first / last sample timestamp of each trial to identify start / stop times
- [ ] use recording timestamps to determine data loss

cbueth:
## Description

This PR introduces a low-level sample measure `data_loss` that computes missing samples in three units (absolute count, missing time, and ratio) using only Polars expressions. It is independent of `Gaze`/`Events` and intended to be aggregated in `Gaze.measure_samples()` per trial or other groupings. Default behaviour derives start/end anchors from first/last timestamps. Optional explicit `start_time`/`end_time` can be provided by higher layers (e.g., from recording start/stop messages).

Units as suggested from https://github.com/pymovements/pymovements/issues/1398#issue-3680907485:
  - `count`: integer number of missing samples.
  - `time`: approximate missing time = `missing_count * median_ISI`.
  - `ratio`: `missing_count / expected`.

## Implemented changes

- [x] `pm.measure.data_loss` in `src/pymovements/measure/measures.py`:
  - Signature: `data_loss(timestamp_col: str = 'time', *, start_time: float | None = None, end_time: float | None = None, unit: Literal['count','time','ratio']='count') -> pl.Expr`
  - Uses median inter-sample interval (ISI) of sorted timestamps to derive expected sample count over `[start, end]`.
  - Guards for degenerate ranges and groups with <2 samples (returns 0).
  - Returns aliased columns: `data_loss_count`, `data_loss_time`, `data_loss_ratio`.
  - Raises `ValueError` on invalid `unit`.
- [x] Registered the measure in the sample measure library and exported it via `pymovements.measure`.

## How Has This Been Tested?

- [x] By-design `tests/unit/measure/data_loss_test.py` with 13 scenarios and `unit in {'count','time','ratio'}`
- [x] Test `ValueError` for invalid `unit`

Passing locally.

## Open to discussion

- The behaviour on start and end times are gracious, no warning is given when they are inside the intervals, also when they are the same, the outputs are just 0, instead of giving an error.
- Choice of median inter-sample interval estimator: As of now, median of diffs is being used to infer the _optimal_ number of samples. Alternatives are: trimmed mean, mode, or using a known sampling rate if provided (e.g., optional `expected_isi` to bypass estimation).
- `expected = floor((end - start)/ISI) + 1`: does this inclusive endpoint convention fit our parsing?
- Any other suggestions for column names `data_loss_count`, `data_loss_time`, `data_loss_ratio`?
- If the input data is always sorted, we could drop the sorting step, but I prefer to keep it, keeping it agnostic. It should not produce a relevant overhead keeping it.
- Should this be added to tutorials?
- Can the start and stop messages mentioned in https://github.com/pymovements/pymovements/issues/1398#issue-3680907485 be hooked into this?

## Type of change

- New functionality

## Context

Resolves #1398

#### related issues:
- #

#### required by:
- #

#### requires:
- [ ] #

#### conflicts with:
- #

## Checklist:

### DIFF
--- a/src/pymovements/measure/samples/measures.py
+++ b/src/pymovements/measure/samples/measures.py
@@ -333,15 +333,6 @@ def null_ratio(column: str, column_dtype: pl.DataType) -> pl.Expr:
     pl.Expr
         Null ratio expression.
     """
-    if column_dtype not in {pl.Float64, pl.Int64, pl.Utf8, pl.List}:
-        # Note: pl.List is a class, but it can also be an instance pl.List(inner).
-        # We should check if it's the base class or an instance.
-        # However, the previous code used == pl.List which might have worked for base class.
-        # Actually, in Polars pl.List is a type constructor.
-        # Let's check if we should keep the same check logic.
-        pass
-
-    # To maintain consistency with the previous error message:
     valid_dtypes = {pl.Float64, pl.Int64, pl.Utf8, pl.List}
     if not any(
             column_dtype == d or (isinstance(column_dtype, pl.List) and d == pl.List)


src/pymovements/measure/measures.py
```diff
@@ -50,11 +52,111 @@ def null_ratio(column: str, column_dtype: pl.DataType) -> pl.Expr:
...
             'column_dtype must be of type {Float64, Int64, Utf8, List}'
             f' but is of type {column_dtype}',
         )
 
     return value.alias('null_ratio')
+
+
+@register_sample_measure
+def data_loss(
+        timestamp_col: str = 'time',
```
dkrako: let's keep it consistent with the `Gaze.__init__()` and name this `time_column`

src/pymovements/measure/samples/measures.py
```diff
@@ -50,11 +52,111 @@ def null_ratio(column: str, column_dtype: pl.DataType) -> pl.Expr:
...
+    pl.Expr
+        A scalar (per-group) expression.
+
+    Raises
+    ------
+    ValueError
+        If ``unit`` is not one of 'count', 'time', or 'ratio'.
+    TypeError
+        If ``timestamp_col`` is not a string.
+    """
```
dkrako, line 489: it would be great to have some examples here.

src/pymovements/measure/measures.py
```diff
@@ -50,11 +52,111 @@ def null_ratio(column: str, column_dtype: pl.DataType) -> pl.Expr:
...
+        If ``unit`` is not one of 'count', 'time', or 'ratio'.
+    TypeError
+        If ``timestamp_col`` is not a string.
+    """
+    if not isinstance(timestamp_col, str):
+        raise TypeError(
+            f'timestamp_col must be a string, but got: {type(timestamp_col).__name__}',
+        )
+
+    timestamps = pl.col(timestamp_col)
```
dkrako: let's use ducktyping here with a `try`/`except` block:

```suggestion
    try:
        timestamps = pl.col(time_column)
    except TypeError as e:
       raise TypeError(
           f"invalid type for 'time_column'. Expected 'str' , got '{type(time_column).__name__}'",
        ) from e
```

This has two benefits:

- we may adapt with less changes in case of a broader input typing of the called 3rd party function
- the error trace for the failing case is not obfuscated from the user
- there is no if condition inference (marginal in this case)

src/pymovements/measure/measures.py
```diff
@@ -50,11 +52,111 @@ def null_ratio(column: str, column_dtype: pl.DataType) -> pl.Expr:
...
+
+    # Determine group anchors: provided or derived.
+    start_expr = pl.lit(start_time) if start_time is not None else timestamps.min()
+    end_expr = pl.lit(end_time) if end_time is not None else timestamps.max()
+
+    # Observed samples in the group.
+    observed = pl.len()
+
+    # Median inter-sample interval (after sorting). Drop nulls before median.
+    isi_median = timestamps.sort().diff().drop_nulls().median()
```
dkrako: same as with `null_ratio` this should also include `drop_nans()` (e.g., `inf`)

src/pymovements/measure/measures.py
```diff
@@ -50,11 +52,111 @@ def null_ratio(column: str, column_dtype: pl.DataType) -> pl.Expr:
...
+        *,
+        start_time: float | None = None,
+        end_time: float | None = None,
+        unit: Literal['count', 'time', 'ratio'] = 'count',
+) -> pl.Expr:
+    """Measure gaze data loss within a group using only Polars expressions.
+
+    Computes the number of missing samples ("count"), missing time ("time"),
+    or the ratio of missing to expected samples ("ratio"). The expected
+    sampling is derived from the group's median inter-sample interval (ISI)
```
dkrako: I'm afraid that this introduces circular logic:

when you automatically derive the sampling rate from the samples, and then use this as a basis to calculate the expected samples, then, if you disregard null/nan values, it will always result in the top value (100%).

It is better to explicitly add `sampling_rate: float` as an argument to `data_loss()`. The way data loss is usually defined the measure can only be applied to evenly sampled time series (or rather to a time series that is expected to be evenly sampled).

A related measure to your inference logic is the *effective sampling rate* (N / delta time) which is also very useful but probably out of scope of this PR.

src/pymovements/measure/measures.py
```diff
@@ -50,11 +52,111 @@ def null_ratio(column: str, column_dtype: pl.DataType) -> pl.Expr:
...
+    start_expr = pl.lit(start_time) if start_time is not None else timestamps.min()
+    end_expr = pl.lit(end_time) if end_time is not None else timestamps.max()
+
+    # Observed samples in the group.
+    observed = pl.len()
+
+    # Median inter-sample interval (after sorting). Drop nulls before median.
+    isi_median = timestamps.sort().diff().drop_nulls().median()
+
+    # Expected sample count over [start, end] with inclusive endpoints.
```
dkrako: the data loss should not just be limited to the time column.

Let's take this samples data frame:

| time | pixel |
|------|-------|
| 1 | [111, 123] |
| 2 | [111, 123] |
| 3 | null |
| 4 | null |
| 5 | [111, 123] |
| 9 | [111, null] |

There are three instances of data loss: one through null values (timesteps 3 to 4), one through completely missing samples (timesteps 6 to 8) and the final one has just a single component missing (timestep 9).

So apart from the `time_column` argument this function also needs an optional argument with the column name that this should be applied on.

I think we can just name it `column: str | None = None`, same as with `null_ratio`.

dkrako (changes requested):
Awesome, thank you so much for your initiative! There are some things that need to be worked out still,.

Regarding some of your discussions points:

> The behaviour on start and end times are gracious, no warning is given when they are inside the intervals, also when they are the same, the outputs are just 0, instead of giving an error.

I think it's fine for now. We don't really know this during `data_loss()` as the function just returns a `polars.Expr` (which is then applied on the data). I currently do not know how to raise an exception within a polars expression.

> Choice of median inter-sample interval estimator: As of now, median of diffs is being used to infer the optimal number of samples. Alternatives are: trimmed mean, mode, or using a known sampling rate if provided (e.g., optional expected_isi to bypass estimation).

As written in a prior comment, I don't think that the sampling rate should be inferred. Calculating the effective sampling rate of samples is another valuable data loss measure method.

> Should this be added to tutorials?

Yes I think a measure tutorial should be added, but that's out of scope of this PR. However, a few examples in the `data_loss()` docstring would be great.

> Can the start and stop messages mentioned in
https://github.com/pymovements/pymovements/issues/1398#issue-3680907485 be hooked into this?

As the function already has the `start_time` and `end_time` arguments that shouldn't be a problem. I think the right place to take care of this is within the `Gaze.measure_samples()` method. That method should take care of putting the correct data from `Gaze` into `data_loss()`.

src/pymovements/measure/measures.py
dkrako: let's leave out the ` | None = None` and make both `column` and `time_column` a required argument.

that's more explicit and we have one less what-if to document and care about.

cbueth:
Thank you for the detailed feedback @dkrako! I have integrated all points raised.

I tried to keep the changes simple, and found using a [UDF](https://docs.pola.rs/user-guide/expressions/user-defined-python-functions/) `_is_invalid_value()` would be the simplest approach, also supporting list-like types, as in your example with the pixel values. About that one:

> Let's take this samples data frame:
> time 	pixel
> 1 	[111, 123]
> 2 	[111, 123]
> 3 	null
> 4 	null
> 5 	[111, 123]
> 9 	[111, null]
> 
> There are three instances of data loss: one through null values (timesteps 3 to 4), one through completely missing samples (timesteps 6 to 8) and the final one has just a single component missing (timestep 9).

The counting I implemented, counts 6 in total, one for each missing timestep, but in your comment this could also be understood as counting "one through null values (timesteps 3 to 4)" as one, not two. Which one is the desired solution?
Please look through the updated test cases to see if the requirements are fitting better now and let me know if there are any open changes :)

src/pymovements/measure/samples/measures.py
```diff
@@ -50,11 +54,197 @@ def null_ratio(column: str, column_dtype: pl.DataType) -> pl.Expr:
...
+    # Observed rows in the group
+    observed = pl.len()
+
+    # Expected sample count over [start, end] with inclusive endpoints for a fixed rate.
+    span = end_expr - start_expr
+    expected = (span * pl.lit(sampling_rate)).floor().cast(pl.Int64) + 1
+
+    # Missing rows due to time gaps, ensure non-negative and valid range
+    valid_range = end_expr > start_expr
+    time_missing = pl.when(valid_range).then(
```
dkrako, line 521: I guess we should raise an exception in case of `end_expr < start_expr` as that would mean something has gone very wrong and reporting zero here would lead to silent errors.

However, I don't think there's a way to raise an exception within a polars expression and we don't know the timestamp values when constructing the expressions, so I'm unsure how to resolve this best.

src/pymovements/measure/samples/measures.py
```diff
@@ -50,11 +54,197 @@ def null_ratio(column: str, column_dtype: pl.DataType) -> pl.Expr:
...
+    if unit == 'count':
+        return total_missing
+
+    if unit == 'time':
+        missing_time = (total_missing.cast(pl.Float64) / pl.lit(float(sampling_rate)))
+        return missing_time.alias('data_loss_time')
+
+    if unit == 'ratio':
+        # Prevent division by zero when range invalid - expected will be >= 1 when valid.
+        expected_safe = pl.when(valid_range).then(expected).otherwise(pl.lit(1))
```
dkrako: this then won't be necessary anymore if checking for valid range beforehand

src/pymovements/measure/samples/measures.py
```diff
@@ -50,11 +54,197 @@ def null_ratio(column: str, column_dtype: pl.DataType) -> pl.Expr:
...
+        ratio = (total_missing.cast(pl.Float64) / expected_safe.cast(pl.Float64)).fill_null(0.0)
+        return ratio.alias('data_loss_ratio')
+
+    raise ValueError(
+        "unit must be one of {'count', 'time', 'ratio'} but got: " f"{unit!r}",
```
dkrako:
```suggestion
        f"unit must be one of {'count', 'time', 'ratio'} but got: {unit!r}",
```

src/pymovements/measure/samples/measures.py
```diff
@@ -50,11 +54,197 @@ def null_ratio(column: str, column_dtype: pl.DataType) -> pl.Expr:
...
+    span = end_expr - start_expr
+    expected = (span * pl.lit(sampling_rate)).floor().cast(pl.Int64) + 1
+
+    # Missing rows due to time gaps, ensure non-negative and valid range
+    valid_range = end_expr > start_expr
+    time_missing = pl.when(valid_range).then(
+        pl.max_horizontal(expected - observed, pl.lit(0)),
+    ).otherwise(pl.lit(0))
+
+    def _is_invalid_value(v: Any) -> bool:  # pragma: no cover
```
dkrako: This out of scope of this PR, but it would be great to write a reuseable `is_valid` expression as the calculation in `null_ratio` is basically the same. This could then be properly tested and the coverage ignore could be removed.

tests/unit/measure/data_loss_test.py
```diff
@@ -0,0 +1,202 @@
...
+    df = pl.from_dict(
+        data={'time': times, 'value': values},
+    )
+
+    expr = pm.measure.data_loss(
+        'time', 'value', sampling_rate=sampling_rate, start_time=start, end_time=end, unit=unit,
+    )
+    result = df.select(expr)
+
+    expected_column_by_unit = {
```
dkrako: instead of putting this logic into the test function, it should be directly included in parametrize:

```python
@pytest.mark.parametrize(
    ('unit', 'expected_column_name'),
    [
        pytest.param('count', 'data_loss_count'),
        pytest.param('time', 'data_loss_time'),
        pytest.param('ratio', 'data_loss_ratio'),
    ],
)
```

This way it's more obvious what is tested.

tests/unit/measure/data_loss_test.py
```diff
@@ -0,0 +1,202 @@
...
+@pytest.mark.parametrize(
+    (
+        'times',
+        'values',
+        'sampling_rate',
+        'start',
+        'end',
+        'expected_count',
+        'expected_time',
+        'expected_ratio',
```
dkrako: I get that testing all three units in one function reduces a lot of parameter repetitions but I'm not sure if this is really optimal because now the function tests at least three different things. It's just minor and not that important to resolve, but maybe there's a better way to do this.

tests/unit/measure/data_loss_test.py
```diff
@@ -0,0 +1,202 @@
...
+    [
+        # 1 Hz with one internal time gap and one invalid value (NaN) -> missing=1
+        # (time) + 1 (invalid)
+        pytest.param(
+            [0.0, 1.0, 2.0, 4.0], [1.0, 1.0, nan, 1.0], 1.0, None, None,
+            2, 2.0, 2 / 5,
+            id='gap_plus_nan_invalid',
+        ),
+
+        # Explicit bounds expand range (2 expected missing) and one +inf value -> total 3
```
dkrako: I like the comment lines above the parametrizations.

However, I think this tests multiple things and would be best split:

- count inf as missing sample
- count missing start samples
- count missing end samples

tests/unit/measure/data_loss_test.py
```diff
@@ -0,0 +1,202 @@
...
+        'sampling_rate',
+        'start',
+        'end',
+        'expected_count',
+        'expected_time',
+        'expected_ratio',
+    ),
+    [
+        # 1 Hz with one internal time gap and one invalid value (NaN) -> missing=1
+        # (time) + 1 (invalid)
```
dkrako: the line break is a bit weird here

tests/unit/measure/data_loss_test.py
```diff
@@ -0,0 +1,202 @@
...
+        ),
+
+        # Explicit bounds expand range (2 expected missing) and one +inf value -> total 3
+        pytest.param(
+            [10.0, 11.0, 12.0], [1.0, inf, 1.0], 1.0, 9.0, 13.0,
+            3, 3.0, 3 / 5,
+            id='explicit_bounds_with_inf',
+        ),
+
+        # Irregular times, 1/3 Hz expectation over [0,12], plus one None value
```
dkrako: Testing multiple things is a bit confusing here.

Which timesteps are counted as missing? 3 (via 2.0?), 9,  and 12?

Replacing the `None` with `1.0` would result in `2`, right?

tests/unit/measure/data_loss_test.py
```diff
@@ -0,0 +1,202 @@
...
+
+        # Clip negative: observed > expected due to narrow end - invalid present but should still be
+        # clipped by expected? No: invalid rows are counted regardless - however our function
+        # sums both. To keep observed>expected scenario only, use no invalids -> 0.
+        pytest.param(
+            [0.0, 1.0, 2.0], [1.0, 1.0, 1.0], 1.0, 0.0, 1.5, 0, 0.0, 0.0,
+            id='clip_negative_time_only',
+        ),
+
+        # Unsorted input with an invalid (-inf) value
```
dkrako: This one is weird. How is the number of expected samples calculated?

dkrako:
This PR conflicts with #1419, however this won't be much of a problem to resolve as this PR here is mostly additions without changes to existing code (apart from a single line).

src/pymovements/measure/samples/measures.py
cbueth: When `end_time < start_time` are provided as literals, now `ValueError` is raised. 

https://github.com/pymovements/pymovements/blob/526f85fa323c215b76efcb5661ecc53a7956ccbb/src/pymovements/measure/samples/measures.py#L511-L515

`pl.when(valid_range).then(...).otherwise(pl.lit(None))` can be used to handle cases where bounds are derived from data, preventing silent 0 results for invalid ranges.

https://github.com/pymovements/pymovements/blob/526f85fa323c215b76efcb5661ecc53a7956ccbb/src/pymovements/measure/samples/measures.py#L530-L532

src/pymovements/measure/samples/measures.py
cbueth: I have removed this now.

src/pymovements/measure/samples/measures.py
cbueth: Good catch, it's changed:

https://github.com/pymovements/pymovements/blob/526f85fa323c215b76efcb5661ecc53a7956ccbb/src/pymovements/measure/samples/measures.py#L554

src/pymovements/measure/samples/measures.py
cbueth: I did this anyways in the same PR, hope that is fine:

https://github.com/pymovements/pymovements/blob/526f85fa323c215b76efcb5661ecc53a7956ccbb/src/pymovements/measure/samples/measures.py#L74

and

https://github.com/pymovements/pymovements/blob/526f85fa323c215b76efcb5661ecc53a7956ccbb/src/pymovements/measure/samples/measures.py#L32

with full coverage, see this file

https://github.com/pymovements/pymovements/blob/526f85fa323c215b76efcb5661ecc53a7956ccbb/tests/unit/measure/samples/measures/is_invalid_test.py#L50

`null_ratio` now uses this, too. For future, PRs, this can be used other measures to have some consistency for `null`, `NaN`, and `inf` across measures, if fitting.

I think the tests are properly parametrised.

tests/unit/measure/data_loss_test.py
cbueth: Thanks for the suggestion:

https://github.com/pymovements/pymovements/blob/526f85fa323c215b76efcb5661ecc53a7956ccbb/tests/unit/measure/samples/measures/data_loss_test.py#L33-L42

I kept most tests inside the class, except some.

tests/unit/measure/data_loss_test.py
cbueth: As per your suggestion, most tests are now inside the `TestDataLoss` class. I hope that is a good solution, even if the `expected_vals` now are dicts.

tests/unit/measure/data_loss_test.py
cbueth: Splitting these up into e.g. `test_internal_gaps_and_invalids`, `test_bounds_missing`, `test_unsorted_input` should to the trick. Let me know if 'count missing start samples' and 'count missing end samples' are covered the way you have the desired behavior in your mind.

tests/unit/measure/data_loss_test.py
cbueth: That should be the case:

https://github.com/pymovements/pymovements/blob/526f85fa323c215b76efcb5661ecc53a7956ccbb/src/pymovements/measure/samples/measures.py#L39-L43

I made the test comments more explicit.

tests/unit/measure/data_loss_test.py
cbueth: The expected sample count is determined by the `min` and `max` timestamps of the input, regardless of their order. If I remember right I had some code that assumed that the timestamp is already sorted, making min/max cheaper. This might be irrelevant now, but there is still a test for clipping

https://github.com/pymovements/pymovements/blob/526f85fa323c215b76efcb5661ecc53a7956ccbb/tests/unit/measure/samples/measures/data_loss_test.py#L185-L189

src/pymovements/measure/samples/measures.py
dkrako: great, thanks for looking into this more in depth!

src/pymovements/measure/samples/measures.py
dkrako: awesome, thanks a lot!

tests/unit/measure/data_loss_test.py
dkrako: interesting idea, never used the test classes before. thanks for digging into this!

dkrako (approved):
Excellent! Thank you so much for taking care of all the tests, I think they are very clearly written now!

dkrako:
> Follow-up: Add a plotting function that visualises the data loss. Idea: Histogram by connected data loss length. And/or scarf plot.

Great idea, thanks for the input! Can you create an issue for that?

cbueth:
> > Follow-up: Add a plotting function that visualises the data loss. Idea: Histogram by connected data loss length. And/or scarf plot.
> 
> Great idea, thanks for the input! Can you create an issue for that?

This was collected during a meeting, I am not sure, I have the right idea of the expected visualisation. Still, I can sketch something, but these days I do not have the capacity, sorry.

dkrako:
Yes, an issue labeled as plotting will at least make it more visible when searching for visualization tasks.

PR #1422, opened 2025-12-17
fix: copy load_kwargs before manipulating them during load

dkrako:
## Description

Loading precomputed events or reading measures failed.

The load kwargs shared the same dictionary object despite manipulating it (handling legacy kwargs from `DatsetDefinition.custom_read_kwargs` and using `load_kwargs.pop()` for some file types.

## Implemented changes

- [x] applied deepcopy to load kwargs in loading functions of each content type

## How Has This Been Tested?

- [x] load multiple precomputed event rda files
- [x] load multiple precomputed reading measure rda files

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -9,39 +9,6 @@ jobs:
   build:
     uses: ./.github/workflows/build.yml
 
-  publish-to-github:
-    name: Attach distribution packages to GitHub release
-
-    needs: build
-    runs-on: ubuntu-latest
-
-    steps:
-    - name: Download distribution packages
-      uses: actions/download-artifact@v4
-      with:
-        name: pymovements-distribution-packages-${{ github.sha }}
-        path: dist/
-
-    - name: Upload wheel
-      uses: actions/upload-release-asset@v1.0.1
-      env:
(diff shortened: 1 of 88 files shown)


tests/integration/public_dataset_processing_test.py
```diff
@@ -42,14 +42,17 @@ def test_public_dataset_processing(dataset_name, tmp_path):
     # Do some basic transformations.
     if dataset.definition.resources.has_content('gaze'):
         if 'pixel' in dataset.gaze[0].columns:
-            dataset.pix2deg()
+            if dataset.gaze[0].experiment.screen.height_cm is not None:
+                dataset.pix2deg()
```
SiQube, line 46: shouldn't this prevent the etdd70 failure?

SiQube: or what is this line for?

SiQube (approved):
feel free to merge

tests/integration/public_dataset_processing_test.py
dkrako: the ETDD70 error is raised in the line after when calling pos2vel(). the whole integration test should be refactored but that's out of scope of this PR.

PR #1425, opened 2025-12-19
dataset: improve potsdam binge dataset definitions

prassepaul:
## Description

Correct copy/paste bugs and improve dataset definitions of the 2 Potsdam Binge datasets.

## How Has This Been Tested?

loading, transformations tested (pix2deg(), pos2vel(), detect('ivt'))

## Type of change
Remove irrelevant items:
- Bug fix
- New dataset
- Documentation update


src/pymovements/datasets/potsdam_binge_pvt.yaml
```diff
@@ -0,0 +1,286 @@
+name: "PotsdamBingePVT"
+
+long_name: 'Potsdam Binge PVT dataset'
+
+resources:
+  - content: gaze
+    url: "https://osf.io/download/9vbs8/"
+    filename: "a.zip"
```
dkrako: does `a.zip` stand for something specific? A better name could be something like `eyelink.zip`.

src/pymovements/datasets/potsdam_binge_pvt.yaml
```diff
@@ -0,0 +1,286 @@
...
+          pupil_size_pupilcore_interpolated: !polars.Float32
+          pupil_confidence_interpolated: !polars.Float32
+          time_to_prev_bac: !polars.Float32
+          time_to_next_bac: !polars.Float32
+          prev_bac: !polars.Float32
+          next_bac: !polars.Float32
+        separator: ","
+
+  - content: gaze
+    url: "https://osf.io/download/9vbs8/"
```
dkrako: this resource is duplicated

src/pymovements/datasets/potsdam_binge_pvt.yaml
```diff
@@ -0,0 +1,286 @@
...
+            width_px: 1920
+            height_px: 1080
+            width_cm: 59.76
+            height_cm: 33.615
+            origin: "center"
+          eyetracker:
+            sampling_rate: 1000
+            left: true
+            right: false
+            model: "Pupil Core eye-tracking glasses"
```
dkrako: the model is inconsistent with the previous `a.zip` entry. Does this resource include the eyelink or pupilcore files?

src/pymovements/datasets/potsdam_binge_pvt.yaml
```diff
@@ -0,0 +1,286 @@
...
+          pupil_confidence_interpolated: !polars.Float32
+          time_to_prev_bac: !polars.Float32
+          time_to_next_bac: !polars.Float32
+          prev_bac: !polars.Float32
+          next_bac: !polars.Float32
+        separator: ","
+
+  - content: gaze
+    url: "https://osf.io/download/yqukn/"
+    filename: "b.zip"
```
dkrako: I'm confused which file includes what data. If this includes the pupil core glasses you can call this `pupilcore.zip`

src/pymovements/datasets/potsdam_binge_pvt.yaml
```diff
@@ -0,0 +1,286 @@
...
+          pupil_confidence_interpolated: !polars.Float32
+          time_to_prev_bac: !polars.Float32
+          time_to_next_bac: !polars.Float32
+          prev_bac: !polars.Float32
+          next_bac: !polars.Float32
+        separator: ","
+
+  - content: gaze
+    url: "https://osf.io/download/yf2xa/"
+    filename: "e.zip"
```
dkrako: what data is included here?

src/pymovements/dataset/dataset_files.py
```diff
@@ -385,11 +386,21 @@ def load_gaze_file(
...
                 }
 
-            gaze = from_csv(
-                filepath,
-                experiment=dataset_definition.experiment,
-                **load_function_kwargs,
-            )
+            if 'experiment' in load_function_kwargs:
+                if type(load_function_kwargs['experiment']) is dict:
+                    dataset_definition.experiment = Experiment.from_dict(
```
dkrako: This is potentially dangerous, as the experiment can differ between resource groups. Overwriting `DatasetDefinition.experiment` could then lead to wrong experiment specs.

dkrako (changes requested):
Great, thanks a lot for your work!

It seems that all resources are duplicated in your merged definition. I'm also bit confused about what is exactly included in which resource archive. Changing the filenames to something more descriptive would help there.

Moreover, I think you can also remove the irrelevant resources from the old definition files (i.e. remove eyelink resources from PotsdamBingeWearablePVT and vice versa).

src/pymovements/datasets/potsdam_binge_pvt.yaml
prassepaul: each zip file contains the complete data (pupilcore + eyelink) for one condition. a= alcohol, b=baseline, e=enrollment, s=sleep.

src/pymovements/datasets/potsdam_binge_pvt.yaml
prassepaul: its duplicated, because here we define the pixel coordinates to be the ones collected with eyelink. in the second (duplicated block) the pixel coordinates are the ones using pupilcore.

src/pymovements/datasets/potsdam_binge_pvt.yaml
prassepaul: see above. a.zip contains the complete data for all alcohol sessions with pupilcore and eyelink data. one eye is tracked using pupilcore the other with eyelink.

src/pymovements/datasets/potsdam_binge_pvt.yaml
prassepaul: see above

src/pymovements/datasets/potsdam_binge_pvt.yaml
prassepaul: see above

src/pymovements/dataset/dataset_files.py
prassepaul: I am not sure how to deal with this. We have data collected using two different devices (-> different experiments) in one *.csv file (extracted from the *.zip files). We could use different columns for the different eyes, but then the experiment needs to have to tracking devices.

src/pymovements/datasets/potsdam_binge_pvt.yaml
dkrako: Oh I see, I totally misunderstood the data.

src/pymovements/datasets/potsdam_binge_pvt.yaml
dkrako: Now I understand the issue, and the limiting factor really is in the way we handle the pixel columns in `Gaze`.

There are two ways around this:
- more comprehensive pixel columns with flexible components (could be enabled by #453)
- support for multiple counts of pixel columns, e.g. pixel_eyelink, pixel_pupilcore. this might be a mess.

All in all, this is a much more complex issue than I thought, so I think we can just close the issue until we sort out the main problems in `Gaze`.

src/pymovements/datasets/potsdam_binge_pvt.yaml
dkrako: Thanks for the clarification. As the combining of the two definitions isn't really possible currently, could you change the filenames in the original two dataset definitions to `alcohol.zip`, `baseline.zip`, `enrollment.zip`, `sleep.zip`?

src/pymovements/dataset/dataset_files.py
dkrako: I think the only critical part in `EyeTracker` is the sampling rate, and as you already fused the eye trackers into a single file (i.e. made sure the sampling rate is equal for both signals) that won't be a problem.

The other fields are purely desriptive, so for the eye tracker model we could just use `EyeLink 1000 Plus, Pupil Core eye-tracking glasses`.

Furtantely, as the main issue is with the inflexibilities in the `Gaze` class we don't need to worry about that currently. However, we should discuss at one time how we want to handle multiple eye trackers in an experiment, especially if we want to support fusing signals from two eye trackers.

dkrako (changes requested):
Alright thank you for all the effort!

I think there are two major issues that block the merging of the two dataset definitions:
- the `Gaze` class and especially the initializer lacks support for multiple variants of the same type (e.g., multiple x/y pixel coordinates for the same eye in `Gaze.samples`, coming from different eye trackers)
- `Gaze.experiment` does not support multiple eye trackers

These two issues are interconnected and will take some time to figure out. I think #453 could be a way to move forward and allow for something like:

```yaml
pixel_columns:
  x_eyelink: x_pix_eyelink
  y_eyelink: y_pix_eyelink
  y_pupilcore: y_pix_pupilcore
  x_pupilcore: x_pix_pupilcore
```

However, this might not fully resolve the issue, as it doesn't map an eye tracker to a `Gaze.samples` column/component. But as the eye trackers are already fused in the dataset, the sampling rate is equal for both your eye trackers it wouldn't be a big problem to just use a comma separated list in the strings of the descriptive fields (e.g., `EyeTracker.model`).

However, the PR still has some valuable changes that improve the existing definitions.

Please take care of these points and we'll merge this:
- [ ] remove the new merged dataset from source and docs
- [ ] update the resource filenames to more descriptive names

PR #1428, opened 2025-12-29
ci: add slack release notification job

dkrako:
## Description

Notify about new release on slack channel (for now just on the AEye workspace). 

pre-releases should be excluded from notification.

## Implemented changes

- [ ] added bot to AEye slack workspace
- [ ] add notify slack job to publish workflow

## How Has This Been Tested?

I'm not sure if 

```
if: ${{ !github.release.prerelease }}
```

is correct, but I also can't really try it until the next pre-release.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -59,3 +59,16 @@ jobs:
 
     - name: Publish distribution to PyPI
       uses: pypa/gh-action-pypi-publish@release/v1
+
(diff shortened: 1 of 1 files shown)


SiQube (approved):

dkrako:
> I guess multipleye + discord would be most important next two servers

Thanks for the review! You're right, I just thought I'd start with the less used one in case I misconfigured something.

PR #1447, opened 2026-01-23
feat: add `segmentation2events` and `events2segmentation` conversion

### LINKED ISSUE #24: Implement segmentation2events and events2segmentation methods
## Description of the problem
Depending on what you want to do with your data you will either use a list of events (which have at least and onset and offset, but can potentially also have further information like peak velocity), or you will use a binary array to encode at which time steps this event type took place (segmentation map).

We therefore need functions to convert between those two.

## Description of a solution
- [ ] implement segmentation2events
- [ ] implement events2segmentation
- [ ] create synthetic test data where expected results are comprehensible
- [ ] test both functions

- [ ] both functions with passing tests

cbueth:
Adds `events2segmentation` and `segmentation2events`: conversion between binary segmentation maps and event representations using Polars expressions.

## Description

Bidirectional conversion between event lists (onset/offset timing information) and binary segmentation maps (encoding event occurrence per time step).

```python
>>> import polars as pl
>>> from pymovements.events import events2segmentation
>>>
>>> events_df = pl.DataFrame(
...     {'name': ['blink', 'blink', 'not_blink'], 'onset': [2, 7, 3], 'offset': [5, 9, 6]}
... )
>>> gaze_df = pl.DataFrame({'time': range(10)})
>>>
>>> # Create a boolean indicator column for blinks using a Polars expression
>>> gaze_df.with_columns(
...     events2segmentation(events_df, name='blink')
... )
shape: (10, 2)
┌──────┬───────┐
│ time ┆ blink │
│ ---  ┆ ---   │
│ i64  ┆ bool  │
╞══════╪═══════╡
│ 0    ┆ false │
│ 1    ┆ false │
│ 2    ┆ true  │
│ 3    ┆ true  │
│ 4    ┆ true  │
│ 5    ┆ true  │
│ 6    ┆ false │
│ 7    ┆ true  │
│ 8    ┆ true  │
│ 9    ┆ true  │
└──────┴───────┘
>>> # With trial columns
>>> events_df = pl.DataFrame({
...     'name': ['blink', 'blink'],
...     'onset': [2, 1],
...     'offset': [3, 3],
...     'trial': [1, 2],
... })
>>> gaze_df = pl.DataFrame({
...     'time': pl.Series([0, 1, 2, 0, 1, 2, 3], dtype=pl.Int64),
...     'trial': [1, 1, 1, 2, 2, 2, 2],
... })
>>> gaze_df.with_columns(
...     events2segmentation(events_df, name='blink', trial_columns=['trial'])
... )
shape: (7, 3)
┌──────┬───────┬───────┐
│ time ┆ trial ┆ blink │
│ ---  ┆ ---   ┆ ---   │
│ i64  ┆ i64   ┆ bool  │
╞══════╪═══════╪═══════╡
│ 0    ┆ 1     ┆ false │
│ 1    ┆ 1     ┆ false │
│ 2    ┆ 1     ┆ true  │
│ 0    ┆ 2     ┆ false │
│ 1    ┆ 2     ┆ true  │
│ 2    ┆ 2     ┆ true  │
│ 3    ┆ 2     ┆ true  │
└──────┴───────┴───────┘
```

```python
>>> import polars as pl
>>> from pymovements.events import segmentation2events
>>> segmentation = pl.Series([0, 0, 1, 1, 1, 0, 0, 1, 1, 0])
>>> segmentation2events(segmentation, name='blink')
shape: (2, 3)
┌───────┬───────┬────────┐
│ name  ┆ onset ┆ offset │
│ ---   ┆ ---   ┆ ---    │
│ str   ┆ i64   ┆ i64    │
╞═══════╪═══════╪════════╡
│ blink ┆ 2     ┆ 4      │
│ blink ┆ 7     ┆ 8      │
└───────┴───────┴────────┘
>>> # Using a time column:
>>> time = pl.Series([0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0])
>>> segmentation2events(segmentation, name='blink', time_column=time)
shape: (2, 3)
┌───────┬───────┬────────┐
│ name  ┆ onset ┆ offset │
│ ---   ┆ ---   ┆ ---    │
│ str   ┆ f64   ┆ f64    │
╞═══════╪═══════╪════════╡
│ blink ┆ 0.3   ┆ 0.5    │
│ blink ┆ 0.8   ┆ 0.9    │
└───────┴───────┴────────┘
```

## Discussion

- ~~As of now `num_samples` is needed for `events2segmentation`. Alternatively this can be determined by the maximum in the offset colum.~~ Not needed anymore. Use `time_column` to align events, avoiding non-zero start times, recording gaps, and irregular sampling.
- ~~Does the assumption hold that Events **always start with an onset**? This PR takes this for granted.~~ Yes
- ~~As this is a first intent of implementation, please suggest improvements.~~

## Implemented changes

- [x] `events2segmentation()`: Convert `Events`/`DataFrame` to binary segmentation array
  - Returns a `pl.Expr`, for use with `with_columns`
- [x] `segmentation2events()`: Convert binary segmentation array to `Events` object
  - Takes an optional `time_column` to return onsets and offsets in temporal units, instead of indices.
- [x] Added reference section on _Segmentation Methods_ https://pymovements--1447.org.readthedocs.build/en/1447/reference/events.html
- [x] Support trialised data: `events2segmentation` can filter by `trial_columns`
- [x] Overlap detection that issues a `UserWarning`, being trial-aware
- [x] Docstrings with examples

## How Has This Been Tested?

- [x] Parametrized tests
  - covered edge cases: empty events, single samples, integer datatypes
- [x] Test errors
- [x] Doctests

Code coverage locally is at 100%.

## Type of change

Remove irrelevant items:
- New functionality

## Context

Resolves #24

#### related issues:
- https://github.com/pymovements/pymovements/pull/1013#issuecomment-3553057073

#### required by:
- #

#### requires:
- [ ] #

#### conflicts with:
- #

## Checklist:

### DIFF
--- a/src/pymovements/gaze/_utils/_parsing_eyelink.py
+++ b/src/pymovements/gaze/_utils/_parsing_eyelink.py
@@ -546,7 +546,7 @@ def parse_eyelink(
 
             for additional_column in additional_columns:
                 events[additional_column].append(
-                    current_event_additional[event_name][eye][additional_column],
+                    current_event_additional[event_name][eye].get(additional_column),
                 )
             current_event_additional[event_name][eye] = {}
 
--- a/tests/unit/gaze/io/asc_test.py
+++ b/tests/unit/gaze/io/asc_test.py
@@ -901,3 +901,48 @@ def test_from_asc_keeps_remaining_metadata_private_and_pops_cal_val(make_example
     # utilities.
     assert 'data_loss_ratio' in gaze._metadata
     assert 'data_loss_ratio_blinks' in gaze._metadata
+
+
+@pytest.mark.filterwarnings('ignore:.*No metadata.*:UserWarning')
+@pytest.mark.filterwarnings('ignore:.*No mount configuration.*:UserWarning')
+@pytest.mark.filterwarnings('ignore:.*No recording configuration.*:UserWarning')
+@pytest.mark.filterwarnings('ignore:.*No samples configuration.*:UserWarning')
+@pytest.mark.filterwarnings('ignore:.*No screen resolution.*:UserWarning')
+def test_from_asc_orphaned_event_end_marker_with_custom_patterns_does_not_raise_keyerror(
+        make_text_file,
+):
+    """Orphaned event end markers with custom patterns should not raise KeyError, but should warn.
+
+    This test reproduces a scenario where an event end marker appears before the
+    associated context dictionary has been populated with keys from custom patterns.
+    """
+    body = (
+        'EFIX R 1000 1100 100 500.0 500.0 1000\n'
+        'MSG 1200 START_TRIAL_1\n'
+    )
+    patterns = [r'START_TRIAL_(?P<trial_id>\d+)']
+    filepath = make_text_file(filename='orphaned_event.asc', body=body)
+
+    with pytest.warns(UserWarning, match='Missing start marker before end for event'):
+        gaze = from_asc(filepath, patterns=patterns, events=True)
+
+    expected_events = pl.from_dict(
+        data={
+            'name': ['fixation_eyelink'],
+            'eye': ['right'],
+            'onset': [1000],
+            'offset': [1100],
+            'duration': [100],
+            'trial_id': [None],
+        },
+        schema={
+            'name': pl.Utf8,
+            'eye': pl.Utf8,
+            'onset': pl.Int64,
+            'offset': pl.Int64,
+            'duration': pl.Int64,
+            'trial_id': pl.Null,
+        },
+    )
+
+    assert_frame_equal(gaze.events.frame, expected_events, check_column_order=False)


src/pymovements/events/segmentation.py
```diff
@@ -0,0 +1,219 @@
...
+
+    >>> from pymovements.events.events import Events
+    >>> events2segmentation(Events(events_df), 10)
+    array([0, 0, 1, 1, 1, 0, 0, 1, 1, 0], dtype=int32)
+    """
+    if num_samples < 0:
+        raise ValueError(f"num_samples must be non-negative, but is {num_samples}")
+
+    onset_column = onset_column or 'onset'
+    offset_column = offset_column or 'offset'
```
dkrako: as `str` is immutable, this is better solved via default parameters:

```
    onset_column: str = 'onset',
    offset_column: str = 'offset',
```

src/pymovements/events/segmentation.py
```diff
@@ -0,0 +1,219 @@
...
+
+import numpy as np
+import polars as pl
+
+from ..events.events import Events
+
+
+def events2segmentation(
+    events: Events | pl.DataFrame,
+    num_samples: int,
```
dkrako: this should also include a `name` argument, such that only events with that particular name are used for creating the segmentation

src/pymovements/events/segmentation.py
```diff
@@ -0,0 +1,219 @@
...
+    onset_column: str | None = None,
+    offset_column: str | None = None,
+) -> np.ndarray:
```
dkrako: I would prefer if the return type would be `polars.Expr`, this way it will be better compatible with our remaining code.

I have a usage like this in mind:

```python
gaze.samples = gaze.samples.with_columns(
    events2segmentation(
        events=gaze.events,
        name='blink',
    )
)
```

within the `events2segmentation` function you could construct the expression like this:

```python
is_event.then(True).otherwise(False).alias(name)
```

`is_event` is also an expression that can be constructed in a for-loop:

```python
is_event = pl.lit(False)  # anchor
for event in events.filter(pl.col('name') == name).to_dicts():
    is_event = is_event | (pl.col(time_column).ge(event[onset_column]) & pl.col(time_column).le(event[offset_column]))
```

However, this is simplified, as it ignores trial columns, so something like this would also be needed within that for-loop:

```python
is_same_trial = pl.lit(True)  # anchor
for trial_column in trial_columns:
    is_same_trial = is_same_trial & (pl.col(trial_column) == event['trial_column'])

is_event = is_event & is_trial
```

It might be possible to implement this without the anchors, it's just what I came up with as a starting point, but I hope that clarifies a bit how that method could be implemented in polars.

This implementation would need a `time_column: str = time` argument instead of `num_samples`.

src/pymovements/events/segmentation.py
```diff
@@ -0,0 +1,219 @@
...
+    if offset_column not in events_df.columns:
+        raise ValueError(f"Offset column '{offset_column}' not found in events.")
+
+    # Extract event data as numpy arrays
+    onsets = events_df[onset_column].to_numpy()
+    offsets = events_df[offset_column].to_numpy()
+
+    # Validation of constraints
+    if np.any(onsets < 0):
+        raise ValueError('Onset must be non-negative, but found negative values')
```
dkrako: I think it's fine to allow for negative values here as it's not critical. Although we currently do not provide any functionality for this. there might be usecases, where a user wants to set t=0 for a stimulus presentation onset or a button response. There it might happen that negative value timestamps exist.

src/pymovements/events/segmentation.py
```diff
@@ -0,0 +1,219 @@
...
+    if np.any(offsets < 0):
+        raise ValueError('Offset must be non-negative, but found negative values')
+    if np.any(onsets >= offsets):
+        raise ValueError(
+            'Onset must be less than offset, but found invalid event(s)',
```
dkrako: this check however is critical.

src/pymovements/events/segmentation.py
```diff
@@ -0,0 +1,219 @@
...
+        for onset, offset in zip(onsets, offsets)
+    ]
+    all_indices = np.concatenate(indices_list)
+
+    # Check for overlaps: total index count vs unique count
+    if len(indices_list) > 1:
+        total_indices = sum(len(indices) for indices in indices_list)
+        unique_indices = len(np.unique(all_indices))
+        if total_indices > unique_indices:
+            raise ValueError('Overlapping events detected')
```
dkrako: instead of an exception a warning should be sufficient here.

tests/unit/events/segmentation_test.py
```diff
@@ -0,0 +1,360 @@
...
+            10,
+            np.array([0, 0, 1, 1, 1, 0, 0, 0, 0, 0], dtype=np.int32),
+            None,
+            None,
+            id='int64_dtype',
+        ),
+    ],
+)
+def test_events2segmentation(events, num_samples, expected, onset_col, offset_col):
+    if onset_col is None and offset_col is None:
```
dkrako: I think it's best to split this test into two test functions

src/pymovements/events/segmentation.py
```diff
@@ -0,0 +1,219 @@
...
+from __future__ import annotations
+
+import numpy as np
+import polars as pl
+
+from ..events.events import Events
+
+
+def events2segmentation(
+    events: Events | pl.DataFrame,
```
dkrako: as this is a low-level function I think it's best to use `events: pl.DataFrame` and drop the dependency on `Events`

src/pymovements/events/segmentation.py
```diff
@@ -0,0 +1,219 @@
...
+        weights=np.ones_like(all_indices, dtype=np.int32),
+    )
+    # Convert counts to binary (handle overlaps by clipping to 1)
+    segmentation = np.clip(segmentation, 0, 1)
+
+    return segmentation.astype(dtype=np.int32)
+
+
+def segmentation2events(
+    segmentation: np.ndarray,
```
dkrako: a `name: str` argument would be useful here, filling the `name` column of the output events

src/pymovements/events/segmentation.py
```diff
@@ -0,0 +1,219 @@
...
+) -> Events:
```
dkrako: it's best to just return a `pl.DataFrame` here

src/pymovements/events/segmentation.py
cbueth: In the case of overlapping events the output would stay strictly 1/0, not summing up to 2. We could also just use bools as dtype, no?

tests/unit/events/segmentation_test.py
cbueth: This answers the question from https://github.com/pymovements/pymovements/pull/1447#issuecomment-3798893651, will do so :)

src/pymovements/events/segmentation.py
cbueth: Right, no reason to do otherwise. Thanks!

src/pymovements/events/segmentation.py
cbueth: A `name` argument has been added.

src/pymovements/events/segmentation.py
cbueth: Great input, will be useful also for the follow-up. I have refactored the whole to return a `polars.Expr`.

src/pymovements/events/segmentation.py
cbueth: Has been added 👍

src/pymovements/events/segmentation.py
cbueth: `events2segmentation` now accepts `pl.DataFrame`, removing the `Events` class

src/pymovements/events/segmentation.py
cbueth: I changed it that `segmentation2events` accepts `np.ndarray` with both `bool` and `int`, and returns a `pl.DataFrame` instead of an `Events` object.

src/pymovements/events/segmentation.py
```diff
@@ -0,0 +1,382 @@
...
+        The values to use for the onset and offset columns. If provided, the indices
+        of the events will be mapped to the values in this column. If None, the
+        indices themselves will be used. Default is None.
+
+    Returns
+    -------
+    pl.DataFrame
+        A DataFrame containing the onset and offset of each event.
+        The onset is inclusive and the offset is exclusive.
+        Onsets and offsets correspond to the values in ``time_column`` if provided,
```
dkrako, line 256: This is currently documented so bad in pymovements that even I got confused, but I'm pretty sure the offset is actually inclusive. Exclusive offset would be problematic in case of irregular sampling rates.

src/pymovements/events/segmentation.py
```diff
@@ -0,0 +1,382 @@
...
+                is_same_trial = is_same_trial & (pl.col(trial_column) == event[trial_column])
+            is_event = is_event | (is_in_time_range & is_same_trial)
+        else:
+            is_event = is_event | is_in_time_range
+
+    return is_event.alias(name)
+
+
+def segmentation2events(
+    segmentation: np.ndarray,
```
dkrako: it would be great if this would support `pl.Series`, the `np.ndarray` support isn't really needed here as it's easy to adjust on the calling side of the function.

One of the things that the `segmentation2events()` function would allow is writing event detection functions in polars.

For example, the [I-VT](https://github.com/pymovements/pymovements/blob/main/src/pymovements/events/detection/_ivt.py) would be trivial to implement in polars that way.

src/pymovements/events/segmentation.py
```diff
@@ -0,0 +1,382 @@
...
+        time_column = time_column.to_numpy()
+
+    onsets_mapped = time_column[onsets]
+
+    offsets_mapped = np.zeros_like(offsets, dtype=time_column.dtype)
+    in_bounds = offsets < len(time_column)
+    offsets_mapped[in_bounds] = time_column[offsets[in_bounds]]
+
+    out_of_bounds = ~in_bounds
+    if np.any(out_of_bounds):
```
dkrako: if I understand correctly then functionally this numpy implementation partly overlaps with our [`transforms_numpy.consecutive()`](https://github.com/pymovements/pymovements/blob/9fd835c929606208f218af02931271b4ab143c0f/src/pymovements/gaze/transforms_numpy.py#L570-L589).

I would prefer that this could be implemented in polars instead, as in the long run I want to get rid of the `transforms_numpy` module (we just need to replace `split()` and `consecutive()`, `consecutive()` is currently used in our event detection functions).

I guess the run-length encoding functions from polars should to the trick: [`polars.Expr.rle_id`](https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.rle_id.html).

src/pymovements/events/segmentation.py
```diff
@@ -0,0 +1,382 @@
...
+        else:
+            is_event = is_event | is_in_time_range
+
+    return is_event.alias(name)
+
+
+def segmentation2events(
+    segmentation: np.ndarray,
+    name: str,
+    time_column: np.ndarray | pl.Series | None = None,
```
dkrako: this would also need some support for trialized data.

dkrako (changes requested):
Thank you so much for quickly taking care! Great that you already provided small examples in the docstrings.

same as with `events2segmentation` I would be very much in favor to implement `segmentation2events` in polars instead of using numpy.

src/pymovements/events/segmentation.py
cbueth: I will make the offset inclusive. Another motivation to do so is that for EyeLink ascii files the reported time in ms is even the difference +2ms, in other words, the length one would get treating the both timestamps as inclusive and still adding one millisecond.

dkrako: yes, so the difference between pymovements and eyelink isn't really the offset definition, but the definition of the duration function. pymovements underestimates event duration by `sampling_interval/2`, eyelink overestimates by the same.

There's no actual right way to do it, but there's definitely some room for improvement on the pymovements side, either documentation or configuration related (e.g. an additional argument in `duration()`).

here's a final discussion comment from when the eyelink event parsing was implemented: https://github.com/pymovements/pymovements/pull/945#issuecomment-2867341724

cbueth: Thanks for the context!

cbueth:
Pulling through with the trialised support for `segmentation2events` is the logical next step, it has been implemented now. I was not sure first if it was very neeccessary, but like this the both conversion functions are more symmetric. Also pointing out `polars.Expr.rle_id` really helped where to look, I did not know it. I updated and added some test cases.

src/pymovements/events/segmentation.py
cbueth: Now the `trial_columns` parameter accepts a `pl.DataFrame` containing trial identifiers.

src/pymovements/events/segmentation.py
cbueth: As suggested the function only uses polars and converts numpy inputs to `pl.Series`. `pl.Expr.rle_id()` is used to create unique IDs for segments. I extract for `True` and aggregate with `.group_by().agg()`, using `.min()` for onsets and `.max()` for offsets 🤓

cbueth: https://github.com/pymovements/pymovements/blob/4389db01f1428b1d2930b15537e0b4c27d8e4d8e/src/pymovements/events/segmentation.py#L359-L370

src/pymovements/events/segmentation.py
cbueth: Both functions have been changed to be inclusive. `events2segmentation` now uses `.le(event[offset_column])` instead of `.lt()`. Inversely, for `segmentation2events` the offset calculation now uses `pl.col("__time__").max()`, which returns the last sample's timestamp/index of the event (inclusive), rather than the following sample. `_has_overlap` has also been updated accordingly. Much much simpler than the first PR's version.

src/pymovements/events/segmentation.py
cbueth: The function now additionally accepts `pl.Series` for the `segmentation` and `time_column`.

src/pymovements/events/segmentation.py
dkrako: yes, that check is less critical than making sure that `onset <= offset` but nice to have. If it gets reused more often we might consider adding a method to `Events` that validates the object for integrity.

PR #1449, opened 2026-01-26
chore: split tests that contain if/else logic

### LINKED ISSUE #1397: get rid of if-else logic in tests
Some of our tests have if-else logic and should be either split into separate test functions or parametrized further (e.g., by adding the expected output to the test parametrization). If this results in lots of code repetition for test setup/teardown it might be worth to refactor some of these blocks into fixtures.

cbueth:
## Description

Some tests use if-else logic that should much rather be put into separate tests or use parametrization properly.
As example, if there are tests that use a parametrisation that totally changes the test, there rather should be two tests. And if there are tests that use if-else that would be included inbto the parametrization, mainly keeping the test code uniformly, this should be the prefered option.

## Implemented changes

- [x] Split up `test_previous_saccades_untouched_after_map_to_aois` in 9b1a9d7d62b2386251e20290affb444aae1bcc65
- [x] Split `test_step_function` to
     - test_step_function
     - test_step_function_exceptions
     - test_step_function_dimensions
- [x] Refactor overlapping AOI tests `test_get_aoi_overlap_warns`. Added `expect_warning` parameter to make warning expectations explicit.
- [x] Separate verbose cases for `test_download_and_extract_archive` into
  - `test_download_and_extract_archive_extract_dirpath_None_verbose_true` and
  - `test_download_and_extract_archive_extract_dirpath_None_verbose_false`
- [x] `test_heatmap_show` and related tests: add `position_column_mapping` fixture (Good or undo?)
- [x] Split `map_to_aois` tests for better coverage of `preserve_structure` scenarios
  - `test_events_map_to_aois_preserve_structure_true`
  - `test_events_map_to_aois_preserve_structure_false`

## How Has This Been Tested?

These are tests themselves.

## Type of change

- Test code

## Context

Resolves #1397

## Checklist:

### DIFF
--- a/src/pymovements/gaze/gaze.py
+++ b/src/pymovements/gaze/gaze.py
@@ -1033,6 +1033,13 @@ class Gaze:
 
             new_events = method(**method_kwargs)
 
+            if len(new_events) == 0:
+                warn(
+                    f"{getattr(method, '__name__', method)}: No events were detected.",
+                    UserWarning,
+                    stacklevel=2,
+                )
+
             self.events.frame = polars.concat(
                 [self.events.frame, new_events.frame],
                 how='diagonal_relaxed',
@@ -1085,6 +1092,13 @@ class Gaze:
                 new_events.add_trial_column(self.trial_columns, group_identifier)
 
                 new_events_grouped.append(new_events.frame)
+
+            if not new_events_grouped or any(len(df) == 0 for df in new_events_grouped):
+                warn(
+                    f"{getattr(method, '__name__', method)}: No events were detected.",
+                    UserWarning,
+                    stacklevel=2,
+                )
 
             self.events.frame = polars.concat(
                 [self.events.frame, *new_events_grouped],
@@ -1155,7 +1169,7 @@ class Gaze:
         if overwrite_columns:
             warn(
                 'The following columns already exist in event and will be overwritten: '
-                f"{overwrite_columns}",
+                f'{overwrite_columns}',
             )
             self.events.drop(overwrite_columns)
         if results.height:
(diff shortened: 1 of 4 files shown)


tests/unit/stimulus/text_aoi_test.py
```diff
@@ -30,24 +30,51 @@
...
+        aois=df,
+        aoi_column='label',
+        start_x_column='sx',
+        start_y_column='sy',
+        width_column='width',
+        height_column='height',
+    )
+
+    row = {'x': x, 'y': y}
+    if expect_warning:
```
dkrako: this should be split too. to avoid the code repetition you can also create a local fixture function to create the stimulus object, but it's not really necessary.

tests/unit/utils/downloads_test.py
```diff
@@ -110,9 +152,11 @@ def test_download_and_extract_archive_invalid_md5(tmp_path):
     with pytest.raises(RuntimeError) as excinfo:
         download_and_extract_archive(url, tmp_path, _download_filename, extract_dirpath, md5)
 
-    msg, = excinfo.value.args
-    assert msg == f"File {os.path.join(tmp_path, 'pymovements-0.4.0.tar.gz')} "\
+    (msg,) = excinfo.value.args
```
dkrako: you can also just write `msg = excinfo.value.args[0]`

this was written when we didn't really know that `pytest.raises` has a `match` argument, so you can decide yourself which way you prefer.

cbueth: Changed in a89db205.

tests/unit/stimulus/text_aoi_test.py
cbueth: Split in 31a784bc.

cbueth:
Thanks for reviewing the changes. I also think it is fine to just merge and coming back to improving others further down the line.

dkrako (approved):

Issue #1082, opened 2025-04-08
improve `null_ratio` functionality and deprecate `parsing._calculate_data_loss`


dkrako:
## Description of the problem

The current implementation of the [`null_ratio`](https://github.com/aeye-lab/pymovements/blob/4ae7f2167c0357cf75385706556577765c27f2f8/src/pymovements/measure/measures.py#L29) measure is very trivial as it only measures the ratio of null values in a series:

https://github.com/aeye-lab/pymovements/blob/4ae7f2167c0357cf75385706556577765c27f2f8/src/pymovements/measure/measures.py#L29-L60

Unrelated to that measure, we have a `_calculate_data_loss()` function in `parsing.py` that calculates data loss on eyelink files and writes it to a metadata dictionary:

https://github.com/aeye-lab/pymovements/blob/4ae7f2167c0357cf75385706556577765c27f2f8/src/pymovements/gaze/_utils/parsing.py#L439-L484

The function included in the parsing module is much more sophisticated:
 - it calculates dataloss by inferring the number of expected samples by using the sampling rate
 - it checks for blink ratio by using eyelink blink event data

Unfortunately the latter function can not be used for a `GazeDataFrame` in general, only via `gaze.from_asc()`.

## Description of a solution

Instead, it would be much better if the two measures could be provided via `GazeDataFrame.measure()`.

The calculation of measures is already grouped by `GazeDataFrame.trial_columns`, so we automatically get the measures for each trial.

The blink ratio will need to extend the current measure functionalities, as these are implemented for samples only.
The blink ratio would probably need both samples and events.

## Minimum acceptance criteria
- [ ] dataloss measure
- [ ] blink ratio measure

PR #1452, opened 2026-01-27
feat:  add generic `measure_events_ratio` method

cbueth:
## Description

Using `events2segmentation` from #1447, this implements `measure_events_ratio()` to `Gaze`  to calculate event coverage ratios.
This should make it possible to deprecate `parsing._calculate_data_loss` with using `measure_events_ratio('blink')`, as per #1028.

## Edit

This now uses a new `events2timeratio` function, calculating `sum(offset - onset)` directly without generating a segmentation. Ratio is now `sum(offset - onset) / (max(time) - min(time))` - time-based, not sample-based.

## Discussion

Isn't there a more memory efficient approach that does not need to generate a segmentation from the on- and offsets to obtain the same as `measure_events_ratio` achieves now? One could think of summing the spans/differences $t_{\mathrm{offset}}-t_{\mathrm{onset}}$ without generating a segmentation that can be significantly long.

~~Regarding the time conversion: I assumed the onset and offset are in seconds, with `sample_rate` this would convert the time into samples. If the onset and offset are using e.g. miliseconds, one might also avoid a complicated conversion.~~

Depending on the implemention to be merged on `events2segmentation` from #1447, `measure_events_ratio()` can still be changed to fit.

## Implemented changes

In `src/pymovements/gaze/gaze.py`:

- [x] Add generic `measure_events_ratio()` method to `Gaze` class for calculating event coverage ratios
  - [x] Using `time_column`
  - [x] Returns `polars.Expr`, allowing usage with `select()`, `with_columns()`, or `group_by().agg()`
  - [x] ratio as `sum(segmentation) / num_samples` (`.mean`)
  - [x] Group calculations by trial columns when present, uses `self.trial_columns` automatically.
- [ ] Replace parsing-based `data_loss_ratio_blinks`. Just use `measure_events_ratio('blink')`.
- [x] Exact name matching (not prefix) (e.g., 'blink', 'saccade', 'fixation')

```python
>>> import polars as pl
... import pymovements as pm
... 
... # Initialize Gaze with 8 samples and two blink events
... # Events cover time points [1, 2] and [5, 6] (offsets are exclusive - to be changed in events2segmentation)
... gaze = pm.Gaze(
...     samples=pl.DataFrame({
...         'time': [0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0],
...         'pixel': [[i, i] for i in range(8)],
...     }),
...     events=pm.Events(
...         name=['blink', 'blink'],
...         onsets=[1.0, 5.0],
...         offsets=[3.0, 7.0],
...     ),
... )
... gaze.samples.select(gaze.measure_events_ratio('blink'))
... 
shape: (1, 1)
shape: (1, 1)
┌───────────────────┐
│ event_ratio_blink │
│ ---               │
│ f64               │
╞═══════════════════╡
│ 0.571429          │
└───────────────────┘
```

Before feedback:
```python
>>> import polars as pl
... import pymovements as pm
... 
... # Samples with a trial column
... samples = pl.DataFrame({
...     'time': [0.0, 1.0, 0.0, 1.0],
...     'trial': [1, 1, 2, 2],
... })
... events = pm.Events(pl.DataFrame([
...     {'name': 'blink', 'onset': 1.0, 'offset': 2.0, 'trial': 1},
... ]))
... 
... gaze = pm.Gaze(samples=samples, events=events, trial_columns=['trial'])
... 
... # Group by trial and aggregate the ratio
... gaze.samples.group_by('trial', maintain_order=True).agg(
...     gaze.measure_events_ratio('blink')
... )
... 
src/pymovements/gaze/gaze.py:2176: UserWarning: Gaze contains samples but no components could be inferred. 
This usually happens if you did not specify any column content and the content could not be autodetected from the column names. 
Please specify 'pixel_columns', 'position_columns', 'velocity_columns' or 'acceleration_columns' explicitly during initialization. Otherwise, transformation methods may fail.
  warn(
shape: (2, 2)
┌───────┬───────────────────┐
│ trial ┆ event_ratio_blink │
│ ---   ┆ ---               │
│ i64   ┆ f64               │
╞═══════╪═══════════════════╡
│ 1     ┆ 0.5               │
│ 2     ┆ 0.0               │
└───────┴───────────────────┘
>>> gaze.samples.select(gaze.measure_events_ratio('saccade'))
shape: (1, 1)
┌─────────────────────┐
│ event_ratio_saccade │
│ ---                 │
│ f64                 │
╞═════════════════════╡
│ 0.0                 │
└─────────────────────┘
```

After feedback:
```python
>>> import polars as pl
... import pymovements as pm
... # Samples with a trial column
... samples = pl.DataFrame({
...     'time': [0.0, 1.0, 0.0, 1.0],
...     'trial': [1, 1, 2, 2],
... })
... events = pm.Events(pl.DataFrame([
...     {'name': 'blink', 'onset': 1.0, 'offset': 2.0, 'trial': 1},
... ]))
... gaze = pm.Gaze(samples=samples, events=events, trial_columns=['trial'])
... # Group by trial and aggregate the ratio - now needs .mean()
... gaze.samples.group_by('trial', maintain_order=True).agg(
...     gaze.measure_events_ratio('blink').mean()
... )
... 
src/pymovements/gaze/gaze.py:2315: UserWarning: Gaze contains samples but no components could be inferred. 
This usually happens if you did not specify any column content and the content could not be autodetected from the column names. 
Please specify 'pixel_columns', 'position_columns', 'velocity_columns' or 'acceleration_columns' explicitly during initialization. Otherwise, transformation methods may fail.
  warn(
shape: (2, 2)
┌───────┬───────────────────┐
│ trial ┆ event_ratio_blink │
│ ---   ┆ ---               │
│ i64   ┆ f64               │
╞═══════╪═══════════════════╡
│ 1     ┆ 1.0               │
│ 2     ┆ 0.0               │
└───────┴───────────────────┘
>>> gaze.samples.select(gaze.measure_events_ratio('saccade'))
shape: (1, 1)
┌─────────────────────┐
│ event_ratio_saccade │
│ ---                 │
│ f64                 │
╞═════════════════════╡
│ 0.0                 │
└─────────────────────┘
```

This should replace the parsing-based event ratio calculations.

## How Has This Been Tested?

- [x] Add tests to `tests/unit/measure/samples/measures/event_ratio_test.py` for 100% coverage
  - normal tests
  - error handling
  - edge cases (e.g. start/end time on an active event)
  - using trials

## Type of change

Remove irrelevant items:
- Internal

## Context

Resolves #

#### related issues:
- #24
- #893

#### required by:
- #1082

#### requires:
- [x] ~~#1447~~

#### conflicts with:
- #

## Checklist:


src/pymovements/events/segmentation.py
```diff
@@ -219,6 +219,137 @@ def events2segmentation(
...
+    ... })
+    >>> samples = pl.DataFrame({'time': [0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0]})
+    >>> samples.select(events2timeratio(events, samples, 'blink'))
+    shape: (1, 1)
+    ┌───────────────────┐
+    │ event_ratio_blink │
+    │ ---               │
+    │ f64               │
+    ╞═══════════════════╡
+    │ 0.571429          │
```
dkrako: this is 4/7 but actually I would expect 2/8 (8 time steps, only timestep 0 and 4 is not covered by blinks).

So it seems the error is on two levels:
- end timestep of event is not treated as inclusive to event
- total time is just using subtraction (offset - onset) but this ignores the actual duration of the last sample.

both problems would probably need passing the sampling rate to get the sampling interval.

src/pymovements/events/segmentation.py
```diff
@@ -219,6 +219,156 @@ def events2segmentation(
...
+        raise ValueError(f"Time column {time_column!r} not found in samples.")
+
+    relevant_events = events.filter(pl.col('name') == name)
+
+    if relevant_events.is_empty():
+        return pl.lit([0.0]).list.sum().alias(f"event_ratio_{name}")
+
+    dt_ms = 0.0
+    if sampling_rate is not None:
+        dt_ms = 1000.0 / sampling_rate
```
dkrako, line 331: After some consideration I think if sampling rate is `None` it would be best to just use the mode of the time differences like this: `dt_ms = pl.col(time_column).diff().mode()`.

We would then have two edge cases:
- empty series -> return null
- one element series -> return 1.0 if there's an event for that sample, return 0.0 if there's no event

Two different definitions regarding inclusive/exclusive timesteps might be too confusing and the differences might be unexpected.

cbueth: I added this in the last four commits.

cbueth: See diff here: https://github.com/pymovements/pymovements/pull/1452/changes/e7ff6a5930e97d1e9bb76421bfb0bcea1c5c731b..4507b6f44c1c7e73739a5b7cb3e7c8bb0330b1d1

PR #1464, opened 2026-01-30
 fix: avoid `KeyError` in EyeLink parsing in case of missing patterns

cbueth:
## Description

`parse_eyelink` breaks with some `.asc` files: This PR fixes a `KeyError` in the EyeLink ASC parser (`parse_eyelink`) that occurs when using custom `RECORDING_PATTERNS`.

The issue was triggered when an event end marker (like `EFIX`, `ESACC`, or `EBLINK`) was processed but the associated context dictionary was missing one or more keys defined in the `patterns` argument. This happens in two scenarios:
1. An "orphaned" event end marker appears without a preceding start marker.
2. An event occurs early in the file, before all defined message patterns (e.g., trial info, stimulus names) have been encountered and matched.

The probematic file starts with a long period of gaze data and events (`EFIX L`, `ESACC L`) **before** the first `start_recording_` message that would populate the `activity`, `trial`, `stimulus`, etc. columns. Because these events appear before the patterns match, `current_additional` is empty or incomplete, leading to the `KeyError: 'activity'` (or other events types) when processing these early events.

## Implemented changes

- [x] Use `.get(additional_column)` defaulting to `None` instead of direct indexing`[additional_column]` when appending values to the `events` dictionary.

## How Has This Been Tested?

- [x] Tested against the problematic file `116slsi1.asc` and `RECORDING_PATTERNS`. It now works without the previous `KeyError: 'trial'`. (flaky, can be other event types, too)
- [x] Existing unit tests continue to pass.

I attempted and failed writing a minimal automated regression. The real data is the most reliable confirmation.

<details>
<summary>Using `RECORDING_PATTERNS` </summary>

```python
RECORDING_PATTERNS = [
    r"start_recording_(?P<trial>(?:PRACTICE_)?trial_\d+)_stimulus_(?P<stimulus>[^_]+_[^_]+_\d+(\.0)?)_(?P<page>.+)",
    r"start_recording_(?P<trial>(?:PRACTICE_)?trial_\d+)_(?P<page>familiarity_rating_screen_\d+|subject_difficulty_screen)",
    {"pattern": r"stop_recording_", "column": "trial", "value": None},
    {"pattern": r"stop_recording_", "column": "page", "value": None},
    {
        "pattern": r"start_recording_(?:PRACTICE_)?trial_\d+_stimulus_[^_]+_[^_]+_\d+(\.0)?_page_\d+",
        "column": "activity",
        "value": "reading",
    },
    {
        "pattern": r"start_recording_(?:PRACTICE_)?trial_\d+_stimulus_[^_]+_[^_]+_\d+(\.0)?_question_\d+",
        "column": "activity",
        "value": "question",
    },
    {
        "pattern": r"start_recording_(?:PRACTICE_)?trial_\d+_(familiarity_rating_screen_\d+|subject_difficulty_screen)",
        "column": "activity",
        "value": "rating",
    },
    {"pattern": r"stop_recording_", "column": "activity", "value": None},
    {
        "pattern": r"start_recording_PRACTICE_trial_",
        "column": "practice",
        "value": True,
    },
    {
        "pattern": r"start_recording_trial_",
        "column": "practice",
        "value": False,
    },
    {"pattern": r"stop_recording_", "column": "practice", "value": None},
]
```

</details>

## Type of change

Remove irrelevant items:
- Bug fix

## Context

Might try a more robust, two-pass architecture.

Resolves #

#### related issues:
- #

#### required by:
- #

#### requires:
- [ ] #

#### conflicts with:
- #

## Checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,8 @@ on:
 jobs:
   tests:
     uses: ./.github/workflows/tests.yml
+    secrets:
+      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
   benchmark:
     uses: ./.github/workflows/benchmark.yml
--- a/.github/workflows/datasets.yml
+++ b/.github/workflows/datasets.yml
@@ -1,7 +1,7 @@
 name: datasets
 
 on:
-  pull_request:
+  pull_request_target:
     branches: [main]
   push:
     tags:
@@ -89,7 +89,7 @@ jobs:
   docs:
     name: docs
     runs-on: ubuntu-latest
-    if: ${{ github.event_name == 'pull_request' }}
+    if: ${{ github.event_name == 'pull_request_target' }}
     steps:
       - name: Checkout pymovements repository
         uses: actions/checkout@v4
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,6 +4,9 @@ on:
   pull_request:
     branches: [main]
   workflow_call:
+    secrets:
+      CODECOV_TOKEN:
+        required: true
   workflow_dispatch:
 
 concurrency:


dkrako (changes requested):
Thanks for the changes! Can you add a test in https://github.com/pymovements/pymovements/blob/fix/parse-eyelink/tests/unit/gaze/io/asc_test.py to prevent regressions?

tests/unit/gaze/io/asc_test.py
```diff
@@ -901,3 +901,27 @@ def test_from_asc_keeps_remaining_metadata_private_and_pops_cal_val(make_example
...
+    """
+    body = (
+        'EFIX R 1000 1100 100 500.0 500.0 1000\n'
+        'MSG 1200 START_TRIAL_1\n'
+    )
+    patterns = [r'START_TRIAL_(?P<trial_id>\d+)']
+    filepath = make_text_file(filename='orphaned_event.asc', body=body)
+
+    with pytest.warns(UserWarning, match='Missing start marker before end for event'):
+        from_asc(filepath, patterns=patterns)
```
dkrako: Something is strange here. I checked the output and `gaze.events` looks like this:

```
(Pdb) gaze.events
shape: (0, 4)
┌──────┬───────┬────────┬──────────┐
│ name ┆ onset ┆ offset ┆ duration │
│ ---  ┆ ---   ┆ ---    ┆ ---      │
│ str  ┆ i64   ┆ i64    ┆ i64      │
╞══════╪═══════╪════════╪══════════╡
└──────┴───────┴────────┴──────────┘
```

Shouldn't there be a fixation with onset `1000` and offset `1100` (and `trial_id` `None`) ?

Please add an `assert` at the end of the function to check that `gaze.events` has expected content.

cbueth: Added in 1b04d712, thanks.

dkrako: Thank you, maybe we should consider changing the default to `True`.

cbueth: I think so, too.

PR #1465, opened 2026-01-30
feat: two-pass architecture for context mapping in eyelink parser

cbueth:
## Description

Introduce a more robust, two-pass architecture for parsing EyeLink ASCII files. The motivation is to address state corruption issues (current code has global state `current_additional` that gets corrupted by pattern processing order -> race condition) and improve the handling of complex event sequences, such as overlapping or nested events, which the previous single-pass state machine could not reliably process. While not being a rewrite and still keeping the general structure, I hope this separation between data collection and event construction improves a bit on the readability/maintability of the code, removing the state dependency.

## Implemented changes

- [x] New Parser Logic: Replaced single-pass state machine with a two-pass pipeline, this makes the event matching deterministic. At construction time all events are available.
  1. `_collect_context_and_events`: iterate through the file and build separate timelines for raw event markers (starts and ends) and metadata.
  2. `_match_events_with_context`: pair event starts and ends correctly per eye and event type (stack-based approach, $O(N)$ complexity).
- [x] Context Merging: Keep more information (timestamp) at onset and offset, for better matching. While keeping track of trial/task information.
- [x] Orphan Handling: Better resilience to "orphaned" end markers (ends without starts) by falling back to the context at the end timestamp and emitting warnings.
- [x] Extract helper functions `_check_patterns` and `_config_inconsistent`
- [x] Update `parse_eyelink_event_start` to include timestamps for better tracking

## How Has This Been Tested?

- [x] Match existing tests
- [x] Before changing implementation, add more edge cases: out-of-order or missing event markers
  - passing before and after changes
- [x] Add tests for full coverage

## Type of change

- New functionality
- Performance improvement (safety/readability)

## Context

Resolves #

#### related issues:
- #

#### required by:
- #

#### requires:
- [x] #1464

#### conflicts with:
- #

## Checklist:


src/pymovements/gaze/_utils/_parsing_eyelink.py
```diff
@@ -424,6 +555,20 @@ def parse_eyelink(
 
     compiled_metadata_patterns.extend(EYELINK_META_REGEXES)
 
+    context_timeline, event_starts, event_ends = _collect_context_and_events(
```
dkrako: as the three return values are only used as an input to `_match_events_with_context()` readability will be improved by calling `_collect_context_and_events()`  within `_match_events_with_context()`.

However, testing `_match_events_with_context()` would probably get more complex that way so I'm indifferent.

src/pymovements/gaze/_utils/_parsing_eyelink.py
```diff
@@ -309,6 +309,146 @@ def parse_eyelink_event_end(line: str) -> tuple[str, str, float, float] | None:
...
+                context.update(match.groupdict())
+    return context
+
+
+def _collect_context_and_events(
+        lines: list[str],
+        compiled_patterns: list[dict[str, Any]],
+        additional_columns: set[str],
+) -> tuple[
+    dict[float, dict[str, Any]], list[tuple[str, str, float]],
```
dkrako: the return value is quite complex so it should be documented in the docstring

src/pymovements/gaze/_utils/_parsing_eyelink.py
```diff
@@ -309,6 +309,146 @@ def parse_eyelink_event_end(line: str) -> tuple[str, str, float, float] | None:
...
+    return context_timeline, event_starts, event_ends
+
+
+def _match_events_with_context(
+        event_starts: list[tuple[str, str, float]],
+        event_ends: list[tuple[str, str, float, float]],
+        context_timeline: dict[float, dict[str, Any]],
+        additional_columns: set[str],
+) -> list[dict[str, Any]]:
+    """Second pass: match starts/ends and build complete events using a stack per eye and type."""
```
dkrako: This comment confuses me. I understand the sample parsing as being the second pass.

src/pymovements/gaze/_utils/_parsing_eyelink.py
```diff
@@ -309,6 +309,146 @@ def parse_eyelink_event_end(line: str) -> tuple[str, str, float, float] | None:
...
+    event_starts: list[tuple[str, str, float]] = []
+    event_ends: list[tuple[str, str, float, float]] = []
+
+    current_additional: dict[str, Any] = {}
+    for col in additional_columns:
+        current_additional[col] = None
+
+    for line in lines:
+        # 1. Check pattern matches and update current_additional
+        matched_context = _check_patterns(line, compiled_patterns)
```
dkrako: I don't like it that much that the patterns are matched twice (second time in the for loop inside of `_parse_eyelink()`).

src/pymovements/gaze/_utils/_parsing_eyelink.py
```diff
@@ -309,6 +309,146 @@ def parse_eyelink_event_end(line: str) -> tuple[str, str, float, float] | None:
...
+    """First pass: collect all context and events separately."""
+    context_timeline: dict[float, dict[str, Any]] = {}
+    event_starts: list[tuple[str, str, float]] = []
+    event_ends: list[tuple[str, str, float, float]] = []
+
+    current_additional: dict[str, Any] = {}
+    for col in additional_columns:
+        current_additional[col] = None
+
+    for line in lines:
```
dkrako: I'm not sure if having two for loops over `lines` is the correct choice (the second one for the samples).

I think it would be better to keep a single for loop of `lines`, and instead call your `_match_events_with_context` after the loop terminates in `_parse_eyelink()`.

src/pymovements/gaze/_utils/_parsing_eyelink.py
dkrako: However, if we get rid of `_collect_context_and_events()` and call that function after the for loop in `_parse_eyelink()` terminates it would actually be the second pass again.

src/pymovements/gaze/_utils/_parsing_eyelink.py
dkrako: not relevant if the function is removed

src/pymovements/gaze/_utils/_parsing_eyelink.py
dkrako: not relevant if the function is removed

src/pymovements/gaze/_utils/_parsing_eyelink.py
```diff
@@ -309,6 +309,130 @@ def parse_eyelink_event_end(line: str) -> tuple[str, str, float, float] | None:
...
+                start_ts = stack.pop()  # (last-in-first-out)
+                # For EyeLink, onset in EFIX/ESACC/EBLINK line should match start_ts from
+                # SFIX/SSACC/SBLINK but, use the onset from the 'end' line
+                # as it's more complete (duration etc)
+                event_onset = onset
+            else:
+                # Orphaned end: no matching start
+                warnings.warn(
+                    f"Missing start marker before end for event '{event_name}' "
+                    f"(onset={onset}, offset={offset}). "
```
dkrako, line 399: This warning would be more helpful if it could include the line number (and maybe also the parsed line).

But that's clearly out of scope of this PR so I'll write an issue for that.

src/pymovements/gaze/_utils/_parsing_eyelink.py
```diff
@@ -309,6 +309,130 @@ def parse_eyelink_event_end(line: str) -> tuple[str, str, float, float] | None:
...
+                event_onset = onset
+                start_ts = onset  # fallback
+
+            # Build event context: prefer end context, but fill with start context
+            start_context = context_timeline.get(start_ts, {})
+            end_context = context_timeline.get(offset, {})
+
+            # Current behaviour: if it was a matched event, it uses context from the start
+            # If it was unmatched, it uses context from END.
+            # One can choose a different behaviour, but this on is conserving previous impl.
```
dkrako:
```suggestion
```

dkrako (approved):
Excellent, thanks a lot for taking care!

src/pymovements/gaze/_utils/_parsing_eyelink.py
dkrako: created #1501

PR #1469, opened 2026-02-04
chore!: remove deprecated utils submodules

dkrako:
## Description

Some of the `utils` modules are scheduled for removal in `v0.26.0`.

## Implemented changes

- [x] remove `utils.aois`
- [x] remove `utils.archives`
- [x] remove `utils.downloads`
- [x] remove `utils.filters`
- [x] remove `utils.parsing`
- [x] remove `utils.paths`
- [x] remove `utils.strings`

### DIFF
--- a/src/pymovements/stimulus/text.py
+++ b/src/pymovements/stimulus/text.py
@@ -325,7 +325,7 @@ def from_file(
         Name of the column which contains the page information of the area of interest.
         (default: None)
     trial_column: str | None
-        Name fo the column that specifies the unique trial id.
+        Name for the column that specifies the unique trial id.
         (default: None)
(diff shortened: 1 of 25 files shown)


src/pymovements/stimulus/text.py
```diff
@@ -299,10 +299,6 @@ def _get_aoi(
     If `end_x_column` is used, calculation: start_x_column <= x_eye < end_x_column.
     Analog for y coordinate and height.
 
-    .. deprecated:: v0.21.1
```
dkrako, line 366: this must've been a copy-paste mistake from `utils/aois.py` as it's a private function and not intended to be used by the user anyway

cbueth (approved):

Issue #865, opened 2024-10-23
measures for evaluating event detection


dkrako:
## Description of the problem

When running event detection algorithms we want to have some quality measure of the resulting event data.
Algorithm and parameter selection potentially depends on the eye tracker, recording setup, ambient conditions and overall data quality. Event detection that works well on one dataset is thus not guarenteed to transfer to other datasets.

Usually there's no annotated ground truth event data available, so we are limited to heuristic measures that expose potential issues in detected events.

## Description of a solution

I propose the following heuristic measures for event detection quality that are agnostic to stimulus data:

- peak / average velocity [dva/s] of fixations
- peak / average dispersion [dva or px] of fixations
- average durations [ms] of fixations and saccades
- proportion of unclassified samples

If taking account of stimulus data like AOIs, following measures are possible:

- proportion of fixations in the background (not on explicit AOI)

Further, experiments can include validation before trials, where participants are instructed to fixate a sequence of points. If raw data is available during these validation sequences, the locations and timings of the presented points can be assumed as ground truth (taking reaction delay of participant into account)

- use validation points as ground truth

PR #1476, opened 2026-02-09
feat: add precision measures for fixations

### LINKED ISSUE #1444: precision metrics for fixations
## Description of the problem

We currently lack a broader range of precision metrics for fixations.

To recap: precision measures the variance or "spread" of recorded values while the real world measured value is kept constant. As real-world eye tracking data recorded with real eyes cannot provide a perfectly still constant value (although artificial eyes can be used for that), fixations are the closest we get to constant real-world values.

Moreover, precision metrics for fixations allow not just for the data quality evaluation of raw gaze samples but also fixation detection algorithms and parameters, so this is basically a subtask of #865 .

As researchers should always provide the precision metrics of their detected fixations we should also include this functionality in pymovements.

## Description of a solution

Currently we have a single sample measure that can be used for measuring precision: https://github.com/pymovements/pymovements/blob/ecc88d52b2a5218d07301ba9a6fda11a7eed2014/src/pymovements/measure/samples/measures.py#L156-L197

There are at least two more measures that we should definitely add:

- standard deviation. this is pretty self-explanatory I think, as it's the default measure in statistics for spread and useful as it relates to the mean (which is the fixation location).
- S2S-RMS: sample to sample root mean square sum. This calculates the RMS of the difference between successive samples.
- BCEA (bivariate contour ellipse area) page 5 of _Niehorster et al._

There are a few more candidates, but they have drawbacks :
- variance
- RMS

- implement and test standard deviation sample measure
- implement and test s2s-rms sample measure

## Context

Here's good reference for fixation precision measures:

[Niehorster et al.: Characterizing gaze position signals and synthesizing noise during fixations in eye-tracking data. Behav Res Methods. 2020.](https://doi.org/10.3758/s13428-020-01400-9)

cbueth:
## Description

Implements three precision metrics for fixations as described in issue #1444:
- `std_dev`: Standard deviation of gaze positions
- `s2s_rms`: Root mean square of sample-to-sample displacements
- `bcea`: Bivariate contour ellipse area

These metrics allow quantifying the precision (variance/spread) of gaze positions during fixations useful for:
- Evaluating eye tracker data quality
- Assessing fixation detection algorithms and parameters
- Comparing eye tracker precision across different systems

See [Niehorster et al.: Characterizing gaze position signals and synthesizing noise during fixations in eye-tracking data. Behav Res Methods. 2020.](https://doi.org/10.3758/s13428-020-01400-9) for details.

## Implemented changes

Added functions as `pl.Expr` with `@register_sample_measure` in `measure/samples/measures.py`:

- [x] Added `std_dev()` 
  - Computes radial standard deviation: `sqrt(STD_x^2 + STD_y^2)`
  - Uses sample standard deviation (ddof=1) (correct?)
  - Returns None for single samples since no variance can be measured
- [x] Added `s2s_rms()` function
  - Computes RMS of successive sample displacements
  - Returns None for single samples (no sample pairs available)
- [x] Added `bcea()` function
  - Computes bivariate contour ellipse area with configurable confidence level
  - Default confidence: 68.27% (equivalent to prob. mass of ±1σ)
  - Returns None when variance cannot be computed (insufficient samples or zero variance)
- [ ] Added examples to docstrings (not today 😞)

## How Has This Been Tested?

- [x] Each measure has their own test module in `tests/unit/measure/samples/measures/`

Fyi: `pytest tests/unit/measure/samples/measures/std_dev_test.py tests/unit/measure/samples/measures/s2s_rms_test.py tests/unit/measure/samples/measures/bcea_test.py -v`

## Type of change

Remove irrelevant items:
- New functionality
- Documentation update: added to measures.rst

## Context

Resolves #1444

#### related issues:
- #865

## Checklist:


src/pymovements/measure/samples/measures.py
```diff
@@ -402,6 +403,301 @@ def _check_has_two_componenents(n_components: int) -> None:
         raise ValueError('data must have exactly two components')
 
 
+@register_sample_measure
+def std_dev(
```
dkrako: As it's the rms of the standard deviation maybe let's call it `std_rms`?

src/pymovements/measure/samples/measures.py
```diff
@@ -402,6 +403,301 @@ def _check_has_two_componenents(n_components: int) -> None:
...
+    *,
+    position_column: str = 'position',
+    n_components: int = 2,
+) -> pl.Expr:  # noqa: D401 - imperative mood
+    r"""Standard deviation of gaze positions during a fixation.
```
dkrako: the reference to positions or fixations isn't really necessary here, as it can be applied on any sampled (two-component) data (e.g. velocities). I would prefer to keep the docstrings here more general and be more specific within the user guide where the measures are applied.

src/pymovements/measure/samples/measures.py
```diff
@@ -402,6 +403,301 @@ def _check_has_two_componenents(n_components: int) -> None:
...
+
+    The standard deviation (STD) measures the spatial spread of gaze positions
```
dkrako:
```suggestion
    The standard deviation (STD) measures the spatial spread of samples
```

src/pymovements/measure/samples/measures.py
```diff
@@ -402,6 +403,301 @@ def _check_has_two_componenents(n_components: int) -> None:
...
+    where :math:`x_i` and :math:`y_i` are the gaze positions for the
+    :math:`i`-th sample, and :math:`\bar{x}` and :math:`\bar{y}` are their
+    respective means. The STD is a radial measure representing the overall
+    spatial extent of the gaze positions around their centroid.
+
+    STD is relatively insensitive compared to RMS-S2S to displacement between
+    successive gaze positions, making it a good measure of spatial spread
+    rather than signal velocity. This makes STD particularly suitable for
+    quantifying the precision of eye trackers as it reflects the area over
+    which gaze positions are distributed during fixations.
```
dkrako: It's great that you added these comparison paragraphs to all the measures. However, as these are debatable I think this should reference publications where this is stated. 

Moreoever I think this would fit best inside the user guide on data quality measures. There we could compare the the precision measures and discuss their suitability, e.g. if you're using the I-DT algorithm or use positional data then std is more appropriate, if you're using the I-VT or microsaccades algorithm or only rely on velocity data RMS-S2S will be better suited.

The docstrings of the measure functions could then have a "see also" box at the bottom that links to the data quality page of the user guide.

src/pymovements/measure/samples/measures.py
```diff
@@ -402,6 +403,301 @@ def _check_has_two_componenents(n_components: int) -> None:
         raise ValueError('data must have exactly two components')
 
 
+@register_sample_measure
+def std_dev(
+    *,
+    position_column: str = 'position',
```
dkrako: this argument can simply be named `column` and passing it as a positional argument should be allowed

docs/source/bibliography.bib
```diff
@@ -647,3 +647,24 @@ @misc{GGTG
...
+  journal = {Behavior Research Methods},
+  publisher = {Springer Science and Business Media LLC},
+  author = {Niehorster,  Diederick C. and Zemblys,  Raimondas and Beelders,  Tanya and Holmqvist,  Kenneth},
+  year = {2020},
+  month = may,
+  pages = {2515–2534}
+}
+
+@inproceedings{Crossland2002,
+  title={The use of an infra-red eyetracker in the assessment of macular disease: a comparison with the scanning laser ophthalmoscope},
```
dkrako: I'm not sure if this is the correct paper. Wouldn't this be the one? 

https://pubmed.ncbi.nlm.nih.gov/12462542/

docs/source/bibliography.bib
```diff
@@ -647,3 +647,24 @@ @misc{GGTG
...
+  author={Crossland, MD and Culham, LE and Kabanarou, SA and Rubin, GS},
+  booktitle={7th International Conference on Low Vision},
+  year={2002}
```
dkrako: please include the DOI

src/pymovements/measure/samples/measures.py
```diff
@@ -402,6 +403,301 @@ def _check_has_two_componenents(n_components: int) -> None:
...
+    To choose the ``confidence`` :math:`P` based on the familiar 1D “±σ” coverage, first compute
+    :math:`P = \Pr(|Z| \le \sigma) = \chi^2_1(\sigma^2)` (since :math:`Z^2 \sim \chi^2_1`),
+    then use the same probability mass to get :math:`k = \chi^2_2(P)`.
+
+    >>> from scipy.stats import chi2
+    >>> sigma = 1.0
+    >>> confidence = chi2.cdf(sigma*sigma, df=1)
+    >>> k = chi2.ppf(confidence, df=2)
+    >>> confidence, k
+    (np.float64(0.6826894921370859), np.float64(2.295748928898636))
```
dkrako: Weird, this fails the doctest in the minimum requirements:

```
_____________ [doctest] pymovements.measure.samples.measures.bcea ______________
592     the same probability mass as a ±1σ interval in 1D.
593     To choose the ``confidence`` :math:`P` based on the familiar 1D “±σ” coverage, first compute
594     :math:`P = \Pr(|Z| \le \sigma) = \chi^2_1(\sigma^2)` (since :math:`Z^2 \sim \chi^2_1`),
595     then use the same probability mass to get :math:`k = \chi^2_2(P)`.
596 
597     >>> from scipy.stats import chi2
598     >>> sigma = 1.0
599     >>> confidence = chi2.cdf(sigma*sigma, df=1)
600     >>> k = chi2.ppf(confidence, df=2)
601     >>> confidence, k
Expected:
    (np.float64(0.6826894921370859), np.float64(2.295748928898636))
Got:
    (0.6826894921370859, 2.295748928898636)
```

This might fix that if you include a `# doctest:+ELLIPSIS` in the line above:

```suggestion
    (... 0.6826894921370859 ..., ... 2.295748928898636 ...)
```

src/pymovements/measure/samples/measures.py
```diff
@@ -402,6 +403,301 @@ def _check_has_two_componenents(n_components: int) -> None:
...
+    r"""Bivariate contour ellipse area (BCEA) of gaze positions during a fixation.
+
+    The Bivariate Contour Ellipse Area (BCEA) :cite:p:`Crossland2002`
+    quantifies the area covered by gaze positions during a fixation.
+    It represents the area of an ellipse that encompasses
+    a specified proportion of all gaze position samples.
+    BCEA accounts for both the spread of gaze positions and their correlation between the horizontal
+    and vertical axes:
+
+    .. math::
```
dkrako, line 560: this should formulated as five separate equations on separate lines. currently you have to scroll horizontally to get the complete equation

src/pymovements/measure/samples/measures.py
```diff
@@ -402,6 +403,301 @@ def _check_has_two_componenents(n_components: int) -> None:
...
+
+    where :math:`x_i` and :math:`y_i` are the gaze positions, :math:`\bar{x}`
+    and :math:`\bar{y}` are their respective means, :math:`\sigma_x` and
+    :math:`\sigma_y` are the standard deviations, :math:`\rho` is the Pearson
+    correlation coefficient between the horizontal and vertical components,
+    :math:`P` is the desired confidence level (as a percentage), and :math:`k`
+    is the scaling factor derived from the chi-square distribution with 2
+    degrees of freedom :cite:p:`Niehorster2020`.
+
+    The relationship between confidence level :math:`P` and scaling factor
```
dkrako: the details on the confidence level should be moved to the Notes section

src/pymovements/measure/samples/measures.py
```diff
@@ -402,6 +403,301 @@ def _check_has_two_componenents(n_components: int) -> None:
...
+    r"""Root-mean-square of sample-to-sample displacements during a fixation.
+
+    The RMS-S2S (Root Mean Square - Sample to Sample) measures the magnitude
+    of displacements between successive gaze position samples. It is computed
+    as the square root of the mean squared Euclidean distance between all
+    adjacent sample pairs:
+
+    .. math::
+        \begin{align}
+        \text{RMS-S2S} &= \sqrt{\frac{1}{n-1} \sum_{i=1}^{n-1} \theta_i^2},\quad
```
dkrako: I think I prefer the first version with the theta, but the theta equation should get a separate line

src/pymovements/measure/samples/measures.py
```diff
@@ -402,6 +403,301 @@ def _check_has_two_componenents(n_components: int) -> None:
...
+    n_components: int = 2,
+) -> pl.Expr:  # noqa: D401 - imperative mood
+    r"""Standard deviation of gaze positions during a fixation.
+
+    The standard deviation (STD) measures the spatial spread of gaze positions
+    around their centroid. It is computed as the root mean square of the
+    squared standard deviations along the horizontal and vertical directions:
+
+    .. math::
+        \text{STD}_x = \sqrt{\frac{1}{n-1} \sum_{i=1}^{n} (x_i - \bar{x})^2},\quad
```
dkrako, line 419: here too I think I'd prefer one line per equation

src/pymovements/measure/samples/measures.py
cbueth: Rigt, I agree with you. For now, I have moved them to the bottom of the Notes section now, to be removed later and replaced with a See Also and the user guide page.

src/pymovements/measure/samples/measures.py
cbueth: I have turned it around, put it before the asterisk across all three metrics.

docs/source/bibliography.bib
cbueth: I was looking for that reference, as it is from the Niehorster paper when BCEA is introduced, but that paper only has a google scholars reference, which was this one. Thanks for finding the right one!

src/pymovements/measure/samples/measures.py
cbueth: Currently, you think it is too mixed up? I think this way the text above and below is not spread out even more. Probably my personal preference from Physics :D

src/pymovements/measure/samples/measures.py
cbueth: Same as above

src/pymovements/measure/samples/measures.py
cbueth: That's weird. It should be on three lines, but the `equation` environment does not seem to take new lines `\\`. I have changed it now to how it was intended to look like.

```rst
.. math::
        \begin{equation}
        \sigma_x = \sqrt{\frac{1}{n-1} \sum_{i=1}^{n} (x_i - \bar{x})^2},\quad
        \sigma_y = \sqrt{\frac{1}{n-1} \sum_{i=1}^{n} (y_i - \bar{y})^2}\\
        \rho = \frac{\sum_{i=1}^{n} (x_i - \bar{x})(y_i - \bar{y})}
        {\sqrt{\sum_{i=1}^{n} (x_i - \bar{x})^2}
        \sqrt{\sum_{i=1}^{n} (y_i - \bar{y})^2}}\\
        k = -2 \ln(1 - P/100),\quad
        \text{BCEA} = k \pi \sigma_x \sigma_y \sqrt{1 - \rho^2}
        \end{equation}
```

src/pymovements/measure/samples/measures.py
cbueth: I was not able to make `# doctest:+ELLIPSIS` work for me on the output line or before, so it is solved with using print.

Issue #971, opened 2025-02-25
set gaze samples to `null` during blink events


dkrako:
## Description of the problem

Gaze samples during blink events are often caused by artifacts and do not correspond to any real eye movements.

## Description of a solution

If we have blinks available, we can set all samples to `null` during these events.

## Context

This is related to #970 but instead of removing events this is applied to the raw samples.

AnnaBhlr:
I've checked in raw EyeLink data and it looks like the samples within the timestamps of SBLINK and EBLINK are nans anyways. That makes total sense as Eyelink defines blinks as periods of pupil loss. However, there is a short time span when the lid is already closing but the pupil is still detectable (same goes for eye opening at end of blink). That's causing the artefacts as the eye is often moving during a blink and would be ignored for further analysis. 

How about having an argument in `from_asc` indicating a threshold around the blink where data should be set to null? Something like 

```
data = pm.gaze.from_asc(
    file = file_path, 
    blink_threshold = 0.05) # either 'None' as default or float for a time in seconds
```

Another option would be to have it as a function for `gaze.frame`, wouldn't it?

dkrako:
Great that you digged deeper! You're right, the issue is more with the artifacts that appear shortly before or after blinks.

I just wouldn't like to add this as an argument to `from_asc()` as this is more related to preprocessing than to parsing. All in all, we plan to add blink detection methods at some time, so having a separate method would be more useful then.

The threshold values could also be given as a tuple in case someone wants to have a shorter threshold before the beginning and a longer after the end.

We would need to come up with a name though, and the ideas I had for could be improved.

- `GazeDataFrame.nullify_blink_samples()`
- `GazeDataFrame.set_blink_samples_to_null()`
- `GazeDataFrame.set_samples_to_null_during_blinks()`

PR #1498, opened 2026-02-18
feat(events): add out_of_screen event detection method

### LINKED ISSUE #1487: detect out of screen samples and classify them as noise
## Description of the problem

An eye tracker might output out-of-screen pixel positions. Usually they are not of interest and the user wants to set these to null.

## Description of a solution

We can implement an event detection method that checks for pixel coordinates that are outside of the given screen boundaries.

As we have the screen boundaries available via `Gaze.experiment.screen` this should be easy to implement.

As a further preprocessing step the samples during the detected out-of-screen events could be set to null, reusing the solution from #971

pandongpeng:
## Description

Add a new `out_of_screen` event detection method that identifies gaze samples with pixel coordinates outside the screen boundaries, grouping consecutive out-of-screen samples into events. Inspired by the `mark_trackloss` function from the VWPre R package.

Screen boundaries are automatically sourced from `Gaze.experiment.screen` when available, so users typically don't need to specify them manually.

Resolves #1487.

## Implemented changes

- [x] `out_of_screen()` detection function with `@register_event_detection`, supporting configurable `x_min`, `x_max`, `y_min`, `y_max` boundaries
- [x] `x_max` and `y_max` are required parameters (no defaults) to prevent silent errors when screen boundaries are not configured
- [x] Upper bounds are exclusive (`>=`), matching screen resolution semantics — for a 1920×1080 display, pixels `[0, 1920)` × `[0, 1080)` are valid
- [x] Auto-fill `x_min`, `x_max`, `y_min`, `y_max` from `self.experiment.screen` in `Gaze._fill_event_detection_kwargs` when using `gaze.detect('out_of_screen')`
- [x] Register `out_of_screen` in `EventDetectionLibrary` and public API exports
- [x] Add tutorial cells in `pymovements-in-10-minutes.ipynb` demonstrating usage with `ToyDatasetEyeLink`

## How Has This Been Tested?

- [x] 7 error/validation tests: `None` pixels, wrong shape, `x_min >= x_max`, `y_min >= y_max`, length mismatch
- [x] 10 detection tests: all within screen, inclusive min / exclusive max boundary, below/above each axis, consecutive events, timesteps, custom name, all out-of-screen
- [x] 4 integration tests via `Gaze.detect()`: auto-fill from experiment, all within screen, with timesteps, explicit boundary override
- [x] All existing tests pass (pytest, pylint, mypy, flake8 via pre-commit)

## Type of change

- New functionality
- Documentation update

## Context

Resolves #1487

#### related issues:
- #971 (blink sample nullification — separate PR #1505)

### Files changed

| Status | File | Purpose |
|--------|------|---------|
| 🆕 Added | `src/pymovements/events/detection/_out_of_screen.py` | Detection function |
| ✏️ Modified | `src/pymovements/events/detection/__init__.py` | Register and export |
| ✏️ Modified | `src/pymovements/events/__init__.py` | Public API export |
| ✏️ Modified | `src/pymovements/gaze/gaze.py` | Auto-fill screen boundaries |
| ✏️ Modified | `docs/source/tutorials/pymovements-in-10-minutes.ipynb` | Tutorial cells |
| 🆕 Added | `tests/unit/events/detection/out_of_screen_test.py` | Unit tests |
| ✏️ Modified | `tests/unit/gaze/detect_test.py` | Integration tests |
| ✏️ Modified | `tests/unit/events/detection/library_test.py` | Library registration test |

### DIFF
--- a/src/pymovements/gaze/gaze.py
+++ b/src/pymovements/gaze/gaze.py
@@ -1949,8 +1949,10 @@ class Gaze:
             The filled keyword argument dictionary.
         """
         # Automatically infer eye to use for event detection.
-        method_args = inspect.getfullargspec(
-            method).args + inspect.getfullargspec(method).kwonlyargs
+        method_args = (
+            inspect.getfullargspec(method).args
+            + inspect.getfullargspec(method).kwonlyargs
+        )
 
         if 'positions' in method_args:
             if 'position' not in samples.columns:


src/pymovements/events/detection/_out_of_screen.py
```diff
@@ -0,0 +1,137 @@
...
+    _checks.check_is_length_matching(pixels=pixels, timesteps=timesteps)
+
+    # A sample is out-of-screen if x or y is outside the screen boundaries or is NaN.
+    x = pixels[:, 0]
+    y = pixels[:, 1]
+
+    out_of_screen_mask = (
+        (x < x_min) | (x > x_max)
+        | (y < y_min) | (y > y_max)
+        | np.isnan(x) | np.isnan(y)
```
dkrako: I would prefer to exclude the `isnan` here to really only detect out of screen events.

For the more general tracker loss instead it would be better to create a separate `tracker_loss` event detection. This way it's easier to discern the causes of data quality degradation.

If you like you can create a similar PR for a `tracker_loss` event detection function.

src/pymovements/events/detection/_out_of_screen.py
```diff
@@ -0,0 +1,137 @@
...
+    )
+
+    # Get indices of out-of-screen samples.
+    candidate_indices = np.where(out_of_screen_mask)[0]
+
+    n_total = len(pixels)
+    n_out = len(candidate_indices)
+    if n_out > 0:
+        pct = round(n_out / n_total * 100, 2)
+        warnings.warn(
```
dkrako: you can remove the warning here as the reported values are already included in the returned `Events`.

src/pymovements/gaze/gaze.py
```diff
@@ -1989,6 +1989,25 @@ def _fill_event_detection_kwargs(
                 ],
             ).transpose()
 
+        if 'pixels' in method_args and 'pixels' not in kwargs:
```
dkrako, line 1995: great that you're already forwarding the column here!

could you also include a similar section for the `x_min`, `x_max` etc arguments?

You can source the values from `self.experiment.screen`.

However for that arguments it would probably be best to include an additional check like `method.__name__ == 'out_of_screen'` for forwarding the screen limits to make sure the forwarding is only applied for that function. 

maybe like this:

```python
if method.__name__ == 'out_of_screen':
    if 'x_min' not in kwargs:
        kwargs['x_min'] = 0
    if 'x_max' not in kwargs:
        kwargs['x_max'] = self.experiment.screen.width_px
    ...
```

This way a user won't need to specify the screen boundaries manually.

The forwarding of the pixel column is fine as is.

src/pymovements/events/detection/_out_of_screen.py
pandongpeng: Removed | np.isnan(x) | np.isnan(y) from the detection mask. The function now only flags samples with coordinates outside the screen boundaries. NaN handling will be a   separate tracker_loss function as the reviewer suggested.

src/pymovements/events/detection/_out_of_screen.py
pandongpeng: warning removed

src/pymovements/gaze/gaze.py
pandongpeng: now the results is like:
Screen resolution: 1280 x 1024 pixels
Valid pixel range: x=[0, 1280], y=[0, 1024]
Detecting events: 100%
2/2 [00:00<00:00, 5.42file/s]

Gaze[0]: 95 out-of-screen events detected
  7030/128342 samples (5.48%) marked as trackloss

Gaze[1]: 40 out-of-screen events detected
  3070/109216 samples (2.81%) marked as trackloss

pandongpeng (commented):
comments addressed

src/pymovements/events/detection/_out_of_screen.py
```diff
@@ -0,0 +1,121 @@
...
+        Continuous 2D pixel coordinate time series. The first column is the x coordinate
+        and the second column is the y coordinate.
+    timesteps: list[int] | np.ndarray | None
+        shape (N, )
+        Corresponding continuous 1D timestep time series. If None, sample based timesteps are
+        assumed. (default: None)
+    x_min: float
+        Minimum valid x pixel coordinate (inclusive). (default: 0)
+    x_max: float
+        Maximum valid x pixel coordinate (inclusive). For a 1920px wide screen, set to 1920.
```
dkrako: actually this operation should be exclusive, because the resolution given in [`Screen`](https://pymovements.readthedocs.io/en/stable/reference/api/pymovements.Experiment.html) would be 1920 (for eyelink files we increment that value during parsing to be consistent).

So for example, with a screen resolution of 1920x1080 the pixel screen positions (0,0), (1919,1079) and (1919.9,1079.9)  would be on the screen, the pixels (-1,-1)  and (1920, 1080)  would not.

src/pymovements/events/detection/_out_of_screen.py
```diff
@@ -0,0 +1,121 @@
...
+        timesteps = np.arange(len(pixels), dtype=np.int64)
+    timesteps = np.array(timesteps)
+    _checks.check_is_length_matching(pixels=pixels, timesteps=timesteps)
+
+    # A sample is out-of-screen if x or y is outside the screen boundaries.
+    x = pixels[:, 0]
+    y = pixels[:, 1]
+
+    out_of_screen_mask = (
+        (x < x_min) | (x > x_max)
```
dkrako: as in the comment above, this should be

```suggestion
        (x < x_min) | (x >= x_max)
```

src/pymovements/events/detection/_out_of_screen.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+@register_event_detection
+def out_of_screen(
+        pixels: list[list[float]] | list[tuple[float, float]] | np.ndarray,
+        timesteps: list[int] | np.ndarray | None = None,
+        x_min: float = 0,
+        x_max: float = 0,
+        y_min: float = 0,
+        y_max: float = 0,
```
dkrako: let's change the order and make some of the arguments required like this.

```python
def out_of_screen(
        pixels: list[list[float]] | list[tuple[float, float]] | np.ndarray,
        x_max: float,
        y_max: float,
        x_min: float = 0,
        y_min: float = 0,
        timesteps: list[int] | np.ndarray | None = None,
        name: str = 'out_of_screen',
) -> Events:
```

Having a default value for `x_max`/`y_max` could lead to silent weird errors if someone forgot to set `Gaze.experiment.screen`

dkrako (changes requested):
Excellent, thank you so much for adding all the tests!

I have two minor comments regarding making some of the arguments required and then the max values should be exclusive.

src/pymovements/events/detection/_out_of_screen.py
pandongpeng: Done. Changed upper bounds to exclusive (`>=`). Now `(1920, 1080)` is correctly detected as out-of-screen for a 1920x1080 display, while `(1919.9, 1079.9)` remains on-screen. Updated tests accordingly.

src/pymovements/events/detection/_out_of_screen.py
pandongpeng: Done. Changed to `x >= x_max` and `y >= y_max`.

src/pymovements/events/detection/_out_of_screen.py
pandongpeng: Done. Reordered parameters: \ and \ are now required positional arguments (no defaults), placed right after \. \ and \ remain optional with default 0. This prevents silent errors when screen boundaries aren't configured.

PR #1499, opened 2026-02-18
feat: add writing system field to text stimulus

kuribayashi4:
## Description

The TextStimulus class has been updated to include a writing-mode attribute (horizontal-lr, horizontal-rl, vertical-lr, vertical-rl) to support writing-direction-sensitive algorithms (e.g., detecting regression) in the future.
I also added several tests to ensure that the current implementation (AOI mapping) is indeed writing direction-agnostic.

## Implemented changes

- [x] Change 1 `src/pymovements/stimulus/text.py` is updated

## How Has This Been Tested?

I have added toy texts (`tests/files/stimuli/toy_text_aoi*`) that have been artificially modified to simulate different writing directions.
- [x] Test A `tests/unit/stimulus/text_test.py` is extended by loading different writing-direction data.
- [x] Test B `tests/unit/events/event_aoi_mapping_test.py` is extended to different writing directions. The same sample eye fixation data is used for all the tests, but the X/Y-axis is mirrored or transposed for different writing directions for now.
- [x] Test C `tests/unit/stimulus/text_writing_mode_test.py` is added just to check if the new attribute works properly

## Type of change

Remove irrelevant items:
- New functionality

## Context
#540

## Checklist:

### DIFF
--- a/src/pymovements/stimulus/text.py
+++ b/src/pymovements/stimulus/text.py
@@ -37,6 +37,7 @@ from pymovements._utils._html import repr_html
 
 
 @dataclass(frozen=True)
+@repr_html()
 class WritingSystem:
     """Writing system specification for text stimuli.
 
--- a/tests/unit/stimulus/text_writing_system_test.py
+++ b/tests/unit/stimulus/text_writing_system_test.py
@@ -111,3 +111,8 @@ def test_writing_system_from_descriptor(descriptor, expected):
     assert writing_system.directionality == expected['directionality']
     assert writing_system.axis == expected['axis']
     assert writing_system.lining == expected['lining']
+
+
+def test_writing_system_from_descriptor_unknown_raises_exception():
+    with pytest.raises(ValueError, match="Unknown descriptor 'test'"):
+        WritingSystem.from_descriptor('test')


src/pymovements/stimulus/text.py
```diff
@@ -25,13 +25,33 @@
...
 from typing import Any
+from typing import Literal
 
 import polars as pl
 
 from pymovements._utils import _checks
 from pymovements._utils._html import repr_html
 
 
+WritingMode = Literal[
```
dkrako: Maybe it would be better to create a new class for that and provide more info there. I just skimmed over https://en.wikipedia.org/wiki/Writing_system#Directionality_and_orientation and it seems there are three dimensions:

- axis
- lining
- directionality

A simple class like would be more explicit and flexible later on:

```python
@dataclass
class WritingSystem:
    axis: Literal['horizontal', 'vertical']
    lining: Literal['top-to-bottom', 'left-to-right', 'right-to-left']
    directionality: Literal['left-to-right', 'right-to-left', 'top-to-bottom']
```

src/pymovements/stimulus/text.py
```diff
@@ -23,15 +23,27 @@
...
 
 import polars as pl
 
 from pymovements._utils import _checks
 from pymovements._utils._html import repr_html
 
 
+@dataclass(frozen=True)
+class WritingSystem:
+    """Writing system specification used by :class:`TextStimulus`."""
```
dkrako: would you like to add a method for shorter initialization?

maybe something like this?

```python
@staticmethod
def from_descriptor(descriptor: str) -> WritingSystem:
    if descriptor in {'left-to_right', 'ltr'}:
        return WritingSystem(axis='horizontal', directionality='left-to-right', lining='top-to-bottom')
    if descriptor in {'right-to-left', 'rtl'}:
        return WritingSystem(axis='horizontal', directionality='right-to-left', lining='top-to-bottom')
    ...
    raise ValueError(
        f"Unknown descriptor '{descriptor}'. Valid descriptors are: {valid_descriptors}"
    )
```

This way it will be more convenient to initialize for the three most used cases (ltr, rtl, top-to-bottom)

src/pymovements/stimulus/text.py
```diff
@@ -79,6 +107,7 @@ def __init__(
             end_y_column: str | None = None,
             page_column: str | None = None,
             trial_column: str | None = None,
+        writing_system: WritingSystem = WritingSystem(),
```
dkrako: python is weird with having default values of mutable type: https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil

If you implement the proposed method above you could use a string as a default:

```suggestion
            writing_system: WritingSystem | str = 'left-to-right',
```

src/pymovements/stimulus/text.py
```diff
@@ -91,6 +120,7 @@ def __init__(
         self.end_y_column = end_y_column
         self.page_column = page_column
         self.trial_column = trial_column
+        self.writing_system = writing_system
```
dkrako:
```suggestion
        if isinstance(writing_system, str):
            self.writing_system = WritingSystem.from_descriptor(writing_system)
        else:
             self.writing_system = writing_system
```

src/pymovements/stimulus/text.py
```diff
@@ -64,6 +76,22 @@ class TextStimulus:
     trial_column: str | None
         Name for the column that specifies the unique trial id.
         (default: None)
+    writing_system: WritingSystem
+                Writing system of the text. Typical settings are:
+                - WritingSystem(axis='horizontal', lining='top-to-bottom', directionality='left-to-right')
```
dkrako: Great that you already include lots of examples! I think it would fit better inside the docstring of `WritingSystem`

dkrako (changes requested):
Great, thanks a lot for all the tests. I think it would be great if the `WritingSystem` class could get an additional method for more convenient initialization.

src/pymovements/stimulus/text.py
kuribayashi4: Thanks for very nice feedback!

src/pymovements/stimulus/text.py
kuribayashi4: okay, I understand what you suggested. I'll implement this

kuribayashi4: updated

src/pymovements/stimulus/text.py
kuribayashi4: updated

src/pymovements/stimulus/text.py
kuribayashi4: updated

src/pymovements/stimulus/text.py
kuribayashi4: updated

src/pymovements/stimulus/text.py
```diff
@@ -23,15 +23,141 @@
...
+    """Writing system specification for text stimuli.
+
+    Attributes
+    ----------
+    directionality: Literal['left-to-right', 'right-to-left', 'top-to-bottom']
+        Direction in which text flows within a line.
+        For horizontal text, this is typically 'left-to-right' or 'right-to-left'.
+        For vertical text, this is typically 'top-to-bottom'.
+        Bidirectional/Boustrophedon scripts (e.g., Arabic with embedded English)
+        is not currently supported. (default: 'left-to-right')
```
dkrako:
```suggestion
        is currently not supported.
        (default: 'left-to-right')
```

PR #1506, opened 2026-02-19
add filter_samples function to Dataset class

kuribayashi4:
## Description

drop events according to property/measure values https://github.com/orgs/pymovements/projects/9/views/1?pane=issue&itemId=157785412&issue=pymovements%7Cpymovements%7C1492

## Implemented changes

- [x] Change 1: added filter_samples function for the Dataset class. 

One can easily exclude data points according to some criteria. Currently, the function has five arguments:
- min_velocity: float | None
            Keep samples with velocity greater than or equal to this value. (default: None)
- max_velocity: float | None
            Keep samples with velocity less than or equal to this value. (default: None)
- min_duration: float | None
            Keep samples with duration greater than or equal to this value.
            Requires a ``duration`` column. (default: None)
- max_duration: float | None
            Keep samples with duration less than or equal to this value.
            Requires a ``duration`` column. (default: None)
- target_event: str | None
            If provided, keep only samples that belong to the specified event name based on
            :py:attr:`~pymovements.Gaze.events`. (default: None)

Note that if there are data points with null velocity/duration, and when the max/min velocity/duration is set, these null data points are automatically dropped. In this case, a warning message is shown to the user.     
Also, this function always displays how many samples were removed to ensure users are aware of the number of samples excluded.

Further note that if the velocity is given as a vector [vx, vy], this function calculates l2 norm.

## How Has This Been Tested?

- [x] dataset_test.py is updated.

## Type of change

- New functionality

## Context

Resolves https://github.com/orgs/pymovements/projects/9/views/1?pane=issue&itemId=157785412&issue=pymovements%7Cpymovements%7C1492

#### related issues:
- #

#### required by:
- #

#### requires:
- [ ] #

#### conflicts with:
- #

## Checklist:


src/pymovements/dataset/dataset.py
```diff
@@ -834,6 +836,168 @@ def detect_events(
...
+        if min_duration is not None and max_duration is not None and min_duration > max_duration:
+            raise ValueError('min_duration must be less than or equal to max_duration')
+
+        self._check_gaze()
+
+        n_removed_invalid_velocity_norm = 0
+        n_removed_invalid_duration = 0
+        n_filtered_samples = 0
+
+        for gaze in self.gaze:
```
dkrako, line 894: the logic inside the for loop should be implemented inside `Gaze.filter_samples()`

dkrako (changes requested):
Thanks a lot for your work! It's nice to have the method available for the `Dataset` class, however it is important to also implement this as a `Gaze` method. The `Gaze.filter_samples()` method can then be reused in `Datazset.filter_samples()`.

cbueth:
Hello @kuribayashi4, what is the status on this PR? We would be happy to integrate this feature. Please let us know if you are able to integrate the feedback anytime soon, need some details/help or want to proceed differently. Take care!

PR #1522, opened 2026-03-04
fix: avoid irrelevant unnest warnings

dkrako:
## Description

In case a user unnested all columns before saving they would get the following warning:

```
 Warning: No columns to unnest. Please specify columns to unnest via the "input_columns" argument.
```

This PR only unnests in `Gaze.save_samples()` if there are columns to unnest and does not clone the `Gaze`.

## Implemented changes

- [x] don't call `Gaze.clone()` in `Gaze.save_samples()`. Otherwise, this would warn that no component columns (e.g. `pixel_columns`)  are specified.
- [x] only unnest in `Gaze.save_samples()` if there are nested columns and extension is csv.
- [x] refactor unnest logic into `_unnest_list_columns()` (out-of-place operation) to make it reusable in `Gaze.save_samples()` without changing the `Gaze` object
- [x] check for nested columns to avoid unnest warning in `map_to_aois()`

## How Has This Been Tested?

- [x] test that no warnings are emitted if saved `Gaze.samples` do not include nested columns
- [x] remove filterwarnings from tests in `gaze_unnest_test.py`

## Checklist:

### DIFF
--- a/src/pymovements/gaze/gaze.py
+++ b/src/pymovements/gaze/gaze.py
@@ -1579,7 +1579,8 @@ class Gaze:
             if nested_columns:
                 try:
                     self.unnest(nested_columns)
-                except (ValueError, AttributeError):  # tolerate common cases
+                except (ValueError, AttributeError):  # pragma: no cover
+                    # tolerate common cases
                     # - ValueError/AttributeError: shape or configuration-related issues
                     # In all these cases: continue without failing and use fallback logic.
                     pass


tests/unit/gaze/gaze_unnest_test.py
```diff
@@ -427,55 +427,49 @@ def test_gaze_unnest_has_expected_frame_multiple_unnest(
...
-                'wrong_acceleration_name': [[1.23, 4.56]],
-            }),
-            {},
-            Warning,
-            'No columns to unnest. '
-            'Please specify columns to unnest via the "input_columns" argument.',
-            marks=pytest.mark.filterwarnings(
-                'ignore:Gaze contains samples but no.*:UserWarning',
-            ),
-            id='df_single_row_two_components_unnest_all_default_values_no_cols_to_unnest',
```
dkrako, line 444: this test case is rewritten as `test_gaze_unnest_no_nested_columns_warns` below.

the filtered warning is not raised anymore.

src/pymovements/gaze/gaze.py
```diff
@@ -1917,72 +1918,13 @@ def unnest(
         Warning
             If no columns to unnest exist and none are specified.
         """
-        if input_columns is None:
```
dkrako, line 1920: all that code is refactored into `_unnest_list_columns()`

src/pymovements/gaze/gaze.py
```diff
@@ -1575,13 +1575,15 @@ def map_to_aois(
...
-            except (Warning, ValueError, AttributeError):  # tolerate common cases
-                # - Warning: nothing to unnest when no list columns exist
-                # - ValueError/AttributeError: shape or configuration-related issues
-                # In all these cases: continue without failing and use fallback logic.
-                pass
+            nested_columns = _get_nested_columns(self.samples)
+            if nested_columns:
+                try:
+                    self.unnest(nested_columns)
+                except (ValueError, AttributeError):  # pragma: no cover
```
dkrako, line 1582: there do not seem to exist any tests that would raise these errors. I therefore added the `pragma: no cover` instruction here.

The warning is now handled by only unnesting if there are nested columns. I would say handling the other errors is out of scope of this PR but it might be worth evaluating if this `except` is still necessary here, especially if there aren't any tests that reproduce this. However I really don't know if removing the `except` would break some existing workflows so I would prefer to keep this as is in this PR to avoid breaking changes.

PR #1523, opened 2026-03-05
fix: add xlabel on every axis in `tsplot`

cbueth:
## Description

This PR simplifies adds missing `xlabel` in `tsplot`. Before it was only on the last plot, now it is on all axes. When `sharex=True` is used, Matplotlib should automatically hide extra labels for axes that are then touching.

### Before

data_image

### After

image

## Implemented changes

- [x] Modified `src/pymovements/plotting/tsplot.py` to set `xlabel` on all subplots.
- [x] Removed unneeded `plt.show()` in notebook.

## How Has This Been Tested?

- [x] Manually with the preceding examples.
- [x] With a script checking that it's hidden when using `sharex=True`.

## Type of change

Remove irrelevant items:
- Bug fix
- Documentation update

## Checklist:


docs/source/user-guide/inspect-raw-samples.ipynb
```diff
@@ -178,13 +174,13 @@
...
     "pm.plotting.tsplot(\n",
     "    gaze_unnested,\n",
+    "    xlabel='time [ms]',\n",
     "    channels=['position_x', 'position_y'],\n",
     "    share_y=False,\n",
     "    line_color=\"darkblue\",\n",
     "    n_rows=2, n_cols=1,\n",
-    "    zero_centered_yaxis=False)\n",
-    "\n",
-    "plt.show()"
```
dkrako, line 187: the plt.show() is necessary here, otherwise we get the `(<Figure size 1500x500 with 2 Axes>, <Axes: ylabel='position_x'>)` as you can see here: (<Figure size 1500x500 with 2 Axes>, <Axes: ylabel='position_x'>)

docs/source/user-guide/inspect-raw-samples.ipynb
```diff
@@ -237,9 +233,8 @@
...
```
dkrako, line 242: here too the `plt.show()` is necessary

src/pymovements/plotting/tsplot.py
```diff
@@ -224,8 +224,9 @@ def tsplot(
         else:
             ax.set_ylabel(channels[channel_id])
 
-    # print x label on the last axis (bottom-most in a column)
-    axs[-1].set_xlabel(xlabel)
+        # set x label on all axes
+        # sharex=True will automatically hide those that are not on the bottom
```
dkrako: sharex=True is not defined in the code. is it applied automatically?

docs/source/user-guide/inspect-raw-samples.ipynb
```diff
@@ -82,9 +80,7 @@
    },
    "outputs": [],
    "source": [
-    "pm.plotting.traceplot(gaze)\n",
-    "\n",
-    "plt.show()"
```
dkrako, line 87: plt.show() is necessary

docs/source/user-guide/inspect-raw-samples.ipynb
```diff
@@ -111,12 +107,12 @@
     "\n",
     "pm.plotting.tsplot(\n",
     "    gaze_unnested,\n",
+    "    xlabel='time [ms]',\n",
     "    channels=['pixel_x', 'pixel_y'],\n",
     "    share_y=False,\n",
     "    line_color='darkblue',\n",
-    "    zero_centered_yaxis=False)\n",
-    "\n",
-    "plt.show()"
```
dkrako, line 119: plt.show() is necessary

docs/source/user-guide/inspect-raw-samples.ipynb
```diff
@@ -251,7 +246,13 @@
    ]
   }
  ],
- "metadata": {},
+ "metadata": {
```
dkrako: didn't a precommit hook automatically clean that metadata?

src/pymovements/plotting/tsplot.py
cbueth: Sorry, meant `share_y`

docs/source/user-guide/inspect-raw-samples.ipynb
cbueth: On my notebooks it does show without, as we have `show: bool = True`.

Should this PR have a readthedocs CI?

docs/source/user-guide/inspect-raw-samples.ipynb
cbueth: see above

docs/source/user-guide/inspect-raw-samples.ipynb
cbueth: We should get both, the interactive plot shown and the `__repr__` of the figure and axes.
Plots neither show if one just has Figure or Axes, they only show with matplotlib logic, not because they are on the last line of a code cell with the html repr. If this is different for the Sphinx build, I'm happy to be wrong :)

Image

docs/source/user-guide/inspect-raw-samples.ipynb
cbueth: see above

docs/source/user-guide/inspect-raw-samples.ipynb
cbueth: It seems like no. I actively wrote this metadata being asked to select a kernel. This happens when using jupyterlab, but not with common IDEs.

I did not quickly find an option for this in nbQA. Probably, we just never wrote this information to notebooks.

cbueth: I manually removed this information in 4272f124.

docs/source/user-guide/inspect-raw-samples.ipynb
cbueth: Now I found the CI, plots show, https://pymovements--1523.org.readthedocs.build/en/1523/user-guide/inspect-raw-samples.html#inspecting-raw-samples-with-plots.

docs/source/user-guide/inspect-raw-samples.ipynb
cbueth: https://pymovements--1523.org.readthedocs.build/en/1523/user-guide/inspect-raw-samples.html#inspecting-raw-samples-with-plots

dkrako: but prior to that the matplotlib text was not displayed at all:

image

I definitely prefer that as there's no information relevant for the reader and thus distracts from the content

docs/source/user-guide/inspect-raw-samples.ipynb
dkrako: the `databooks` hook removes the metadata from notebooks, or at least it should

docs/source/user-guide/inspect-raw-samples.ipynb
cbueth: But then it is only not shown because the last line is `plt.show()`, but this is non functional. In other notebooks we als do not have it.

I think there are two perspectives:

- what do we want the user to see: I think by this perspective one might say, they want no attention to the fix and axes.
- what is functionally most instructive: not teaching the user wrongly that `plt.show()` does anything; but the code is hidden anyways. Then I'd prefer to make it not show the fig and axes a different way.

People who know what to do with the fig and axes now can know they can edit the figure.

My suggestion is, introducing a breaking change that will return `None`, not the fig and axes, if `show` is true (which it is by default).
What do you think, follow-up?

docs/source/user-guide/inspect-raw-samples.ipynb
cbueth: The hook was run locally and in CI. I am not sure why it did not do the expected.

dkrako: The databooks hook broke with py314 (ast has no Num attribute anymore) so maybe that might be the issue?

docs/source/user-guide/inspect-raw-samples.ipynb
dkrako: Thanks for your detailed response, it took me a while until I understood what the actual issue was. If I understand correctly the additional text box is printed because it's the output of the plotting function, so I see your point and agree on your proposal on the breaking change. 

However I would still prefer to keep the changes in that separate PR and revert them here as they are unrelated to this PR and have unwanted side effects on the presentation. Feel free to create a follow-up PR with your proposed changes.

cbueth: Ok, I understand!

docs/source/user-guide/inspect-raw-samples.ipynb
cbueth: Is the hook running on py314?

docs/source/user-guide/inspect-raw-samples.ipynb
dkrako: But if you like to work on the follow-up PR soon it's also fine to leave it as is as it won't matter much

docs/source/user-guide/inspect-raw-samples.ipynb
dkrako: probably fixed with #1533

docs/source/user-guide/inspect-raw-samples.ipynb
```diff
@@ -111,12 +107,12 @@
     "\n",
     "pm.plotting.tsplot(\n",
     "    gaze_unnested,\n",
+    "    xlabel='time [ms]',\n",
     "    channels=['pixel_x', 'pixel_y'],\n",
     "    share_y=False,\n",
     "    line_color='darkblue',\n",
```
dkrako, line 113:
```suggestion
    "    line_color='darkblue',\n",
    "    n_rows=2, n_cols=1,\n",
```

src/pymovements/plotting/tsplot.py
dkrako:
```suggestion
        # share_y=True will automatically hide those that are not on the bottom
```

PR #1539, opened 2026-04-01
chore!: remove `show` and `closefig` arguments from plotting functions

dkrako:
## Description

Following the discussion in #1523 this PR removed the `show` and `closefig` arguments from all plotting functions.

I originally started out with overloading the plotting function signatures such that `None` is returned if `show is True` but if an `ax` is provided the `show` is ignored. So that didn't work out completely and also overloading the signature created a very messy specification as there are so many arguments.

So I decided to remove the `show` and `closefig` arguments completely because a user can easily call these by themselves (`plt.close()`, `plt.show()`)  and other libraries that have plotting functionalities also do not provide these arguments. This makes the signature cleaner so I think this is the best for the long run.

## Implemented changes

- [x] remove `show` from all plotting functions
- [x] remove `closefig` from all plotting functions
- [x] drop returning `own` in `prepare_figure()` as it wasn't used anymore

## How Has This Been Tested?

- [x] adjusted existing tests
- [x] test that all plotting functions return a plt.Figure and a plt.Axes

## Checklist:


dkrako:
> One could argue that `test_traceplot_noshow` could be removed.

You're right, I was unsure about that one but I kept it to prevent regressions and as tests are a form of specification I think it's fine to keep it.

PR #1535, opened 2026-03-24
feat: add bids-conformant participants data structure

dkrako:
## Description

This adds a `Participants` class that conforms to the [BIDS specification](https://bids-specification.readthedocs.io/en/stable/modality-agnostic-files/data-summary-files.html#participants-file).

It includes a field `data` for the tabular data (saved in the `participants.tsv`)  and a `metadata` field that holds additional descriptions on `data` and its columns.

## Implemented changes

- [x] add `Participants` class
- [x] add `Participants.load()` method
- [x] add `Participants.save()` method
- [x] auto infer column formats
- [x] cast `data` columns according to the format in `metadata`

## How Has This Been Tested?

- [x] tested inferring of column formats
- [x] tested casting according to metadata
- [x] tested loading participants files
- [x] tested saving participants files

## Context

First step towards #609

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

PR #1543, opened 2026-04-02
feat: add BIDS participant conformity verification

### LINKED ISSUE #1540: feat: BIDS conformity verification for participants
- [ ] improve BIDS conformity verification
  - add `Participants.verify_bids(level: Literal['REQUIRED', 'RECOMMENDED'] = 'REQUIRED')`.  String values may be `REQUIRED` and `RECOMMENDED` to check different levels.
  - add `verify_bids: str | bool = 'REQUIRED'` argument to `Participants.save()` and warn if saved data does not conform to BIDS specification.
  - add `verify_bids: str | bool = False` to `Participants.__init__()` and `Participants.load()` methods to opt-in for veryfing data on initialization. avoid exceptions during initialization (apart from one requirement: must have `participant_id` column)

_Originally posted by @dkrako in https://github.com/pymovements/pymovements/issues/1535#issuecomment-4126758833_
            
Emit detailled warning at non conformity, but do not error.
            
[Bids Specification](https://bids-specification.readthedocs.io/en/stable/modality-agnostic-files/data-summary-files.html#participants-file)

cbueth:
## Description

Add BIDS conformity verification to Participants class. This feature adds validation of participants.tsv files according to the BIDS specification, supporting REQUIRED and RECOMMENDED validation levels with proper handling of missing values (n/a).

## Implemented changes
- Added `verify_bids` parameter to `__init__`, load, and save methods:
  - False - No verification
  - 'REQUIRED' - Warn for REQUIRED-level non-conformities
  - 'RECOMMENDED' - Warn for both REQUIRED and RECOMMENDED
  - True - Raise ValueError on any non-conformity
- Added `verify_bids(level)` method for manual validation
- Added validation helper functions:
  - `_validate_participant_id` - Validates format (`sub-<label>`), column order, uniqueness
  - `_validate_age` - Validates numeric values, caps at 89 per BIDS
  - `_validate_sex` - Validates allowed values (M/F/O and variants)
  - `_validate_handedness` - Validates allowed values (L/R/A and variants)
  - `_validate_species` - Placeholder for species validation
  - `_validate_strain` - Placeholder for strain validation
  - `_validate_strain_rrid` - Validates RRID format
- Handles n/a missing values per BIDS specification

## How Has This Been Tested?

- 109 unit tests covering all validation functions
- Tests pass locally

## Type of change

- New functionality

## Context

Resolves #1540

#### related issues:
- #1535

#### requires:
- [x] #1535

## Checklist:


src/pymovements/dataset/participants.py
dkrako, line 80: this check should be removed as it's now included in `_verify_participant_id()`

src/pymovements/dataset/participants.py
```diff
@@ -51,6 +54,11 @@ class Participants:
         Additional metadata on participant data conforming to BIDS side car json files.
         If ``None``, initialize an empty dictionary.
         (default: ``None``)
+    verify_bids: str | bool
+        Verify BIDS conformity. If True, raise exception on non-conformity.
```
dkrako, line 58: This should include which level is checked if `True`

src/pymovements/dataset/participants.py
```diff
@@ -83,15 +92,29 @@ def __init__(
         self.data = data
         self.metadata = metadata
 
+        if verify_bids is not False:
+            level: Literal['REQUIRED', 'RECOMMENDED'] = 'REQUIRED'
+            if isinstance(verify_bids, str):
+                level = verify_bids  # type: ignore[assignment]
```
dkrako, line 98: if you change the argument signature above to `verify_bids: Literal['REQUIRED', 'RECOMMENDED'] | bool = False` you should not need that ignore

src/pymovements/dataset/participants.py
```diff
@@ -218,6 +265,328 @@ def save(
         with open(metadata_path, 'w', encoding=metadata_encoding) as opened_file:
             json.dump(self.metadata, opened_file)
 
+    def verify_bids(
+        self,
+        level: Literal['REQUIRED', 'RECOMMENDED'] = 'REQUIRED',
+    ) -> list[str]:
```
dkrako, line 271: I like that idea with returning a list here instead of warning/raising inside of this function as this makes it more reusable and easier to test.

src/pymovements/dataset/participants.py
```diff
@@ -195,6 +229,19 @@ def save(
             Use this encoding for loading the metadata json file.
             (default: ``utf-8``)
         """
+        if verify_bids is not False:
+            level: Literal['REQUIRED', 'RECOMMENDED'] = 'REQUIRED'
+            if isinstance(verify_bids, str):
+                level = verify_bids  # type: ignore[assignment]
+            warnings_list = self.verify_bids(level)
```
dkrako, line 236: verification during saving should also verify that

- the filename is participants.tsv
- the file is tab separated
- string values including tab chars are escaped with "
- null values are encoded as `n/a`
- files are utf-8 encoded
- and if a metadata_path is provided it should be participants.json within the same directory.

src/pymovements/dataset/participants.py
```diff
@@ -162,12 +190,13 @@ def load(
         else:
             metadata_dict = metadata
 
-        return Participants(data, metadata_dict)
+        return Participants(data, metadata_dict, verify_bids=verify_bids)
```
dkrako, line 193: same as with `save()` this should also verify that:

- the filename is participants.tsv
- the file is tab separated
- string values including tab chars are escaped with `"`
- files are utf-8 encoded
- null values are encoded as n/a
- and if a metadata_path is provided it should be participants.json within the same directory.

src/pymovements/dataset/participants.py
```diff
@@ -218,6 +265,328 @@ def save(
...
+    """
+    validation_warnings: list[str] = []
+
+    if 'participant_id' not in data.columns:
+        return ['participant_id column is missing']
+
+    if data.columns[0] != 'participant_id':
+        validation_warnings.append('participant_id column must be the first column')
+
+    participant_ids = data['participant_id'].drop_nulls().to_list()
```
dkrako, line 375: wouldn't that line ignore `null` values as `participant_id`? please add a test to make sure that a `null` value is treated as non-conformity.

src/pymovements/dataset/participants.py
```diff
@@ -218,6 +265,328 @@ def save(
...
+
```
dkrako, line 376: also, please add a test to make sure the column dtype is `Series`

src/pymovements/dataset/participants.py
```diff
@@ -218,6 +265,328 @@ def save(
...
+    pattern = re.compile(r'^sub-[a-zA-Z0-9]+$')
```
dkrako, line 377: actually this is slightly too restrictive as `+` seems to be allowed for `label`: https://bids-specification.readthedocs.io/en/stable/glossary.html#label-formats

src/pymovements/dataset/participants.py
```diff
@@ -218,6 +265,328 @@ def save(
...
+            warnings_list.extend(_validate_participant_id(self.data))
+
+        if level == 'RECOMMENDED':
+            warnings_list.extend(_validate_age(self.data))
+            warnings_list.extend(_validate_sex(self.data))
+            warnings_list.extend(_validate_handedness(self.data))
+            warnings_list.extend(_validate_species(self.data))
+            warnings_list.extend(_validate_strain(self.data))
+            warnings_list.extend(_validate_strain_rrid(self.data))
+
```
dkrako, line 350: this should also include checking columns other than the recommended ones.

if there are additional columns, it is recommended that they are included in the `metadata` json with at least a `Description` field. (https://bids-specification.readthedocs.io/en/stable/modality-agnostic-files/data-summary-files.html#participants-file before the `participants.json` example)

moreover, according to the section on [tabular files](https://bids-specification.readthedocs.io/en/stable/common-principles.html#tabular-files)  it is recommended that all column names:
- are written in snake case
- are not blank/null

src/pymovements/dataset/participants.py
```diff
@@ -218,6 +265,328 @@ def save(
...
+    Returns
+    -------
+    list[str]
+        List of warning messages for any non-conformities found.
+    """
+    validation_warnings: list[str] = []
+
+    if 'age' not in data.columns:
+        return validation_warnings
+
```
dkrako, line 409: the column type must be either integer or float

src/pymovements/dataset/participants.py
```diff
@@ -218,6 +265,328 @@ def save(
...
+    ages = data['age'].drop_nulls().to_list()
+    na_values = {'n/a', 'N/A', 'NA', 'NaN', 'nan', ''}
+    for age in ages:
+        if str(age).lower() in na_values:
```
dkrako, line 413: >  Missing and non-applicable values MUST be coded as n/a.
https://bids-specification.readthedocs.io/en/stable/common-principles.html#tabular-files

this means that `level` has to be included here, as this is `REQUIRED` not `RECOMMENDED`.

src/pymovements/dataset/participants.py
```diff
@@ -218,6 +265,328 @@ def save(
...
+
+    Returns
+    -------
+    list[str]
+        List of warning messages for any non-conformities found.
+    """
+    validation_warnings: list[str] = []
+
+    if 'sex' not in data.columns:
+        return validation_warnings
```
dkrako, line 443: column dtype must be string

src/pymovements/dataset/participants.py
```diff
@@ -218,6 +265,328 @@ def save(
...
+
+    valid_male = {'male', 'm', 'M', 'MALE', 'Male'}
+    valid_female = {'female', 'f', 'F', 'FEMALE', 'Female'}
+    valid_other = {'other', 'o', 'O', 'OTHER', 'Other'}
+    valid_sex = valid_male | valid_female | valid_other
+    na_values = {'n/a', 'N/A', 'na', 'NA', 'NaN', 'nan', ''}
```
dkrako, line 449: > Missing and non-applicable values MUST be coded as n/a.
> https://bids-specification.readthedocs.io/en/stable/common-principles.html#tabular-files

src/pymovements/dataset/participants.py
```diff
@@ -218,6 +265,328 @@ def save(
...
+    Returns
+    -------
+    list[str]
+        List of warning messages for any non-conformities found.
+    """
+    validation_warnings: list[str] = []
+
+    if 'handedness' not in data.columns:
+        return validation_warnings
+
```
dkrako, line 483: column dtype must be string

src/pymovements/dataset/participants.py
```diff
@@ -218,6 +265,328 @@ def save(
...
+    Returns
+    -------
+    list[str]
+        List of warning messages for any non-conformities found.
+    """
+    validation_warnings: list[str] = []
+
+    if 'age' not in data.columns:
+        return validation_warnings
```
dkrako, line 408: this field is `RECOMMENDED` so it should generate a warning.

src/pymovements/dataset/participants.py
```diff
@@ -218,6 +265,328 @@ def save(
...
+        The participants DataFrame to validate.
+
+    Returns
+    -------
+    list[str]
+        List of warning messages for any non-conformities found.
+    """
+    validation_warnings: list[str] = []
+
+    if 'sex' not in data.columns:
```
dkrako, line 442: this field is `RECOMMENDED` so it should generate a warning.

src/pymovements/dataset/participants.py
```diff
@@ -218,6 +265,328 @@ def save(
...
+        The participants DataFrame to validate.
+
+    Returns
+    -------
+    list[str]
+        List of warning messages for any non-conformities found.
+    """
+    validation_warnings: list[str] = []
+
+    if 'handedness' not in data.columns:
```
dkrako, line 481: this field is `RECOMMENDED` so it should generate a warning.

src/pymovements/dataset/participants.py
```diff
@@ -218,6 +265,328 @@ def save(
...
+        The participants DataFrame to validate.
+
+    Returns
+    -------
+    list[str]
+        List of warning messages for any non-conformities found.
+    """
+    validation_warnings: list[str] = []
+
+    if 'species' not in data.columns:
```
dkrako, line 520: I'm unsure if we should warn here if it's missing.

src/pymovements/dataset/participants.py
```diff
@@ -218,6 +265,328 @@ def save(
...
+        return validation_warnings
+
```
dkrako, line 522: column dtype must be either number or string

src/pymovements/dataset/participants.py
dkrako: it is probably a good idea to refactor the na value check into a separate function and reuse it everywhere.

I think it's great to support all the `na_values` that you included here, but if it's not `n/a` this must fail the `REQUIRED` conformity check.

src/pymovements/dataset/participants.py
```diff
@@ -218,6 +265,328 @@ def save(
...
+    Returns
+    -------
+    list[str]
+        List of warning messages for any non-conformities found.
+    """
+    validation_warnings: list[str] = []
+
+    if 'strain' not in data.columns:
+        return validation_warnings
```
dkrako, line 546: this is only `RECOMMENDED` if the species is not `homo sapiens` so I think it's fine to not warn if it's missing

tests/unit/dataset/participants_test.py
```diff
@@ -19,13 +19,21 @@
...
 import json
+import warnings
 from copy import deepcopy
 
 import polars as pl
 import pytest
 from polars.testing import assert_frame_equal
 
 from pymovements import Participants
+from pymovements.dataset.participants import _validate_age
```
dkrako, line 30: please don't test private functions but rather test the output of the user facing method (`Participants.verify_bids()`). The output of that function is easily testable so changing that should be quick.

tests/unit/dataset/participants_test.py
```diff
@@ -395,46 +446,645 @@ def test_participants_save_data_to_dirpath(tmp_path):
...
+        )
 
         save_path = tmp_path / metadata_filename
         assert save_path.is_file()
         with open(save_path, encoding=encoding) as opened_file:
             saved_metadata = json.load(opened_file)
         assert saved_metadata == metadata
+
+
+class TestVerifyBids:
```
dkrako, line 525: I wonder why you use a test `class` here. as far as I understand this is mainly useful if you want to share the same parametrizations or other marks across test functions but it's just a single method here.

tests/unit/dataset/participants_test.py
```diff
@@ -395,46 +446,645 @@ def test_participants_save_data_to_dirpath(tmp_path):
...
+        ],
+    )
+    def test_verify_bids_init(
+        self,
+        data,
+        verify_bids,
+        expected_exception,
+        expected_message,
+    ):
+        if expected_exception:
```
dkrako, line 702: as you're already using a `class` here, please get rid of the `if/else` and rewrite it as separate methods and apply the parametrizatio decorator on the class instead

tests/unit/dataset/participants_test.py
```diff
@@ -395,46 +446,645 @@ def test_participants_save_data_to_dirpath(tmp_path):
...
+                _ = Participants(data, verify_bids=verify_bids)
+                if verify_bids not in (False, None):
+                    if expected_message:
+                        warning_messages = [str(warning.message) for warning in w]
+                        assert any(expected_message in msg for msg in warning_messages)
+                else:
+                    assert not w
+
+
+class TestVerifyBidsLoad:
```
dkrako, line 717: does the usage of `class` here and below serve any other purpose besides grouping?

tests/unit/dataset/participants_test.py
```diff
@@ -395,46 +446,645 @@ def test_participants_save_data_to_dirpath(tmp_path):
...
+                pl.DataFrame(
+                    {'participant_id': ['sub-01'], 'strain': [None]},
+                ),
+                [],
+                id='strain_all_null',
+            ),
+        ],
+    )
+    def test_validate_functions_edge_cases(self, data, expected):
+        if 'age' in data.columns:
```
dkrako, line 1026: please get rid of all if/else logic here

tests/unit/dataset/participants_test.py
```diff
@@ -395,46 +446,645 @@ def test_participants_save_data_to_dirpath(tmp_path):
...
+                id='no_sex_column',
+            ),
+        ],
+    )
+    def test_validate_sex(self, data, expected):
+        warnings_list = _validate_sex(data)
+        assert warnings_list == expected
+
+
+class TestValidateEdgeCases:
```
dkrako, line 897: I don't think that the functions included here are related enough to group them in this class

dkrako (changes requested):
Great, thank you so much for your work!

I have a number of comments regarding parts of the verifcation and the testing. The saving/loading must also verify that the data is saved in a correct format so it needs some additional logic that's not part of `verify_bids()`.

src/pymovements/dataset/participants.py
cbueth: I've removed the redundant check for `participant_id` in the `__init__` method.

src/pymovements/dataset/participants.py
cbueth: updated the docstring to clarify that `verify_bids=True` defaults to verifying at the `REQUIRED` level and raising a `ValueError` on any non-conformity.

src/pymovements/dataset/participants.py
cbueth: I've updated the type hints for `verify_bids` in `__init__`, `load`, and `save` to use `Literal['REQUIRED', 'RECOMMENDED'] | bool`.

src/pymovements/dataset/participants.py
cbueth: - Added warnings if the filename or `metadata_path` deviates from BIDS standards (`participants.tsv`/`.json`).
 - Ensured that `separator='\t'`, `encoding='utf-8'`, and `quoting=1` (for escaping) are used by default or verified when `verify_bids` is active.
 - Handled null values to ensure they are written as `n/a` in the output file.

src/pymovements/dataset/participants.py
cbueth: se above

src/pymovements/dataset/participants.py
cbueth: Yup. I've updated `_validate_participant_id` to explicitly check for null values in the `participant_id` column. I've also added a unit test `test_verify_bids_participant_id_null` to ensure this correctly triggers a warning.

src/pymovements/dataset/participants.py
cbueth: Added a check in `_validate_participant_id` to verify that the `participant_id` column has a string-compatible data type.

src/pymovements/dataset/participants.py
cbueth: Thank you for finding this! I've updated the regex pattern to `^sub-[a-zA-Z0-9]+$` which allows alphanumeric characters for the label.
Should be conforming to the BIDS specification now.

src/pymovements/dataset/participants.py
cbueth: I've added checks
- to `verify_bids` (at the `RECOMMENDED` level) to verify that all columns have a corresponding entry in `metadata` with a `Description` field. 
- for snake_case naming and ensured column names are not null/blank.

src/pymovements/dataset/participants.py
cbueth: Added check in `_validate_age` to ensure the column data type is numeric.

src/pymovements/dataset/participants.py
cbueth: Refactored the `n/a` check into a common helper function `_check_na_conformity`. This check is now performed at the `REQUIRED` level for all standard BIDS columns (`age`, `sex`, `handedness`, etc.), ensuring that any null or alternative `n/a` representations (like `NaN` or empty strings) trigger a warning if they aren't explicitly `n/a`.

src/pymovements/dataset/participants.py
cbueth: see above

src/pymovements/dataset/participants.py
cbueth: is now being validated

src/pymovements/dataset/participants.py
cbueth: is now being validated

src/pymovements/dataset/participants.py
cbueth: `verify_bids`, when `level='RECOMMENDED'` now explicitly warns if any of the recommended columns (`age`, `sex`, `handedness`) are missing from the dataset.

src/pymovements/dataset/participants.py
cbueth: see above

src/pymovements/dataset/participants.py
cbueth: Doing this now
https://github.com/pymovements/pymovements/blob/efa4cf65cdcd8bfd77dd92b2aa0510f2fd367157/src/pymovements/dataset/participants.py#L709-L710

src/pymovements/dataset/participants.py
cbueth: Yes:
https://github.com/pymovements/pymovements/blob/efa4cf65cdcd8bfd77dd92b2aa0510f2fd367157/src/pymovements/dataset/participants.py#L712

tests/unit/dataset/participants_test.py
cbueth: I agree that testing the user facing method is most important. Testing the private functions is okay additionally, or should we not do that generally?

I've moved the logic from testing `_validate_*` functions to testing the public `Participants.verify_bids()` method directly. This makes the tests more robust to internal implementation changes.

tests/unit/dataset/participants_test.py
cbueth: It does not have any added value here, but can be a mental grouping.

tests/unit/dataset/participants_test.py
cbueth: I removed classes and refactored the tests for top-level test functions, eliminating the `if/else` branching within tests.

tests/unit/dataset/participants_test.py
cbueth: That is true. I've flattened the test structure and grouped tests by BIDS field or functionality.

cbueth:
Thank you for the detailled review. I integrated it all and it is now ready for your secound round.

src/pymovements/dataset/participants.py
```diff
@@ -240,21 +311,471 @@ def save(
...
+    validation_warnings: list[str] = []
+
+    if 'participant_id' not in data.columns:
+        return ['participant_id column is missing']
+
+    if data.columns[0] != 'participant_id':
+        validation_warnings.append('participant_id column must be the first column')
+
+    # Check dtype
+    if not data['participant_id'].dtype.is_temporal(
```
dkrako: I'm a bit confused about that check here, as it doesn't do anything (the body consists just of a `pass`). The string dtype is correctly checked below.

I guess it's safe to remove this codeblock?

src/pymovements/dataset/participants.py
dkrako: thanks, I typed series instead of str, good that you got it right

src/pymovements/dataset/participants.py
```diff
@@ -240,21 +311,471 @@ def save(
...
+        List of warning messages for any non-conformities found.
+    """
+    validation_warnings: list[str] = []
+
+    if 'age' not in data.columns:
+        return validation_warnings
+
+    # Check dtype
+    if not data['age'].dtype.is_numeric():
+        validation_warnings.append("Column 'age' must be of numeric type (integer or float)")
```
dkrako:
```suggestion
        validation_warnings.append(
            f"Column 'age' must be of numeric type (integer or float), got '{data['age'].dtype}'",
        )
```

src/pymovements/dataset/participants.py
```diff
@@ -240,21 +311,471 @@ def save(
...
+        >>> data = pl.DataFrame({"participant_id": ["01"]})
+        >>> with warn.catch_warnings(record=True) as w:
+        ...     warn.simplefilter("always")
+        ...     participants = Participants(data, verify_bids="REQUIRED")
+        ...     print(str(w[0].message))
+        participant_id values must match 'sub-<label>' pattern. Invalid values: ['01']
+        """
+        warnings_list: list[str] = []
+
+        if level in {'REQUIRED', 'RECOMMENDED'}:
```
dkrako, line 433: you can remove the if here as this will always be true.

cbueth:
There is still much coverage missing, I need to improve this still.

cbueth:
> There is still much coverage missing, I need to improve this still.

After 80ffc8e, patch coverage is now at 100%, see https://app.codecov.io/gh/pymovements/pymovements/pull/1543. Adding the tests was more of a chore, so feel free to suggest any reformatting.

tests/unit/dataset/participants_test.py
dkrako: > I agree that testing the user facing method is most important. Testing the private functions is okay additionally, or should we not do that generally?

In general I try to avoid testing private functions because it can make refactoring work tedious. I'm guilty of testing private functions myself and there are lots of old tests that do that (and I learned it the hard way, .e.g. during refactoring the download logic into `WebSource`). I tend to think if a functionality can only be tested well by testing private internals, then there's probably a design issue at hand.

In this case however, `Participants.verify_bids()` was easily testable. Thank you so much for updating your tests!

src/pymovements/dataset/participants.py
dkrako: I added a `ValueError` if the level is unsupported

src/pymovements/dataset/participants.py
dkrako: I removed this codeblock, hope it's fine.

dkrako (approved):
@cbueth Thank you so much for taking care of all my comments! I added a `ValueError` and removed a noop code block. This is now ready to be merged.

cbueth:
Thanks for merging!

PR #1544, opened 2026-04-06
docs: update GGTG reference

saeub:
from preprint to published version


dkrako (approved):
thanks for the update!

PR #1558, opened 2026-04-10
chore: add logo file with FontAwesome compatibility

cbueth:
This PR adds a simplified pymovements logo compatible with FontAwesome. The original `logo.svg` uses stroked wave lines, a custom viewBox, and paths that extend beyond the icon bounds, which works well for standard SVGs but is not suitable for FontAwesome custom icons. For the simplified version, those stroke-based details, colors, and out-of-bounds path shapes cannot be used, so the logo is rebuilt as a much simpler fill-only shape made of loops.

See here in use:

image


dkrako:
Interesting, did not `FontAwesome` yet. How do you see that being used in our documentation?

cbueth:
It is also my first time creating such icon. I have used it here in another repository, as an icon linking to `pymovements`: https://multipleye-cost.github.io/multipleye-preprocessing/
It could be used similarly in our docs, but I do not have a specific usecase I would suggest. But having it could beuseful to reference it from other projects to here.

dkrako:
Thanks for the infos!

dkrako (approved):

PR #1562, opened 2026-04-15
feat: add new fields `size` and `resolution` to `Screen`

### LINKED ISSUE #1490: add fields `resolution` and `size` to `Screen`
## Description of the problem

The current way of keeping each component of screen size and resolution as a separate field could be improved.

## Description of a solution

Instead of having `height_px`, `width_px` and `height_cm`, `width_cm` it would be much nicer to have just `size` and `resolution` field.

For backwards compatiblity, fields like `height_px` could be implemented as a python property that access `Screen.size` or `Screen.resolution`.

dkrako:
## Description

This PR adds the fields `size` and `resolution` to `Screen`. Initial work was done by @li-kloostra

## Implemented changes

- [x] add properties for `width_cm`, `height_cm`, `width_px` and `height_px` 
- [x] add property `resolution`
- [x] add property `size`
- [x] add argument `exclude_private` to `Screen.to_dict()`
- [x] add arguments `prefer_resolution` and `prefer_size` to dict
- [x] add `__str__()` method to exclude private attributes and include properties

## How Has This Been Tested?

- [x] tested initialization, including exceptions
- [x] tested setting properties, including exceptions
- [x] tested new to_dict arguments

## Context

Resolves #1490 

## Checklist:


src/pymovements/gaze/screen.py
```diff
@@ -285,7 +436,14 @@ def _check_numerical_attribute(self, key: str) -> None:
         assert isinstance(value, (int, float))
         _checks.check_is_greater_than_zero(**{key: value})
 
-    def to_dict(self, *, exclude_none: bool = True) -> dict[str, Any]:
+    def to_dict(
+            self,
+            *,
+            exclude_none: bool = True,
+            exclude_private: bool = True,
```
dkrako: I don't know if this additional argument is really worth adding. I don't see any use case where someone would like to include these underscore private arguments.

dkrako: declaring the private attributes via `InitVar[]` might solve this

src/pymovements/gaze/screen.py
```diff
@@ -294,22 +452,69 @@ def to_dict(self, *, exclude_none: bool = True) -> dict[str, Any]:
...
+            dictionary.
+            (default: False)
 
         Returns
         -------
         dict[str, Any]
             Screen as dictionary.
         """
-        _dict = asdict(self)
+        data = asdict(self)  # this does not include properties, just explicit attributes.
```
cbueth:
```suggestion
        data = {field.name: getattr(self, field.name) for field in fields(self)}
```

Fyi: the same in different words.

src/pymovements/gaze/screen.py
cbueth: In my opinion, we can remove this.

li-kloostra: just FYI, i asked Gemini and ChatGPT to help me and give suggestions; there may be some stuff that is not necessary in the code

dkrako: initvar didn't solve this, as the attributes aren't registered this way. I chose using `fields()` over `asdict()`

src/pymovements/gaze/screen.py
dkrako: To be fully clear: this is only the same if no fields are a dataclass (shallow copy). `asdict()` applies itself recursively in that case.

However, as all our fields are basic datatypes this is irrelevant, so I extended the dictionary expression to exclude the private attributes.

I also put this after sourcing the resolution/size attributes to make sure the order in the to_dict() output is not changed in this PR.

[`134a07e`](https://github.com/pymovements/pymovements/pull/1562/commits/134a07e8e08eea5be46ad55b9b9f702edb9ffeb2)

src/pymovements/gaze/screen.py
dkrako: > just FYI, i asked Gemini and ChatGPT to help me and give suggestions; there may be some stuff that is not necessary in the code

ah don't worry that part of the code was added by myself but proved to be unnecessary

PR #1566, opened 2026-04-19
dataset: update GGTG dataset

saeub:
Had to update one of the files, here's the updated hash.


dkrako (approved):
Thanks for the update!

Issue #37, opened 2022-10-14
Integrate Identification by Hidden Markov Model (I-HMM) eye movenent detection algorithm


dkrako:
## Description of the problem
The algorithm is defined here:
https://dl.acm.org/doi/10.1145/355017.355028
There exist several implementations of this on the internet where you can get inspiration.

## Description of a solution
- [ ] implement basic algorithm
- [ ] create synthetic test data and test for expected result

## Minimum acceptance criteria
- [ ] passing tests with synthetic test data

PR #1570, opened 2026-04-20
Feat/i hmm issues with minimum duration units

Di0nigi:
## Description

I added a few robustness features to ensure the correctness of predictions, including a minimum duration threshold for fixations, as suggested in the Salvucci paper. The only thing I'm unsure about is how to handle potential differences in the unit of measure between the threshold and the timesteps. In the IDT algorithm, from what I gathered, this is handled by dividing by the timestep scale and then using indices instead of timesteps to check the length of a state. I'm using actual durations, so I can't directly do it this way. Is there a best practice I should follow? Or should I just specify it in the documentation and assume the user uses the same unit of measure, even when creating the timesteps?

Regarding testing I provided in the folder 'testsIHMM', as we discussed, three files one is a pytest file for unit testing of the functions (testing.py), one is a notebook to test against IDT for event detection correctness (testingCorrectness.ipynb) and the last one is a work in progress notebook to show and test the different algotihm's functionalities and parameters, also comparing it to idt. DISCLAIMER: I used AI to create the unit testing and the first notebook and then reviewed and corrected the code myself since I didn't really have much experience with proper testing procedures. Disclaimer: I used AI to create the unit tests and the first notebook, and then reviewed and corrected the code myself, since I don't have much experience with proper testing procedures.

Lastly, since the implementation seems correct and mostly done, what should I focus on next? Should I focus on trying to optimize it? If yes, are there suggested procedures or parts of the code that need it more than others?

Now I plan to focus on:

* Polishing the docstrings and documentation

* Further testing and error handling (if necessary)

* Adding comments where missing

Does it look good for now? Is there anything else I should prioritize?

## Type of change

Remove irrelevant items:

- New functionality

## Context

Resolves # https://github.com/pymovements/pymovements/issues/37


dkrako:
Hi @Di0nigi sorry for my late reply as I was on vacation.

> I currently compute velocities inside the function because it is an explicit step of the algorithm. However, I’m wondering if it would be better to assume velocities as input and let the user apply the pos2vel function beforehand, in order to be more consistent with other algorithms that use velocity.

Yes I think I would prefer that and it would also be consistent with the other velocity based algorithms I-VT and microsaccades.

> Secondly, after reestimation is applied, I think it would be useful to return the optimal parameters found by the algorithm to the user. However, I’m not sure what the best way to do this is without interfering with the function’s primary return value.

For now I guess the easiest way is to generate some arrays where you know that fixations should or should not be detected. In your doctest in the example you already use that approach so you can just write more paremetrizations for that to check a few more cases.

> I still plan to clean up the code further, but for now, does everything look more or less okay? I tried to follow the coding guidelines from the contributing guide, but I would still like to confirm that the overall code structure is sound.

The code structure is fine but at some playces it might have been worth trying to reuse existing functionality. For example there's already a `norm()` and `consecutive` function in `transforms_numpy.py`.

> I removed the HMM class and refactored the implementation into standalone functions. I also added a parameter that allows printing the optimized HMM parameters when needed. Additionally, I replaced the velocity computation with the pos2vel function; however, since it returns a 2D output, I converted it into a 1D representation by computing the L2 norm, which yields the resulting velocity vector.

Great, I think that's better for now.

> I used unit tests to verify the mathematical correctness of the implementation, but I am still uncertain how to properly evaluate the overall correctness of the results. In particular, I am unsure how to validate whether the inferred states are accurate.

Can you include the tests here? The doctest in the example docstring is already a good start for that.

> Another open issue is handling NaN values in the velocity array. At the moment, I simply replace them with zero as a placeholder.

Replacing NaNs with zero here might lead to false positives so maybe we should avoid that. Per default these values should probably be ignored.

PR #1577, opened 2026-05-09
fix: reduce overlapping AOI matches to single match in mapping methods

### LINKED ISSUE #1574: `map_to_aois()` with overlapping AOIs results in misalignment
If there fixations that are inside multiple (overlapping) AOIs and `map_to_aois()` is applied, the result is misaligned:

```py
import pymovements as pm
import polars as pl

events = pm.Events(
    pl.DataFrame({
        "name": ["fixation", "fixation", "fixation"],
        "location": [(50, 50), (150, 150), (250, 250)],
        "onset": [0, 1000, 2000],
        "offset": [500, 1500, 2500],
    })
)

stimulus = pm.stimulus.TextStimulus(
    pl.DataFrame({
        "content": ["AOI1", "AOI2"],
        "left": [0, 100],
        "right": [200, 200],
        "top": [0, 100],
        "bottom": [200, 200]
    }),
    aoi_column="content",
    start_x_column="left",
    end_x_column="right",
    start_y_column="top",
    end_y_column="bottom"
)

events.map_to_aois(stimulus)
events.frame
```

**Result:**

| name     | onset | offset | duration | location_x | location_y | content | left | right | top  | bottom |
| -------- | ----- | ------ | -------- | ---------- | ---------- | ------- | ---- | ----- | ---- | ------ |
| fixation | 0     | 500    | 500      | 50         | 50         | AOI1    | 0    | 200   | 0    | 200    |
| fixation | 1000  | 1500   | 500      | 150        | 150        | AOI1    | 0    | 200   | 0    | 200    |
| fixation | 2000  | 2500   | 500      | 250        | 250        | AOI2    | 100  | 200   | 100  | 200    |
| null     | null  | null   | null     | null       | null       | null    | null | null  | null | null   |

Note how the third fixation is mapped to AOI2, even though it's outside both AOIs. It seems that the second fixation was mapped to both AOI1 and AOI2, and so both AOI rows were added to the data frame, leading to the misalignment.

**Expected result:**

| name     | onset | offset | duration | location_x | location_y | content | left | right | top  | bottom |
| -------- | ----- | ------ | -------- | ---------- | ---------- | ------- | ---- | ----- | ---- | ------ |
| fixation | 0     | 500    | 500      | 50         | 50         | AOI1    | 0    | 200   | 0    | 200    |
| fixation | 1000  | 1500   | 500      | 150        | 150        | AOI1    | 0    | 200   | 0    | 200    |
| fixation | 2000  | 2500   | 500      | 250        | 250        | null    | null | null  | null | null   |

According to 5c896da762ee413287fb84cfe4860933e5666b27, in case of overlapping AOIs, only the first matching AOI should be mapped and a warning should be output. None of these things are actually happening in the example above. Not sure what is happening here.

carlosjperez:
## Summary

- Fix `Events.map_to_aois()` so overlapping AOI matches keep one AOI row per event row.
- Preserve existing `TextStimulus.get_aoi()` behavior: overlapping AOIs still warn and the first match is used by event mapping.
- Add a regression test covering the misalignment example from #1574.
- Add myself to `CITATION.cff` as suggested by the first-time contributor message.

## Tests

- `uv run --extra dev pytest tests/unit/events/map_to_aois_behaviour_test.py::test_map_to_aois_uses_first_overlapping_aoi_without_misalignment -q`
- `uv run --extra dev pytest tests/unit/events/map_to_aois_behaviour_test.py -q`
- `uv run --extra dev pytest tests/unit/events -q`
- `uv run --extra dev pytest tests/unit/events/map_to_aois_behaviour_test.py --cov=pymovements --cov-report=term-missing -q`
- `uv run --extra dev pylint --rcfile=pylintrc --output-format=parseable src/pymovements/events/events.py`
- `uv run --extra dev pylint --rcfile=pylintrc --output-format=parseable --disable=missing-class-docstring,missing-function-docstring,protected-access tests/unit/events/map_to_aois_behaviour_test.py`
- `uv run --extra dev pre-commit run --files CITATION.cff src/pymovements/events/events.py tests/unit/events/map_to_aois_behaviour_test.py`

Closes #1574.


dkrako (approved):
@carlosjperez Thank you so much for taking care of this bug!

After adding an equivalent `Gaze.map_to_aois()` test to `gaze_aoi_mapping_test.py` I noticed that the same problem also persists there.

I therefore decided to move your changes in logic from `Event.map_to_aois()` to `TextStimulus.get_aoi()` by adding a new argument `max_matches`. This way it was simpler to adjust the more complex logic in `Gaze.map_to_aois()`.

This PR is now ready to be merged. Thanks again for taking care!

PR #1154, opened 2025-05-14
feat: add warning and improve error message if no gaze components

osmaneTKT:
…azeDataFrame
Resolves #1016

## Description

Please include a clear and concise description of the change and which issue is fixed.
Please also include excepted improvements, relevant motivation and context.
List any dependencies that are required for this change.

## Implemented changes

Insert a description of the changes implemented in the pull request.

- [ ] Change 1
- [ ] Change 2
- [ ] Change 3

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes.
Provide instructions so we can reproduce.
Please also list any relevant details for your test configuration

- [ ] Test A
- [ ] Test B

## Type of change

Remove irrelevant items:
- Bug fix
- New functionality
- Performance improvement
- New dataset
- Breaking change
- Documentation update

## Context

Resolves #

#### related issues:
- #

#### required by:
- #

#### requires:
- [ ] #

#### conflicts with:
- #

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] I have checked my code and corrected any misspellings

PR #1567, opened 2026-04-19
docs: fix some more cross-references

saeub:
## Description
fixes cross references in the documentation

PR #1579, opened 2026-05-14
feat: add validation and data quality reports to `Dataset` and `Gaze`

### LINKED ISSUE #1386: sanity checks for missing trial columns in gaze samples
## Description of the problem

As @SiQube noted in #1377 misconfigurations on user side (in particular: missing info on trial columns) may lead to catastrophic errors, as events and samples couldn't be auto-grouped by trials anymore.

Missing trial columns are hard to detect, as it varies by dataset if trial columns are present in a gaze sample/event file or not.

## Description of a solution

After some consideration, there might be a simple way of detecting missing trial columns which has a basic assumption:

*Each row in the `samples` dataframe must have a unique value-tuple of timestamp and specified trial columns.* *

Examples:

- In case of a single trial column `trial_id`, the value-tuple of the `samples` columns `time` and `trial_id` must be unique
- In case of a two trial columns `task` and `trial_id`, the `samples` columns `time`, `task` and `trial_id` must be unique
- In case of no trial columns, the `time` column must be unique

This could be checked either in `Gaze.__init__()` (preferably via a `Gaze.samples` setter) or at processing steps where trial columns would be of relevance (more flexible and permissive approach although more scattered and checks are repeated).

\* Unfortunately this isn't the case if eyelink asc files are created with a sampling rate of 2000 Hz and no floating point option, which leads to duplicated timestamps. see #688

- [ ] warn if there exist non-unique `(time, *trial_column_values)` in the `samples` dataframe
- [ ] make warning optional (opt-out, warn by default)

Insalatone:
## Description
This PR implements a dataset validation layer for pymovements, addressing the three classes of silent misconfiguration errors described in issue #1386 (Sanity Checks). It introduces Dataset.validate() and a lightweight GazeDataFrame-level helper that cross-check the dataset configuration against the actual schema of the loaded data, before errors can propagate silently through the preprocessing pipeline.

The core problem: fields such as trial_columns, time_column, and pixel_columns are currently accepted at construction time without being verified against the Polars frame schema. A column-name casing mismatch (e.g. trial_id vs trialId), an incorrect dtype (e.g. Float64 trial identifiers from CSV parsing), or an undeclared gaze component all produce either silent data corruption or cryptic downstream errors deep inside transformation logic. This PR moves those checks to load time and surfaces them as structured, human-readable reports.

## Implemented changes
- [x] Added `data_quality.py` inside `src/pymovements/dataset/` with `CheckResult`, `DataQualityReport`, `GazeDataValidationError`, and all seven check functions.
- [x] Added `Dataset.report_data_quality()` public method with `output_path`, `checks`, `measures`, `levels`, and `raise_on_error` keyword arguments.
- [x] Exported `CheckResult`, `DataQualityReport`, and `GazeDataValidationError` via the existing `pm.dataset` namespace.
- [x] Added tutorial notebook under `docs/source/tutorials/` demonstrating the three motivating failure modes and the corrective actions.

## How Has This Been Tested?
Tests were run using synthetic Polars DataFrames and a minimal mock Dataset. To reproduce, run `pytest tests/unit/dataset/data_quality_test.py` from the repository root.

- [x] Unit tests covering pass, warn, and fail branches for each of the seven checks using synthetic Polars DataFrames (40+ parametrised cases).
- [x] Integration tests for the full `report_data_quality()` path with multiple gaze frames, `raise_on_error`, output path, and selective checks.

## Type of change
- New functionality
- Documentation update

## Context
Resolves #1386

#### related issues:
- #1386 (Sanity Checks) direct implementation response
- #1154 (add warning and improve error message if no gaze components)  that PR improves one error message at call time; this PR moves the equivalent check to load time, generalises it to the full set of column dependencies, and wraps it in a structured ValidationReport

#### required by:
- None

#### requires:
- [ ] None

#### conflicts with:
- None

## Checklist:


src/pymovements/dataset/dataset.py
```diff
@@ -1240,6 +1245,130 @@ def save_preprocessed(
...
+        Examples
+        --------
+        >>> import pymovements as pm
+        >>> # dataset = pm.Dataset('ExampleDataset', path='data/')
+        >>> # dataset.load()
+        >>> # report = dataset.report_data_quality()
+        >>> # print(report.summary())
+        """
+        # pylint: disable=import-outside-toplevel
+        from pymovements.dataset.data_quality import _ALL_CHECKS
```
dkrako: please import these at the top level not inside the function

src/pymovements/dataset/dataset.py
```diff
@@ -1240,6 +1245,130 @@ def save_preprocessed(
...
+        checks_to_run = checks if checks is not None else list(_ALL_CHECKS.keys())
+        levels_to_run = (
+            levels if levels is not None else ['dataset', 'subject', 'session', 'trial']
+        )
+
+        # Extract source paths from fileinfo when available.
+        source_paths: list[str] = []
+        if isinstance(self.fileinfo, dict) and 'gaze' in self.fileinfo:
+            try:
+                fi_dicts = self.fileinfo['gaze'].to_dicts()
```
dkrako: is there a specific reason why you don't just write `self.fileinfo['gaze']['filepath']`?

src/pymovements/dataset/dataset.py
```diff
@@ -1240,6 +1245,130 @@ def save_preprocessed(
...
+                source_paths = [str(row.get('filepath', i)) for i, row in enumerate(fi_dicts)]
+            except Exception:  # noqa: BLE001
+                source_paths = [str(i) for i in range(len(self.gaze))]
+        else:
+            source_paths = [str(i) for i in range(len(self.gaze))]
```
dkrako: how do you discern between paths to actual files and these indexes later on? it seems that the check functions would then have an `affected_files` value that doesn't point to an actual file

src/pymovements/dataset/data_quality.py
```diff
@@ -0,0 +1,982 @@
...
+        check_id='gaze_range',
+        severity='pass',
+        message=(
+            f"{ratio * 100:.1f}% of samples lie within screen bounds. "
+            'Threshold: 95%.'
+        ),
+    )
+
+
+_ALL_CHECKS: dict[str, Callable[[Gaze, str], CheckResult]] = {
```
dkrako: all of these functions validate a single gaze object and have no dependency on `Dataset`. Please move all the validation functions to a new file (e.g.`gaze/validation.py`) and create a method `Gaze.validate()` that calls these checks.

I would probably prefer to have a boolean argument for each check instead of providing a list of checks but we can discuss that further.

src/pymovements/dataset/data_quality.py
```diff
@@ -0,0 +1,982 @@
...
+    check_results : list[CheckResult]
+        Results of all validation checks that were run.
+    measures : dict[str, pl.DataFrame]
+        Quality measures keyed by aggregation level:
+        ``'dataset'``, ``'subject'``, ``'session'``, ``'trial'``.
+    """
+
+    check_results: list[CheckResult] = field(default_factory=list)
+    measures: dict[str, pl.DataFrame] = field(default_factory=dict)
+    _warning_log: list[str] = field(default_factory=list, repr=False)
```
dkrako: this should be a public field

docs/source/tutorials/data-quality-report.ipynb
```diff
@@ -0,0 +1,503 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "a0",
+   "metadata": {},
+   "source": [
```
dkrako, line 7: as that issue will be closed on merging this PR referencing that is of no use in a tutorial so you can just remove that

docs/source/tutorials/data-quality-report.ipynb
```diff
@@ -0,0 +1,503 @@
...
+    "\n",
+    "    experiment = Experiment(\n",
+    "        screen_width_px=1280,\n",
+    "        screen_height_px=1024,\n",
+    "        screen_width_cm=38.0,\n",
+    "        screen_height_cm=30.0,\n",
+    "        distance_cm=68.0,\n",
+    "        origin='center',\n",
+    "        sampling_rate=sampling_rate,\n",
+    "    )\n",
```
dkrako, line 88: this code should be simplified for the tutorial. Using the proposed `Gaze.validate()` method would avoid all this code

src/pymovements/dataset/data_quality.py
```diff
@@ -0,0 +1,982 @@
...
+    """
+    all_measures = {'data_loss', 'std_rms', 'rms_s2s', 'bcea'}
+    requested = set(measures) if measures is not None else all_measures
+
+    # Extract fileinfo rows if available
+    fileinfo_rows: list[dict[str, Any]] = []
+    if isinstance(fileinfo, dict) and 'gaze' in fileinfo:
+        try:
+            fileinfo_rows = fileinfo['gaze'].to_dicts()
+        except Exception:  # noqa: BLE001  # pylint: disable=broad-exception-caught
```
dkrako: please avoid adding noqa or pylint ignores whenever possible

src/pymovements/dataset/data_quality.py
```diff
@@ -0,0 +1,982 @@
...
+            fileinfo_rows = []
+
+    results: dict[str, pl.DataFrame] = {}
+
+    # Build per-gaze-frame measure rows
+    rows_all: list[dict[str, Any]] = []
+
+    for i, gaze in enumerate(gaze_list):
```
dkrako: I would very much prefer that the content of this for loop is refactored into a `Gaze.report()` method or similar. Otherwise this functionality can not be used on a single `Gaze`  object and always needs an initialized `Dataset`.

tests/unit/gaze/gaze_init_test.py
```diff
@@ -1758,6 +1743,20 @@ def test_gaze_init_exceptions(init_kwargs, exception, exception_msg):
     msg, = excinfo.value.args
     assert msg == exception_msg
 
+def test_gaze_init_trial_columns_not_in_samples_warns() -> None:
```
dkrako: I'm a bit confused about this test change. Currently it does not relate to any changes in `Gaze.__init__()`  so it seems it is planned. Our current rudimentary validation during initialization could reuse the proposed `Gaze.validate()` method.

However in this case at hand I'm not sure if it makes sense to replace the exception here with a warning as I would consider that an error severity and it will most probably lead to exceptions in later preprocessing steps.

pymovements
dkrako, line 1: this file should be removed

dkrako (changes requested):
@Insalatone @IngaleOmkar Thank you so much for this extensive pull request! I especially like that you added functionality to save the reports as a derivative bids dataset. Also really great that you already added lots of tests and a tutorial to showcase your changes as this made it much easier to review.

However, I think there's still some room for improvement so expect to have some more review iterations. Generally I recommend making smaller self-contained PRs as this usually leads to a shorter turnaround time, but splitting the PR is not strictly necessary.

I added a few comments regarding some of your changes. In general I would very much prefer to move most of the logic from the `dataset` module into the `gaze` module. The logic inside the `dataset` module should be limited mostly to looping over dataset files/objects and aggregating results. Having methods lke `validate()` and `report()` available for `Gaze` will be very useful later on, especially for exploratory analyses where a `Dataset` is not initialized.

Also, please avoid adding any `noqa` or `pylint: disable` comments to your code as far as possible. We can discuss single instances where it's not really feasible but the ones that I saw are easy to take care of.

There also seems to be an issue with uploading the coverage report. It might be because a test is failing but usually this should still lead to a coverage report. We usually require 100% patch test coverage to merge a PR so it's important to write tests that cover each edge case.

IngaleOmkar:
@dkrako - we have updated the code. 3 of the tests are failing because we are getting a deprecation warning in Polars which is triggered when `.list.explode()` is called. The fix is adding `empty_as_null=True` to the 3 calls present in `smooth.py` and `savitzky_golay.py`. I wanted your confirmation before making this change.

Also, any advice on readthedocs failure would be appreciated. We've tried resolving it multiple times but it keeps popping up. Thanks!

cbueth:
Hoi @IngaleOmkar. Thanks for raising these two issues. About the failing tests, this was an upstream change and is not relating to your PR, so I added the fix in #1596. If you want to test this change, you can merge `fix/polars-deprecation-explode` into your feature branch, or wait until it is merged into main.

The documentation error also seems to be global. Did you build the documentation locally to validate your changes render correctly? edit: it is fixed in #1597.

As your review responses were from May 19, to make the process easier, would you please summarize what you changed after the last review. Then, @dkrako will come to your PR.

Insalatone:
Hoi! @cbueth we didn't build the documentation locally in order to validate our changes and if you prefer we can rebase or merge these changes directly with the current PR branch. Let us know how you'd like to proceed. 

In addition these are all the changes that we did in order to address @dkrako feedback received: 

I) Move logic to gaze module: validation.py, quality.py, Gaze.validate(), Gaze.report_data_quality() all implemented at Gaze level. Dataset.report_data_quality() now just loops and delegates.

II) Remove noqa/pylint comments: Zero such comments in any new file. Code structured to pass linters naturally through proper type hints and docstrings.

III) 100% patch coverage: 277+ tests covering all 7 checks × (pass, warning, error), all edge cases (null samples, empty frames, missing columns, error recovery paths), and full measure computation logic.

IV) Dataset only loops/aggregates: ~130-line Dataset.report_data_quality() method that loops over self.gaze, calls individual gaze.validate(), and delegates aggregation to shared _compute_measures() function.

V) Exploratory use cases enabled: Users can now call gaze.validate() and gaze.report_data_quality() directly without requiring a Dataset object, enabling ad-hoc data quality checks in notebooks and interactive analyses.

docs/source/conf.py
```diff
@@ -195,7 +195,12 @@ def setup(app):
     ),
 
     # Fully-qualified references to our classes that aren't resolvable via intersphinx inventory
-    (r'py:class', r'^pymovements\.dataset\.(?:Dataset|DatasetDefinition|DatasetPaths)$'),
+    (
+        r'py:class',
+        r'^pymovements\.dataset\.(?:Dataset|DatasetDefinition|DatasetPaths'
+        r'|CheckResult|DataQualityReport)$',
+    ),
+    (r'py:exc', r'^(?:pymovements\.dataset\.)?GazeDataValidationError$'),
```
cbueth: While you mentioned that you did not build the documentation pages, but you make this addition to the documentation build config. Please build the docs locally and verify if these really need to be added to the `nitpick_ignore_regex`.

tests/unit/dataset/data_quality_test.py
```diff
@@ -0,0 +1,1164 @@
+# Copyright (c) 2022-2026 The pymovements Project Authors
```
dkrako: As this is a new file only 2026 should be used.

```suggestion
# Copyright (c) 2026 The pymovements Project Authors
```

src/pymovements/gaze/quality.py
```diff
@@ -0,0 +1,624 @@
+# Copyright (c) 2022-2026 The pymovements Project Authors
```
dkrako:
```suggestion
# Copyright (c) 2026 The pymovements Project Authors
```

src/pymovements/gaze/validation.py
```diff
@@ -0,0 +1,630 @@
+# Copyright (c) 2022-2026 The pymovements Project Authors
```
dkrako:
```suggestion
# Copyright (c) 2026 The pymovements Project Authors
```

tests/unit/gaze/validation_test.py
```diff
@@ -0,0 +1,729 @@
+# Copyright (c) 2022-2026 The pymovements Project Authors
```
dkrako:
```suggestion
# Copyright (c) 2026 The pymovements Project Authors
```

docs/source/conf.py
dkrako: I agree, let's undo these changes and see where we get the warnings. I guess it will be resolved by replacing things like ```:py:class:`~pymovements.dataset.CheckResult` ``` with ```:py:class:`~pymovements.CheckResult` ```

Moreover I think most of the ignores in this file are bandaids introduced by #1344 and should be resolved by using the correct (minimal) reference associated by sphinx. #1567 is related to that. I wrote issue #1599 to address this.

src/pymovements/gaze/gaze.py
```diff
@@ -1994,6 +2004,163 @@ def clone(self) -> Gaze:
...
+            acceleration) is present. (default: True)
+        trial_continuity : bool
+            Check that timestamps are strictly monotone increasing within each trial
+            and that no gap exceeds 5× the expected inter-sample interval.
+            (default: True)
+        sampling_rate_consistency : bool
+            Check that the empirical median ISI matches the declared sampling rate
+            within 5%. (default: True)
+        gaze_range : bool
+            Check that ≥95% of gaze samples fall within screen bounds. (default: True)
```
dkrako: The checks for `trial_continuity`, `sampling_rate_consistency` and `gaze_range` are a great idea! However the thresholds are arbitrary. Although we could extend the typing to `bool | float` to define custom thresholds it wouldn't be usable within `report_data_quality()` that way.

Do you have an idea on how to improve on that?

src/pymovements/gaze/gaze.py
```diff
@@ -1994,6 +2004,163 @@ def clone(self) -> Gaze:
...
+        source_path : str
+            Identifier for this gaze object (e.g. a file path). Included in the
+            ``affected_files`` field of any failing :py:class:`CheckResult`.
+            (default: ``''``)
+
+        Returns
+        -------
+        list[CheckResult]
+            One :py:class:`~pymovements.gaze.validation.CheckResult` per enabled
```
dkrako: Use the (minimal)  sphinx reference instead:

```suggestion
            One :py:class:`~pymovements.CheckResult` per enabled
```

This will resolve your sphinx nitpicky warnings.

src/pymovements/gaze/gaze.py
```diff
@@ -1994,6 +2004,163 @@ def clone(self) -> Gaze:
...
+            check, in the order listed above.
+
+        Examples
+        --------
+        >>> import polars as pl
+        >>> from pymovements.gaze.gaze import Gaze
```
dkrako: Wherever possible please always use the global namespace for imports in examples and documentation:

```suggestion
        >>> from pymovements import Gaze
```

src/pymovements/gaze/gaze.py
```diff
@@ -1994,6 +2004,163 @@ def clone(self) -> Gaze:
...
+        >>> samples = pl.DataFrame(
+        ...     {'time': [0, 1, 2], 'x': [0.0, 1.0, 2.0], 'y': [0.0, 1.0, 2.0]}
+        ... )
+        >>> gaze = Gaze(samples=samples, pixel_columns=['x', 'y'])
+        >>> results = gaze.validate()
+        >>> all(r.severity in {'pass', 'warning', 'error'} for r in results)
+        True
+        """
+        flag_fn_pairs: list[tuple[bool, object]] = [
```
dkrako: This should resolve the mypy error:

```suggestion
        flag_fn_pairs: list[tuple[bool, Callable[[Gaze, str], CheckResult]]] = [
```

src/pymovements/gaze/gaze.py
```diff
@@ -1994,6 +2004,163 @@ def clone(self) -> Gaze:
...
+            (trial_columns_exist, check_trial_columns_exist),
+            (trial_columns_dtype, check_trial_columns_dtype),
+            (time_column_exists, check_time_column_exists),
+            (gaze_components_defined, check_gaze_components_defined),
+            (trial_continuity, check_trial_continuity),
+            (sampling_rate_consistency, check_sampling_rate_consistency),
+            (gaze_range, check_gaze_range),
+        ]
+        return [
+            fn(self, source_path)  # type: ignore[operator]
```
dkrako: Please remove the type ignore by applying the suggestion above.

src/pymovements/gaze/gaze.py
```diff
@@ -1994,6 +2004,163 @@ def clone(self) -> Gaze:
...
+            levels: list[str] | None = None,
+            raise_on_error: bool = False,
+            output_path: Path | str | None = None,
+            source_path: str = '',
+    ) -> DataQualityReport:
+        """Generate a data quality report for this gaze object.
+
+        Runs validation checks via :py:meth:`validate` and computes quality
+        measures (data loss, fixation precision) for this single gaze file.
+        The result is a :py:class:`~pymovements.gaze.quality.DataQualityReport`
```
dkrako:
```suggestion
        The result is a :py:class:`~pymovements.DataQualityReport`
```

src/pymovements/gaze/gaze.py
```diff
@@ -1994,6 +2004,163 @@ def clone(self) -> Gaze:
...
+        GazeDataValidationError
+            If *raise_on_error* is ``True`` and any check produces an error
+            result.
+        ValueError
+            If any name in *checks* is not a valid check identifier.
+
+        Examples
+        --------
+        >>> import polars as pl
+        >>> from pymovements.gaze.gaze import Gaze
```
dkrako:
```suggestion
        >>> from pymovements import Gaze
```

src/pymovements/gaze/validation.py
```diff
@@ -0,0 +1,630 @@
...
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+"""Validation checks for individual :py:class:`~pymovements.gaze.Gaze` objects."""
```
dkrako:
```suggestion
"""Validation checks for individual :py:class:`~pymovements.Gaze` objects."""
```

src/pymovements/gaze/validation.py
```diff
@@ -0,0 +1,630 @@
...
+    >>> gaze = Gaze(
+    ...     samples=pl.DataFrame({'time': [0, 1, 2], 'x': [0.0, 1.0, 2.0], 'y': [0.0, 1.0, 2.0]}),
+    ...     pixel_columns=['x', 'y'],
+    ... )
+    >>> check_time_column_exists(gaze).severity
+    'pass'
+    """
+    schema = gaze.samples.schema
+
+    if 'time' not in schema:
```
dkrako:
```suggestion
    if 'time' not in gaze.samples.columns:
```

src/pymovements/gaze/validation.py
```diff
@@ -0,0 +1,630 @@
...
+            severity='error',
+            message=(
+                "No 'time' column found in the sample schema. "
+                'Specify time_column during Gaze initialisation or provide an Experiment '
+                'with a sampling_rate to auto-generate timestamps.'
+            ),
+            affected_files=[source_path] if source_path else [],
+        )
+
+    if type(schema['time']) not in _NUMERIC_DTYPES:
```
dkrako: Use the polars built-in method for dtypes instead:

```suggestion
    if not gaze.samples['time'].dtype.is_numeric():
```

This way you don't need to define `_NUMERIC_DTYPES` yourselves which might get outdated in future polars versions.

Reference: https://docs.pola.rs/api/python/stable/reference/api/polars.datatypes.DataType.html#polars.datatypes.DataType

src/pymovements/gaze/validation.py
```diff
@@ -0,0 +1,630 @@
...
+from __future__ import annotations
+
+from collections.abc import Callable
+from dataclasses import dataclass
+from dataclasses import field
+from typing import TYPE_CHECKING
+
+import polars as pl
+
+if TYPE_CHECKING:
```
dkrako, line 30: I don't think this `if` is necessary. Can you elaborate more on that?

dkrako (changes requested):
@IngaleOmkar @Insalatone 

Here's a preliminary review because I didn't manage to review all files yet. I will submit a further review EOD. Your code looks great so far but I have some minor comments that should be easy to address. Please use the minimal paths for doctests and sphinx references wherever possible (e.g., `pymovements.Gaze` instead of `pymovements.gaze.gaze.Gaze`).

The irrelevant CI errors are addressed in #1596, #1597 and #1598, and merged into main.

src/pymovements/dataset/dataset.py
```diff
@@ -35,6 +36,9 @@
 from pymovements._version import __version__
 from pymovements.dataset import dataset_download
 from pymovements.dataset import dataset_files
+from pymovements.dataset.data_quality import _compute_measures
+from pymovements.dataset.data_quality import DataQualityReport
+from pymovements.dataset.data_quality import GazeDataValidationError
```
dkrako: you can directly import from `pymovements.gaze`

src/pymovements/dataset/data_quality.py
dkrako, line 1: this file won't be necessary as you can directly import from `pymovements.gaze` in `dataset.py`.

src/pymovements/__init__.py
```diff
@@ -28,11 +28,14 @@
 from pymovements import transforms
 from pymovements import warnings
 from pymovements._version import __version__
+from pymovements.dataset import CheckResult
```
dkrako: these imports should use `pymovements.gaze`

src/pymovements/gaze/validation.py
```diff
@@ -0,0 +1,630 @@
...
+
+    Attributes
+    ----------
+    check_id : str
+        Short identifier, e.g. ``'trial_columns_exist'``.
+    severity : str
+        One of ``'pass'``, ``'warning'``, or ``'error'``.
+    message : str
+        Human-readable description of the outcome.
+    affected_files : list[str]
```
dkrako: BIDS defines a `Sources` metadata field for derivative data files: https://bids-specification.readthedocs.io/en/stable/derivatives/common-data-types.html

I guess we can just name this field `sources` then.

src/pymovements/gaze/validation.py
```diff
@@ -0,0 +1,630 @@
...
+        File paths of Gaze objects that failed. Empty on pass.
```
dkrako:
```suggestion
        File paths of source files that were checked. (default: ``[]``)
```

We may use this class for non-gaze objects in the future so let's be more general here. The 'empty on pass' can be left out as the source files should be reported on a passing check.

src/pymovements/gaze/validation.py
```diff
@@ -0,0 +1,630 @@
...
+
+    Examples
+    --------
+    >>> from pymovements.gaze.validation import CheckResult
```
dkrako:
```suggestion
    >>> from pymovements import CheckResult
```

src/pymovements/gaze/validation.py
```diff
@@ -0,0 +1,630 @@
...
+_NUMERIC_DTYPES: frozenset[type] = frozenset({
+    pl.Int8, pl.Int16, pl.Int32, pl.Int64,
+    pl.UInt8, pl.UInt16, pl.UInt32, pl.UInt64,
+    pl.Float32, pl.Float64,
+})
+
+_FLOAT_DTYPES: frozenset[type] = frozenset({pl.Float32, pl.Float64})
+
+
+@dataclass
```
dkrako:
```suggestion
@dataclass(frozen=True, slots=True)
```

the results shouldn't change after creation and dynamic attributes also won't be needed

src/pymovements/gaze/validation.py
```diff
@@ -0,0 +1,630 @@
...
+class CheckResult:
+    """Result of a single validation check.
+
+    Attributes
+    ----------
+    check_id : str
```
dkrako: let's rename this to `code`

src/pymovements/gaze/validation.py
```diff
@@ -0,0 +1,630 @@
...
+    ... )
+    >>> check_trial_columns_exist(gaze).severity
+    'pass'
+    """
+    if not gaze.trial_columns:
+        return CheckResult(
+            check_id='trial_columns_exist',
+            severity='pass',
+            message='No trial_columns declared; check skipped.',
+        )
```
dkrako, line 109: a passing check should also include the affected files

src/pymovements/gaze/validation.py
```diff
@@ -0,0 +1,630 @@
...
+                f'Available columns: {list(schema.keys())!r}'
+            ),
+            affected_files=[source_path] if source_path else [],
+        )
+
+    return CheckResult(
+        check_id='trial_columns_exist',
+        severity='pass',
+        message='All declared trial_columns are present in the sample schema.',
+    )
```
dkrako, line 129: a passing check should also include the affected files

src/pymovements/gaze/validation.py
```diff
@@ -0,0 +1,630 @@
...
+    ... )
+    >>> check_trial_columns_dtype(gaze).severity
+    'pass'
+    """
+    if not gaze.trial_columns:
+        return CheckResult(
+            check_id='trial_columns_dtype',
+            severity='pass',
+            message='No trial_columns declared; check skipped.',
+        )
```
dkrako, line 171: a passing check should also include the affected files

src/pymovements/gaze/validation.py
```diff
@@ -0,0 +1,630 @@
...
+
+    schema = gaze.samples.schema
+    float_cols = [
```
dkrako: you should check the opposite instead. what if the dtype is an object?

```python
    invalid_cols = [
      col for col in gaze.trial_columns
      if not (gaze.samples[col].dtype == pl.String or gaze.samples[col].dtype.is_integer()) 
    ]
```

src/pymovements/gaze/validation.py
```diff
@@ -0,0 +1,630 @@
...
+    if not gaze.trial_columns or 'time' not in gaze.samples.schema:
+        return CheckResult(
+            check_id='trial_continuity',
+            severity='pass',
+            message='Preconditions not met (no trial_columns or no time column); check skipped.',
+        )
+
+    schema = gaze.samples.schema
+    missing_tc = [c for c in gaze.trial_columns if c not in schema]
+    if missing_tc:
```
dkrako: if `trial_columns` are missing the data should be treated as one single trial. you should still validate the time continuity then.

src/pymovements/gaze/validation.py
```diff
@@ -0,0 +1,630 @@
...
+    ... )
+    >>> gaze = Gaze(samples=samples, trial_columns=['trial'], pixel_columns=['x', 'y'])
+    >>> check_trial_continuity(gaze).severity
+    'pass'
+    """
+    if not gaze.trial_columns or 'time' not in gaze.samples.schema:
+        return CheckResult(
+            check_id='trial_continuity',
+            severity='pass',
+            message='Preconditions not met (no trial_columns or no time column); check skipped.',
```
dkrako: actually the only precondition here is a time column. If there are no trial columns just treat the whole dataframe as a single trial

src/pymovements/gaze/validation.py
```diff
@@ -0,0 +1,630 @@
...
+        severity='pass',
+        message=f'Gaze coordinate columns present: {sorted(present)!r}.',
+    )
+
+
+def check_trial_continuity(gaze: Gaze, source_path: str = '') -> CheckResult:
+    """Check that timestamps within each trial are strictly monotone increasing.
+
+    Also checks that no gap between consecutive timestamps exceeds five times the
+    expected inter-sample interval (ISI), derived from
```
dkrako: please split the checks for monotonically increasing time values and maximum inter-sample gaps into two separate checks.

src/pymovements/gaze/validation.py
```diff
@@ -0,0 +1,630 @@
...
+
+    non_monotone_trials: list[str] = []
+    gap_trials: list[str] = []
+
+    sampling_rate = (
+        gaze.experiment.sampling_rate
+        if gaze.experiment is not None
+        else None
+    )
+    max_gap: float | None = 5.0 * (1000.0 / sampling_rate) if sampling_rate else None
```
dkrako: please add the threshold as a parameter instead of hardcoding it here

src/pymovements/gaze/validation.py
```diff
@@ -0,0 +1,630 @@
...
+    positive_diffs = diffs.filter(diffs > 0)
+
+    if len(positive_diffs) == 0:
+        return CheckResult(
+            check_id='sampling_rate_consistency',
+            severity='pass',
+            message='No positive time differences found; check skipped.',
+        )
+
+    empirical_isi = float(positive_diffs.median())  # type: ignore[arg-type]
```
dkrako: the type ignore shouldn't be necessary here. Can you post the error message that you get from mypy?

src/pymovements/gaze/validation.py
```diff
@@ -0,0 +1,630 @@
...
+    empirical_rate = 1000.0 / empirical_isi
+    deviation = abs(empirical_rate - declared_rate) / declared_rate
+
+    if deviation > 0.05:
```
dkrako: add this as a function parameter instead of hardcoding it

src/pymovements/gaze/validation.py
```diff
@@ -0,0 +1,630 @@
...
+    y_vals = non_null[coord_col].list.get(1)
+
+    in_range = (
+        (x_vals >= x_min) & (x_vals <= x_max)
+        & (y_vals >= y_min) & (y_vals <= y_max)
+    )
+    n_in_range = int(in_range.sum())
+    ratio = n_in_range / n_total
+
+    if ratio < 0.95:
```
dkrako: add this as a function parameter instead of hardcoding it

src/pymovements/gaze/quality.py
```diff
@@ -0,0 +1,624 @@
...
+
+
+__all__ = [
+    'CheckResult',
+    'DataQualityReport',
+    'GazeDataValidationError',
+]
+
+
+class GazeDataValidationError(Exception):
```
dkrako: rename this to `ValidationError`

src/pymovements/gaze/quality.py
```diff
@@ -0,0 +1,624 @@
...
+                trial_df = trial_df.rename({'data_loss_ratio': 'data_loss'})
+            trial_df = trial_df.with_columns(pl.lit(subject_id).alias('subject_id'))
+            trial_rows.append(trial_df)
+        except (pl.exceptions.PolarsError, ValueError, ArithmeticError):
+            continue
+
+    return trial_rows
+
+
+def _compute_measures(
```
dkrako: you import this in `dataset.py` so this function shouldn't be private

src/pymovements/gaze/quality.py
```diff
@@ -0,0 +1,624 @@
...
+
+    if 'trial' in levels:
+        trial_rows = _compute_trial_rows(gaze_list, fileinfo_rows, requested)
+        if trial_rows:
+            results['trial'] = pl.concat(trial_rows, how='diagonal')
+
+    return results
+
+
+def _run_report(
```
dkrako: you import this in `gaze.py` so this function shouldn't be private

src/pymovements/gaze/quality.py
```diff
@@ -0,0 +1,624 @@
...
+        gaze: Any,
```
dkrako:
```suggestion
        gaze: Gaze,
```

this means you need the same 

```python
if TYPE_CHECKING:
```

at the top as in `validation.py`

src/pymovements/gaze/quality.py
```diff
@@ -0,0 +1,624 @@
...
+``dataset.__init__`` imports ``dataset.dataset`` which imports
+``pymovements.gaze.Gaze``. Keeping the classes in the ``gaze`` package
+breaks that cycle while letting :py:class:`~pymovements.Dataset` re-export
+them from :py:mod:`~pymovements.dataset.data_quality` for backwards
+compatibility.
+"""
+from __future__ import annotations
+
+import json
+import warnings as _warnings
```
dkrako:
```suggestion
import warnings
```

src/pymovements/gaze/quality.py
```diff
@@ -0,0 +1,624 @@
...
+    >>> report = DataQualityReport(
+    ...     check_results=[CheckResult('time_column_exists', 'pass', 'OK')],
+    ... )
+    >>> report.passed
+    True
+    """
+
+    check_results: list[CheckResult] = field(default_factory=list)
+    measures: dict[str, pl.DataFrame] = field(default_factory=dict)
+    passed: bool = field(init=False)
```
dkrako: this should be implemented as a read-only property:

```python
@property
def passed(self) -> bool:
    return all(result.severity in {'pass', 'warning'} for result in self.check_results)
```

This way you won't need to do this in `__post_init__()` and can have a frozen dataclass `@dataclass(frozen=True)`

IngaleOmkar:
Hi @dkrako ! We've updated the PR now to address your feedback. Just a few points:
1. We've removed all hardcoded thresholds and allowed the users the ability to define these thresholds themselves. 
2. We have kept the IF statement for the circular dependency issue for now. Do let us know if you have a specific resolution for this in mind. 

Thanks for the thorough review!

src/pymovements/dataset/__init__.py
dkrako, line 1: the changes in this file don't seem to be necessary anymore as you import via `pymovements.gaze` in `dataset.py`. You can revert these.

src/pymovements/gaze/gaze.py
```diff
@@ -1994,6 +2005,202 @@ def clone(self) -> Gaze:
...
+            results.append(
+                check_sampling_rate_consistency(self, source_path, max_deviation=max_deviation),
+            )
+        if gaze_range:
+            results.append(check_gaze_range(self, source_path, min_fraction=min_fraction))
+        return results
+
+    def report_data_quality(
+            self,
+            checks: list[str] | None = None,
```
dkrako, line 2116:
```suggestion
            *
            checks: list[str] | None = None,
```

src/pymovements/gaze/validation.py
```diff
@@ -0,0 +1,741 @@
...
+    ... )
+    >>> check_time_monotone(gaze).severity
+    'pass'
+    """
+    sources = [source_path] if source_path else []
+
+    if 'time' not in gaze.samples.columns:
+        return CheckResult(
+            code='time_monotone',
+            severity='pass',
```
dkrako, line 360: This deserves it's own severity class as it's a precondition that is not being satisfied. Currently there are 'pass', 'warning' and 'error' severities. I think it would be best to split 'error'  into 'fail' (check failed, opposite of 'pass') and 'error' (check preconditions not met, computation errors and similar). This means all current 'error' CheckResults should be changed to 'fail'  and CheckResults like this here would be an 'error' due to not meeting check preconditions. A 'pass' however would definitely be misleading.

src/pymovements/gaze/validation.py
```diff
@@ -0,0 +1,741 @@
...
+            message="No 'time' column available; check skipped.",
+            sources=sources,
+        )
+
+    groups = _get_trial_groups(gaze)
```
dkrako: This returns an empty list in case trial columns are missing from the samples dataframe. the for loop below using this list will not execute and the checks will simply pass without any actual check. I think this should be implemented more robustly.

First, I would prefer using the `Gaze.split()`  method instead:

```python
    if gaze.trial_columns:
        groups = gaze.split()
    else:
        groups = [gaze]
```

As this is just an if-else the helper function probably won't be worth it any more.

In case a trial column is missing from the sample data frame the `Gaze.split()` method will raise a `ColumnNotFoundError`. When calling the helper function you should catch that, however it's a bit more tricky to return a meaningful check result here as we have a dependency between checks (i.e. if trial_columns_exist or check_time_column_exists fails then ).

```python
    if gaze.trial_columns:
        try:
           groups = gaze.split()
        except: polars.exceptions.ColumnNotFoundError:
            missing = [col for col in gaze.trial_columns if col not in gaze.samples.columns]
            return CheckResult(
                code='time_monotone',
                severity='error',
                message=(
                    f'trial_columns {missing!r} not found in sample schema. '
                    f'Available columns: {gaze.samples.columns!r}'
                ),
                sources=sources,
            )
    else:
        groups = [gaze]
```

Alternatively the check preconditions could be run more explicitly by reusing them at the beginning:

```python
    check_result = check_time_column_exists()
    if check_result.severity in {'fail', 'error'}:
        return CheckResult(
            code='time_monotone',
            severity='error',
            message="No 'time' column available.",
            sources=sources,
        )

    if gaze.trial_columns:   # the check is only relevant if the we have trial_columns
        check_result = check_trial_columns_exists()
        if check_result.severity in {'fail', 'error'}:
            return CheckResult(
                code='time_monotone',
                severity='error',
                message=check_result.message,  # no need for inferring missing columns this way
                sources=sources,
            )
```

I prefer the latter solution as you can just reuse existing check logic and make check dependencies more obvious.

src/pymovements/gaze/validation.py
```diff
@@ -0,0 +1,741 @@
...
+    if len(positive_diffs) == 0:
+        return CheckResult(
+            code='sampling_rate_consistency',
+            severity='pass',
+            message='No positive time differences found; check skipped.',
+            sources=sources,
+        )
+
+    median_isi = positive_diffs.median()
+    if not isinstance(median_isi, (int, float)):  # pragma: no cover
```
dkrako: this code seems unreachable. What input can result in this?

src/pymovements/gaze/validation.py
dkrako: yes my bad, I forgot pylint complains about that even if it's in a shared submodule

tests/unit/gaze/report_data_quality_test.py
dkrako, line 1: rename this file to `report_data_quality_test.py`

src/pymovements/gaze/validation.py
```diff
@@ -0,0 +1,741 @@
...
+    ... )
+    >>> check_max_gap(gaze).severity
+    'pass'
+    """
+    sources = [source_path] if source_path else []
+
+    if 'time' not in gaze.samples.columns:
+        return CheckResult(
+            code='max_gap',
+            severity='pass',
```
dkrako: this should be changed to 'error' (as it breaks a check precondition)

src/pymovements/gaze/validation.py
```diff
@@ -0,0 +1,741 @@
...
+            message="No 'time' column available; check skipped.",
+            sources=sources,
+        )
+
+    if gaze.experiment is None or gaze.experiment.sampling_rate is None:
+        return CheckResult(
+            code='max_gap',
+            severity='pass',
```
dkrako: this should be changed to 'error' (as it breaks a check precondition)

src/pymovements/gaze/validation.py
```diff
@@ -0,0 +1,741 @@
...
+
+    declared_rate = gaze.experiment.sampling_rate
+    diffs = gaze.samples['time'].cast(pl.Float64).diff().drop_nulls()
+    positive_diffs = diffs.filter(diffs > 0)
+
+    if len(positive_diffs) == 0:
+        return CheckResult(
+            code='sampling_rate_consistency',
+            severity='pass',
+            message='No positive time differences found; check skipped.',
```
dkrako, line 591: will this be caught by check_time_monotone?

dkrako (changes requested):
@IngaleOmkar @Insalatone 

Thanks a lot for all your changes, this PR looks excellent! I have mostly just a few minor comments that should be quick and easy to address.

However, two comments are slightly major and relate to a similar issue: Please consider splitting your 'error' severity into 'fail'  and 'error' to discern between check failures and failures due to preconditions not being met (you can see this in a similar way with pytest where we have a FAIL for failing tests but an ERROR for example when there's an error during test parametrization/setup). You can treat this as optional as it won't be required for merging, but consider this as a nice-to-have as it will resolve some of the skip-ambiguities in your current implementation.

IngaleOmkar:
Hi @dkrako , I've made the changes. However, one of the tests (`tests / docs (ubuntu-latest) (pull_request)` ) is failing. I believe it is unrelated to the code changes. Can you confirm this? Is it possible to rerun this test?

cbueth:
Yes, that one was unrelated. I restarted it.

IngaleOmkar:
Hi @dkrako, we've implemented all of the suggestions from the latest review (including the fail severity suggestion). Could you please take a look? Thanks!

src/pymovements/gaze/validation.py
dkrako: After some playing around I'm quite sure that this code is not reachable because the type is already casted to float and all nulls are dropped. Thus this check result is already covered in the branch above (the one for no positive diffs).

If you think it is reachable, please write a test that covers this line. Otherwise, please remove the code block.

However, if you like you could add more checks above before casting (maybe via a new `check_time_dtype`) or dropping nulls (via a new `check_time_non_null`) but it's not required.

src/pymovements/gaze/validation.py
dkrako: I'm asking because there is no check that would lead to `fail` for that case. Additional checks as `check_time_dtype` and a`check_time_non_null` could catch that.

src/pymovements/gaze/validation.py
```diff
@@ -0,0 +1,751 @@
...
+            if gaze.trial_columns:
+                key_vals = {c: grp[c][0] for c in gaze.trial_columns if c in grp.columns}
+                non_monotone.append(str(key_vals))
+            else:
+                non_monotone.append('(single trial)')
+
+    if non_monotone:
+        return CheckResult(
+            code='time_monotone',
+            severity='warning',
```
dkrako: I think this should be a `fail` as this is critical. Having non-monotone timestamps points to a trial column misconfiguration or corrupt data. Moreover, having monotone timestamps is critical to a lot of preprocessing algorithms.

dkrako (changes requested):
@IngaleOmkar @Insalatone Thanks for the updates, we are one step before the finish line! 

There remains one small bug in `check_time_monotone` where if individual columns from `trial_columns` are missing in `samples` this will lead to a `pass`. It should return an `error` instead, as check preconditions are not met. Please review my [comment](https://github.com/pymovements/pymovements/pull/1579#discussion_r3553119007) from last week that proposes a refactor of your `_get_trial_groups()` helper function by using the `Gaze.split()` method instead. I prefer the latter code proposal as you can just reuse existing check logic and make check dependencies more obvious.

The same bug is also present in `check_max_gap` as it reuses `_get_trial_groups()` in the same way.

I think it is debatable where to return a `pass`, `error` or `warning` as a check result, e.g., for me it makes more sense to return a `fail` instead of a `warning` when `check_time_monotone()` fails because this is actually what the test is about and it's critical.

Apart from that please eliminate the unreachable code in `check_sampling_rate_consistency` (or write a test that covers that to prove the code is reachable).

All in all this shouldn't take much effort to finalize this.

The additional checks are optional and not required for merging as they can be easily added afterwards.

IngaleOmkar:
@dkrako I have made the requested changes. Do let me know if any further changes are needed. Thanks!

dkrako (approved):
@IngaleOmkar @Insalatone 

Thanks for the continued work on this, the latest revision looks great! The fix for missing declared trial columns in `check_time_monotone`/`check_max_gap` (now `'error'` instead of `'pass'`) looks good. Below are my remaining findings, roughly in order of importance, but see the summary at the end: no further action is needed from your side.

## Correctness

**1. `check_gaze_range` counts null/NaN coordinate samples as out of range.** The filter `pl.col(coord_col).is_not_null()` only drops rows where the whole list is null. Rows like `[NaN, NaN]` or `[null, null]` (the typical representation of blinks/tracking loss) stay in the denominator, and their comparisons evaluate false/null, so they count as out-of-range. Any recording with more than 5% blink loss will warn about gaze range even if every valid sample is on screen. Blink loss belongs to the `data_loss` measure; this check should filter element-level null/NaN before computing the ratio.

**2. Unknown `levels` and `measures` are silently ignored.** `checks=['typo']` raises `ValueError`, but `levels=['Trial']` or `measures=['dataloss']` silently produces nothing. All three parameters should be validated the same way.

**3. Minor: `compute_measures` builds `pl.DataFrame(rows_all)` from dicts with heterogeneous keys** (`session_id`/measure keys present only for some files). Polars schema inference scans a limited number of rows, so a dataset where a key first appears after row ~100 can error out. Emitting all keys with `None` in every row avoids this.

## Design / code quality

- `Dataset.report_data_quality()` reimplements `run_report()` almost line for line (check-id validation, per-check raise, warnings capture, save). This should be extracted into one shared helper taking a list of (gaze, source_path) pairs.
- Broad `except (PolarsError, ValueError, ArithmeticError)` blocks silently degrade measures to `None` or silently switch `data_loss` to the semantically different `_compute_data_loss_simple` (whole-row nulls only, no gaps/NaN). A quality-report tool should at least record these fallbacks in the report/warning log.
- In `_compute_file_row`, `_compute_data_loss_simple(gaze, coord_col) if coord_col else None` is dead code, the branch is already guarded by `coord_col is not None`.
- Binocular data: `list.get(0)`/`list.get(1)` silently evaluates left-eye components only for 4/6-component columns; this should be documented or handled.
- The `warnings.catch_warnings(record=True)` block means warnings raised during checks/measures never reach the user's console, only `report.warning_log`. Probably intended, but worth documenting (and it's not thread-safe).
- Subject/session mapping assumes `fileinfo` columns named `subject_id`/`session_id`; datasets whose `filename_format` uses other group names silently fall back to file-index strings as subject ids.

## API / docs

- `src/pymovements/__init__.py`: adding `'datasets'` to `__all__` is unrelated to this PR and only works because `importlib.resources` imports the subpackage as a side effect. Also `'ValidationError'` is inserted out of alphabetical order in both `__init__` files and in `dataset.rst` (between `DatasetPaths` and `ResourceDefinition`).
- The tutorial notebook still uses deep imports (`from pymovements.gaze.gaze import Gaze`, `from pymovements.gaze.quality import ValidationError`) instead of the minimal paths requested earlier (`import pymovements as pm`). The notebook also has empty top-level `metadata` (no kernelspec).
- `Dataset.report_data_quality()` has a fully commented-out Example block. It should either be a real doctest or dropped.

## Tests

Coverage is genuinely broad (pass/warn/fail/error branches per check, threshold passthrough, BIDS output, raise-on-error). Two structural notes:

- `TestDatasetReportDataQuality._call_report` in `data_quality_test.py` reimplements the production logic inside the test file and then tests that copy, so it proves nothing about `Dataset.report_data_quality()`. The `TestDatasetReportDataQualityDirect` class already covers the real method; the mock-based class should be deleted or reworked.
- `_make_gaze` bypasses `Gaze.__init__` via `Gaze.__new__` with a hardcoded list of private attributes (`_metadata`, `validations`, ...) and is duplicated (along with the experiment factory) across both test files. Any `Gaze` refactor will break these silently; one copy should live in a shared test util, with the `__new__` trick kept only where invalid state is actually needed.

## Summary

Strong contribution, thank you both! Rather than another review round, I'll merge this as soon as CI passes and address the points above in a follow-up PR.

PR #1588, opened 2026-05-21
docs: add acknowledgements

dkrako:
# Description

Adds a new funding page to our sphinx documentation. Also adds a funding section to the readme.

# Implemented changes

- [x] add funding page to about-us page
- [x] add readme funding section


cbueth:
The placing is looking good, but funding info still missing. There will be a more detailled text in the dedicated page, and a short paragraph on the landing page?

dkrako:
I added the funding info from our previous papers, together with logos of the organizations. However I'm not fully satisfied regarding their scaling as if I set the height the aspect ratio is not kept constant. I calculated the ratios from the source images to keep the aspect ratio. It's not great but probably good enough for this first iteration.

cbueth:
> I added the funding info from our previous papers, together with logos of the organizations. However I'm not fully satisfied regarding their scaling as if I set the height the aspect ratio is not kept constant. I calculated the ratios from the source images to keep the aspect ratio. It's not great but probably good enough for this first iteration.

The aspect ration of the DAAD logo can be fixed adding a box to the source. But it woud be nicer not having to serve the logo from the repo. I tried using css on the original file but was not successful.
```
< <svg width="380" height="102" xmlns="http://www.w3.org/2000/svg"
---
> <svg width="380" height="102" viewBox="0 0 380 102" xmlns="http://www.w3.org/2000/svg"
```

For the other ones, only give one `:height:` or `:width:` so it will not stretch it. I will push a suggestion.

docs/source/about-us/acknowledgments.rst
```diff
@@ -0,0 +1,67 @@
...
+   .. grid-item::
+      .. image:: https://fundit.fr/sites/default/files/styles/max_650x650/public/actors/2318-autorite-nationale-pour-recherche-scientifique-roumanie-ancs.png
+         :target: https://www.cncs-nrc.ro
+         :width: 285
+         :height: 180
+         :align: center
+
+      the Romanian National Research Council (CNCS) through the Executive Agency for Higher Education, Research, Development and Innovation Funding (UEFISCDI) under grant PN-IV-P2-2.1-TE-2023-2007 (InstRead)
+
+and was supported by work from
```
cbueth, line 50: This line appears a bit off on the grid. I'd stick with only the grid and not a flowing text.

dkrako:
@cbueth Thank you so much for improving the page, your choice to use cards is great!

PR #1593, opened 2026-06-09
docs: update MCFW-Gaze dataset with paper citation

dkrako:
## Description

As the MCFW-Gaze paper is now fully published, I added a citation to the dataset description.

As the citations for the paper and the dataset resulted in the same bibliography label I had to update `AuthorYearLabelStyle` in our sphinx `conf.py` to automatically add suffixes to avoid collisions.


dkrako:
I had to update the way how we set citation labels in `docs/src/conf.py`. The way it was implemented labels would collide if the same author published multiple papers per year. Should be fixed within this PR.

cbueth (commented):
All references were updated, looks good.

The change in `conf.py` will now use the raw bibtex keys, e.g. [Nagasawa et al., 2026] -> [MCFW-Gaze-Paper]?
The alternative would be keeping 'author_year' and relying solely on the dedup suffixes, but that might not cleanly handle collisions at the reference level (the built-in author_year reference style and the custom AuthorYearLabelStyle are separate systems).

If you're fine with reference labels showing bibtex keys, the changes look good and I'd accept.

cbueth:
If we set `bibtex_reference_style = 'author_year_style'` (the added custom class) instead of `'label'`. Then both in-text citations and reference list entries use the same dedup-aware logic, preserving author-year labels everywhere: `bibtex_reference_style = 'author_year_style'`
That way I would expect [Nagasawa et al., 2026] and [Nagasawa et al., 2026a] in both places, rather than [MCFW-Gaze-Paper] in the reference list.

How was this tested?

dkrako:
> If we set `bibtex_reference_style = 'author_year_style'` (the added custom class) instead of `'label'`. Then both in-text citations and reference list entries use the same dedup-aware logic, preserving author-year labels everywhere: `bibtex_reference_style = 'author_year_style'` That way I would expect [Nagasawa et al., 2026] and [Nagasawa et al., 2026a] in both places, rather than [MCFW-Gaze-Paper] in the reference list.
> 
> How was this tested?

You're right, it will just create a,b,c suffixes (26 max). I didn't write any tests yet but will do

docs/source/conf.py
```diff
@@ -276,13 +277,31 @@ def setup(app):
 # -- Options for BibTeX ------------------------------------------------------
 bibtex_bibfiles = ['bibliography.bib']
 bibtex_default_style = 'author_year_style'
-bibtex_reference_style = 'author_year'
+bibtex_reference_style = 'label'
```
dkrako: resolving the correct references only seems to work with `label`. `author_year` leads to duplicated references.

dkrako: setting it to `author_year_style` does not work: 
```ImportError: plugin sphinxcontrib.bibtex.style.referencing.author_year_style not found ```

dkrako:
added some tests and verified the build outputs: https://pymovements--1593.org.readthedocs.build/en/1593/datasets/MCFW-Gaze.html#mcfw-gaze

dkrako:
> If we set `bibtex_reference_style = 'author_year_style'` (the added custom class) instead of `'label'`. Then both in-text citations and reference list entries use the same dedup-aware logic, preserving author-year labels everywhere: `bibtex_reference_style = 'author_year_style'` That way I would expect [Nagasawa et al., 2026] and [Nagasawa et al., 2026a] in both places, rather than [MCFW-Gaze-Paper] in the reference list.

I'm confused. It does not output [MCFW-Gaze-Paper] in the reference list. It outputs [Nagasawa et al., 2026] and [Nagasawa et al., 2026a] correctly (using `label`).

dkrako:
I wouldn't like to put much more effort in this. The global bibliography page will be removed anyway in the future.

cbueth:
> added some tests and verified the build outputs: https://pymovements--1593.org.readthedocs.build/en/1593/datasets/MCFW-Gaze.html#mcfw-gaze

Thanks for the reference, this looks good to me. Sorry for the confusion, I was just asking, as I was not sure about the behaviour just by reading the diffs and the PR did not specify what updates were done and how this was verified.

cbueth (approved):

PR #1601, opened 2026-07-06
feat: allow passing column indices to `gaze.from_numpy`

### LINKED ISSUE #500: improve usability of gaze.from_numpy() by allowing lists of column indices
## Description of the problem

The current implementation expects a named list of columns for the numpy array:
https://github.com/aeye-lab/pymovements/blob/279d886c9c380a3bf17439c08cb30cf84ad8aa51/src/pymovements/gaze/integration.py#L33-L42

This list of names is then used to specify the columns in the data via `pixel_columns` etc.

## Description of a solution

Instead it shouldn't be necessary to provide a list of names, but we should be able to specify the columns by their index.

```python
def from_numpy(
        data: np.ndarray,
        schema: list[str] = None,
        experiment: Experiment | None = None,
        orient: Literal['col', 'row'] = 'col',
        pixel_columns: list[str] | list[int] | None = None,
        position_columns: list[str] | list[int] | None = None,
        velocity_columns: list[str] | list[int] | None = None,
        acceleration_columns: list[str] | list[int] | None = None,
) -> GazeDataFrame:
```

kevinmanase:
Fixes #500.

`gaze.from_numpy()` accepted only column names for `trial_columns`, `time_column`, `pixel_columns`, `position_columns`, `velocity_columns`, `acceleration_columns`, and `distance_column` when using a single `samples` array. This makes it possible to pass column indices instead, so callers do not need to invent/provide a schema just to select columns from a numpy array.

The implementation resolves integer specifiers against the Polars dataframe columns after `pl.from_numpy(...)`, so it works with explicit schemas and with the default `column_0`, `column_1`, ... names.

Verification:

```console
$ uv run --extra dev -- python -m pytest tests/unit/gaze/integration_numpy_test.py -q
.............                                                            [100%]
13 passed in 0.34s

$ python3 -m py_compile src/pymovements/gaze/integration.py
```


src/pymovements/gaze/integration.py
```diff
@@ -32,6 +33,27 @@
...
+
+
+def _resolve_columns(
+        columns: list[ColumnSpecifier] | None,
+        available_columns: list[str],
+) -> list[str] | None:
+    """Resolve a list of column names or indices against a dataframe's columns."""
+    if columns is None:
+        return None
+    return [available_columns[column] if isinstance(column, int) else column for column in columns]
```
cbueth: Here you can use the functionality.

```suggestion
    return [_resolve_column(col, available_columns) for col in columns]
```

tests/unit/gaze/integration_numpy_test.py
```diff
@@ -126,6 +126,73 @@ def test_from_numpy_with_schema():
...
+            [101, 102, 103, 104],
+            [100, 100, 100, 100],
+            [0, 1, 2, 3],
+            [4, 5, 6, 7],
+            [9, 8, 7, 6],
+            [5, 4, 3, 2],
+            [1, 2, 3, 4],
+            [5, 6, 7, 8],
+            [2, 3, 4, 5],
+            [6, 7, 8, 9],
```
cbueth: Using `@pytest.mark.parametrize`, this test could be parametrized and include multiple test cases.

Generally, please also cover the edge cases and test with:
1. schema + indices: to verify column indices resolve to named columns when a schema is provided
2. `trial_columns` as a single `int`: to exercise `trial_columns=0` (triggers the `isinstance(trial_columns, list)` -> `False` branch -> `_resolve_column`)

src/pymovements/gaze/integration.py
```diff
@@ -32,6 +33,27 @@
 from pymovements.gaze.experiment import Experiment
 from pymovements.gaze.gaze import Gaze
 
+ColumnSpecifier = str | int
```
cbueth: To resolve the docs CI warning try this: 

Add a Sphinx `#:` doc-comment, so the cross-reference resolves. This makes `:py:class:\ColumnSpecifier\` resolvable in docstrings. Without it, the `ref.class` warning fires for every parameter docstring that mentions `ColumnSpecifier`.

```suggestion
#: A column can be specified by name (str) or index (int).
ColumnSpecifier = str | int
```

Alternatively, you can try to add to `napoleon_type_aliases` in `conf.py` an entry for this `'ColumnSpecifier': 'str | int',`. This would replace `ColumnSpecifier` with `str | int` inline in rendered docs, avoiding the reference entirely. But we lose the semantic name.

From the top of my head I am not totally sure what will work out best.

cbueth (changes requested):
Hello @kevinmanase, thank you for freely working on this issue. Generally it is in good state, there are just three suggestions I put into the code.

This PR implements issue #500 by allowing column indices (int) alongside column names (str) in `from_numpy()` parameters. It adds `_resolve_column` and `_resolve_columns` helper functions, a `ColumnSpecifier = str | int` type alias, updates type annotations and docstrings, and adds a test for column indices without a schema.

CI is passing, except the docs job, se comment.

Best, Carlson

kevinmanase:
Thanks for the review, Carlson. I pushed cc7d107 plus follow-up d581acc to address the requested changes:

- `_resolve_columns()` now reuses `_resolve_column()`; I added overloads so the return types stay precise.
- Expanded the column-index test with `pytest.mark.parametrize` to cover:
  - no schema + index lists
  - schema + index lists
  - `trial_columns=0` as a single int
- Fixed the docs warning while keeping the `ColumnSpecifier` parameter docstrings consistent with the actual type hints.

Local validation:

```console
$ env -u PYTHONPATH PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run --extra dev python -m pytest tests/unit/gaze/integration_numpy_test.py -q
15 passed in 0.10s

$ env -u PYTHONPATH uv run --extra dev pylint --rcfile=pylintrc --output-format=parseable --disable=missing-class-docstring,missing-function-docstring,protected-access tests/unit/gaze/integration_numpy_test.py && env -u PYTHONPATH uv run --extra dev pylint --rcfile=pylintrc --output-format=parseable --ignore-paths=$(pwd)/src/pymovements/_version.py src/pymovements/gaze/integration.py
Your code has been rated at 10.00/10
Your code has been rated at 10.00/10

$ env -u PYTHONPATH uv tool run --from pydoclint==0.9.0 pydoclint --config=pyproject.toml src/pymovements/gaze/integration.py tests/unit/gaze/integration_numpy_test.py
🎉 No violations 🎉

$ env -u PYTHONPATH uv run --extra docs sphinx-build --color -W --keep-going -d docs/doctree -b html docs/source docs/build
build succeeded.
```

cbueth:
Thanks for integrating the feedback. It seems the pydoclint is still failing, but I assume you will be able to fix this running the precommit hooks locally and reviewing the messages.
I would suggest that @dkrako adds their opinion next. He has some other reviews on priority, but surely will come by soon.

kevinmanase:
Thanks — I pushed follow-up 623c33f for the remaining `pydoclint`/docs cross-reference issue.

The latest PR head is now green on the relevant checks:

- pre-commit.ci
- docs / ReadTheDocs
- docs on Ubuntu + Windows
- pylint
- minimum requirements
- Python 3.13 tests on Ubuntu, macOS, and Windows
- codecov patch

I’ll wait for @dkrako’s review from here.

src/pymovements/gaze/integration.py
dkrako: As this is just a simple union of two basic types I don't think `ColumnSpecifier` is really worth it. I see the point in making it a bit more descriptive but in the end this increases the number of characters needed (`str | int` is shorter) and a user will understand `str | int` right away while they would have to check again what `ColumnSpecifier` really resolves to.

src/pymovements/gaze/integration.py
```diff
@@ -334,10 +378,10 @@ def from_numpy(
         time_column=time_column,
         time_unit=time_unit,
         trial_columns=trial_columns,
-        pixel_columns=pixel_columns,
-        position_columns=position_columns,
-        velocity_columns=velocity_columns,
-        acceleration_columns=acceleration_columns,
+        pixel_columns=cast(list[str] | None, pixel_columns),
```
dkrako: the changes in this block shouldn't be necessary. this part of the code is only run if `samples` is None, e.g., the pixels are passed like so `from_numpy(pixel=pixel_np_array)` and a new combined `samples` data frame with named columns is constructed on the fly.

src/pymovements/gaze/integration.py
kevinmanase: Agreed — removed `ColumnSpecifier` and now use `str | int` directly in the overloads, public type hints, and docstrings. I also removed the alias-specific Sphinx configuration.

src/pymovements/gaze/integration.py
kevinmanase: Removed the casts. The separately supplied array path now keeps its generated string column names in `generated_*_columns` locals, so they pass to `Gaze` as `list[str] | None` without conflating them with the input column specifiers.

src/pymovements/gaze/integration.py
```diff
@@ -33,6 +34,35 @@
...
+def _resolve_column(column: str | int | None, columns: list[str]) -> str | None:
+    """Resolve a column name or index against a dataframe's columns."""
+    if column is None:
+        return None
+    if isinstance(column, int):
+        return columns[column]
+    return column
+
+
+def _resolve_columns(
```
dkrako, line 66: this might raise an `IndexError` in case the number of columns is lower than the column index. it would probably be best to catch that and provide an informative error message.

dkrako (changes requested):
@kevinmanase Sorry for reviewing this so late, I was so busy during the last weeks. Thank you so much for your initiative, your code looks great already, there are just a few minor comments I have.

Apart from that it might be worth evaluating if passing `schema` is really necessary if only indices are given. In that case we might generate a `schema` on the fly. But that can be addressed in a separate PR after merging this one.

src/pymovements/gaze/integration.py
dkrako: Thank you so much for your quick reaction! Yes, the original code was probably a bit confusing there

dkrako: As this wasn't addressed well in the original code I added some value errors in [`c650b08`](https://github.com/pymovements/pymovements/pull/1601/commits/c650b08332000c90416de15bbfe39c62568e7da9)

src/pymovements/gaze/integration.py
kevinmanase: Good catch — `_resolve_column()` now catches the out-of-range lookup and raises an `IndexError` that identifies the invalid index and the number of available columns. I also rebased this onto your `c650b08`/`c208762` updates, kept the expanded tests, and corrected the no-schema expectation to use Polars’ generated `column_0` name.

dkrako (approved):
@kevinmanase Excellent! Thank you so much for taking care and your lightning-speed responses! I added one further check for a wrong type because of pythons implicit bool->int casting and an otherwise cryptic error. This PR is now ready to be merged.

dkrako:
One thing I forgot and strangely the welcome bot didn't post in this PR here: Feel encouraged to add yourself to our [`CITATION.cff`](https://github.com/pymovements/pymovements/blob/main/CITATION.cff) file (including your ORCID). This way your authorship will be correctly attributed on our [zenodo page](https://doi.org/10.5281/zenodo.17397139). I will keep this open for now, if you don't want to add yourself please react with a thumbs down or similar.

kevinmanase:
done! thanks :)

PR #1603, opened 2026-07-06
feat!: add checksum verification to WebSource and make integrity checks opt-out

dkrako:
## :boom: Breaking Change

This PR removes trying out changing the `https` prefix to `http` in an url in case of a failed download. This was a potential security risk and is eliminated with this PR.

## Description

This pull request introduces an optional check_integrity parameter to the dataset download API, allowing users to skip MD5 checksum verification when downloading resources.

`WebSource` now encapsulates checksum computation and verification, keeping download-related functionality and checksum metadata together (adding the logic to `DatasetFile` leads to a circular import). In addition, dataset downloads now accept an optional check_integrity parameter, allowing users to skip checksum verification when downloading resources. Integrity verification remains enabled by default, preserving the existing behavior.

In addition, the download workflow has been refactored to improve error handling around checksum verification by introducing a dedicated ChecksumError exception and clarifying the handling of existing local files. If a checksum verification fails, the user is now presented with the expected and actual checksums. As a byproduct a potential security risk was eliminated by removing trying out changing the `https` prefix to `http` in an url in case of a failed download.

A nice benefit of having `WebSource.checksum()` is using the public API for calculating checksums in the dataset contribution guide.

## Implemented changes

- [x] Added `checksum()` and `verify_checksum()` methods to `WebSource`
- [x] Added a check_integrity parameter to the public dataset download API and propagated it through the complete download pipeline.
- [x] Updated the download logic to optionally skip MD5 verification for both existing and newly downloaded files while maintaining checksum validation as the default behavior.
- [x] Refactored integrity checking to raise a dedicated ChecksumError, improved handling of existing files with invalid checksums, and updated unit tests to cover the new API and behavior.
- [x] Added expected and actual checksum output when a ChecksumError occurs

## How Has This Been Tested?

- [x] added tests for `WebSource.download()` in case target file already exists
- [x] added tests for success and exceptions of `Websource.checksum()`
- [x] added test for `WebSource.verify_checksum()`

## Checklist:


tests/unit/dataset/websource_test.py
```diff
@@ -527,19 +597,78 @@ def test_websource_download_first_of_two_mirrors_gaze_fails(
...
+        pytest.param('monocular_example.csv', 'b56e208f6442bab645789defe865183b', id='csv'),
+    ],
+)
+class TestWebSourceChecksum:
+    def test_websource_checksum(self, filename, expected_checksum, testfiles_dirpath):
+        path = testfiles_dirpath / filename
+
+        result = WebSource.checksum(path)
+
+        assert result == expected_checksum
```
dkrako, line 640: There is a really strange issue with calculating checksums for text files on Windows: the results are wrong, even when using the built-in `CertUtil`  tool. How can this be? I'm left totally clueless. Can it be a problem with how this is tested, not the calculation per se?

This is the implementation using os builtins:

https://github.com/pymovements/pymovements/blob/c36610e99d3c77a900c0d753da791158c6a4c434/src/pymovements/dataset/websource.py#L210-L228

and this is the implementation prior to this:

https://github.com/pymovements/pymovements/blob/a1c1a307a297c809edc34f894466ec30f3b1b695/src/pymovements/dataset/websource.py#L213-L218

Both result in the wrong checksum '4a539e0c08d2075ad4410ca40f55cce3' for the csv file on windows.

cbueth:
For repository test files adding `*.csv text eol=lf` to `.gitattributes` should fix it.

dkrako:
> For repository test files adding `*.csv text eol=lf` to `.gitattributes` should fix it.

@cbueth Thank you so much, this did it!

tests/unit/dataset/websource_test.py
dkrako: I will revert the changes to the subprocess solution

docs/source/contributing/contributing-datasets.ipynb
```diff
@@ -115,22 +115,16 @@
...
-    "```\n",
-    "\n",
-    "or Python code:\n",
-    "\n",
-    "```py\n",
-    "from pymovements.dataset.websource import _calculate_md5\n",
-    "_calculate_md5(\"path/to/gaze.csv\")\n",
+    "```python\n",
+    "from pymovements import WebSource\n",
+    "WebSource.checksum(''path/to/your/dataset.zip')\n",
```
cbueth: Tiny typo:

```suggestion
    "WebSource.checksum('path/to/your/dataset.zip')\n",
```

src/pymovements/dataset/websource.py
```diff
@@ -178,45 +257,48 @@ def _download_file(
...
-    except OSError as e:
-        if url[:5] == 'https':
-            print('Download failed. Trying https -> http instead.')
-            url = url.replace('https:', 'http:')
-
-            if verbose:
-                print(f'Downloading {url} to {filepath}')
-            _download_url(url=url, destination=filepath, verbose=verbose)
-        else:
-            raise e
```
cbueth, line 213: This error has been removed. Where it this covered elsewhere?

tests/unit/dataset/websource_test.py
```diff
@@ -324,12 +392,6 @@ def test__DownloadProgressBar_tsize_not_None():
             'WebSource.url must not be None',
             id='url_none',
         ),
-        pytest.param(
-            WebSource(url='https://example.com/test.gz.tar', filename=None),
-            AttributeError,
-            'WebSource.filename must not be None',
-            id='filename_none',
-        ),
```
cbueth, line 332: `WebSource.download()` still validates `self.filename is not None`. Either keep both or remove both.

src/pymovements/exceptions.py
```diff
@@ -20,6 +20,38 @@
...
+from dataclasses import dataclass
+from pathlib import Path
+
+
+@dataclass(slots=True, eq=False)
+class ChecksumError(Exception):
+    """Exception raised when a checksum integrity check fails.
+
+    Attributes
+    ----------
```
cbueth, line 32: Looking super good!

cbueth (approved):
Thanks for the PR. I can approve, dependent conditionally on fixing the typo and giving a reason why the  `OSError` and https -> http retry was removed.

src/pymovements/dataset/websource.py
dkrako: You're right, I removed the replacing of https with http because I regarded that as unsafe. I then forgot to add that to the PR description. In consequence we have a breaking change here that needs to be documented in the PR at least.

I still think that we shouldn't do this, but maybe you think there could be a valid use case for this? The only thing I came up with is a ssl/certificates failure, but even there I would rather recommend fixing the ssl error instead of defaulting to http.

cbueth: Ok, valid point. I think for our case it is not so crucial. Fine for me, feel free to close.

tests/unit/dataset/websource_test.py
dkrako: good catch, I reverted removing the test

dkrako: Yes I did a small spike to automatically infer the filename from the URL but it seemed to be more complicated for redirected urls like those from OSF. I created #1609 and labeled it as nonessential.

PR #1608, opened 2026-07-15
chore!: remove `data_loss_ratio` and `data_loss_ratio_blinks` from parsing

### LINKED ISSUE #893: Store calibration and other metadata in well-documented form

### LINKED ISSUE #1584: Remove `data_loss_ratio` and `data_loss_ratio_blinks` from parsed metadata
Remove `pre_processed_metadata['data_loss_ratio']` and `pre_processed_metadata['data_loss_ratio_blinks']` fields from `pre_processed_metadata`.

https://github.com/pymovements/pymovements/blob/95a30e2cdfa21bf1e0b53fe955390137da70e63c/src/pymovements/gaze/_utils/_parsing_eyelink.py#L891-L894

- [ ] Remove the two fields from metadata dict
- [x] Test equality using measure_event_ratio and `pymovements.measure.samples.measures.data_loss`.
- [ ] see if to be removed parsing tests have interesting test cases for measure_event_ratio

cbueth:
## Description

This PR removes the `data_loss_ratio` and `data_loss_ratio_blinks` fields from `pre_processed_metadata` in the EyeLink ASC parser. These were EyeLink-specific values computed during parsing and are now replaced by format-independent equivalents:
- a Polars expression applicable to any gaze DataFrame `pymovements.measure.data_loss(column, sampling_rate=sr, unit='ratio')`
- a Gaze method that computes blink coverage ratio from events `gaze.measure_events_ratio('blink_eyelink', sampling_rate=sr)`

Both are functionally equivalent when applied per-trial, see comment https://github.com/pymovements/pymovements/issues/1584#issuecomment-4967488359. The `sampling_rate` argument triggers the `+dt` inclusive-endpoint convention, matching the old metadata formula.

After this, `pre_processed_metadata` still has the `calibrations`, `validations` and `recording_config` that are used in `io.py` to be stored in their respective `gaze.` instance attributes. Then, `total_recording_duration_ms` is still left to be removed, still used in `multipleye-preprocessing`. But no code inside `src/pymovements` ever reads `gaze._metadata`.

## Implemented changes
- [x] Remove `_calculate_data_loss_ratio()` function
- [x] Remove data loss computation from `gaze/_utils/_parsing_eyelink.py`: `num_expected_samples`, `num_valid_samples`, `num_blink_samples`, `blinking` flag, `blink_intervals` tracking, and the full `if/else` block that assigned `data_loss_ratio`/`data_loss_ratio_blinks` to metadata
- [x] Retain `total_recording_duration_ms` in metadata (still used downstream)
- [x] Remove data loss test cases from `tests/unit/gaze/_utils/_parsing_eyelink_test.py`
- [x] Remove `tests/unit/gaze/_utils/_parsing_test.py`
- [x] Remove data loss assertions from `tests/unit/gaze/io/asc_test.py`

## How Has This Been Tested?
- [x] test `parsing_eyelink`, `asc_test`, `data_loss`
- [x] Validation in https://github.com/MultiplEYE-COST/multipleye-preprocessing/pull/140

## Type of change
- Removal of deprecated functionality (replaced by measure-based equivalents)

This PR itself is not to be considered deprecation, but breaking change, right?

#### Context
Resolves #1584

#### Depends on
- https://github.com/MultiplEYE-COST/multipleye-preprocessing/pull/140

#### related issues:
- #893
- #1082
- #1398
- #1420 

## Checklist:


dkrako (approved):
@cbueth Thank you so much for the thorough analysis in #1584 and the clean removal. The direction is right: the fields were EyeLink-only, had the confirmed negative-ratio bug, and nothing inside `src/pymovements` reads them. One substantive finding, plus some cleanup. I added commits to this branch addressing both, see below.

## Blink replacement is not equivalent for binocular data

The removed metadata code merged overlapping blink intervals before counting blink samples. `events2timeratio` sums `(offset - onset + dt)` per event without merging. Binocular recordings emit separate left-eye and right-eye `blink_eyelink` events that typically overlap, so the overlap is counted twice and the ratio can exceed 1.

Reproduced with the scenario from the removed `test_parse_eyelink_binocular_missing_samples_data_loss` (left blink fully containing a right blink over a 105 ms span at 1000 Hz):

| | blink ratio |
|---|---|
| removed `data_loss_ratio_blinks` (merged intervals) | 97 / 105 ≈ 0.924 |
| `measure_events_ratio('blink_eyelink', sampling_rate=1000)` | 177 / 105 ≈ 1.686 |

So the equivalence demonstrated in #1584 holds for monocular data only. I don't consider this blocking for this PR; it is tracked in #1661. I added tests pinning this down on both levels (unit `events2timeratio` in `segmentation_test.py` and functional `from_asc` + `measure_events_ratio` in `tests/functional/data_loss_measures_test.py`), each parametrized with a passing case asserting the current double-counted value and a `strict` xfail case asserting the correct merged value. The PR resolving #1661 then drops the current-behavior cases and the xfail marks.

## Dead-code cleanup (commit e73542d)

- `recording_config[-1].get('sampling_rate')` was left behind as a no-op statement. Removing it also fixes a latent crash: ASC files with START/END but no RECCFG raised an uncaught `IndexError` on that line.
- The corrupted-file warning still referred to the removed data-loss metrics; it now references the total recording duration.
- `_config_inconsistent()` had no production callers left (it only decided whether to suppress the also-removed generic sampling-rate warning) and was only kept alive by its own unit test; both removed. The user-facing inconsistency warnings all live in `_check_reccfg_key` / `_check_samples_config_key` and are untouched.

## Remaining notes

- To answer the question in the description: yes, this is a removal, not a deprecation. Since `gaze._metadata` is private I'm fine with skipping a deprecation cycle, but I'd swap the `deprecation` label (nothing is deprecated, and it would land under "Deprecations" in the release notes) for `breaking`.
- Two known remaining differences to the old metadata: the measure's denominator is the observed sample span rather than the START/STOP block duration (loss at block edges is underestimated), and varying sampling rates across blocks cannot be expressed in a single `sampling_rate=` argument. Both are addressed by the `Session`/`Recording` data model proposed in #1663

PR #1611, opened 2026-07-17
feat!: unnest columns in tsplot

### LINKED ISSUE #450: Add support for list-columns in tsplot
## Description of the problem
Currently the signal columns need to be exploded before using this plotting method.
It would be much nicer to pass the list-columns.

## Description of a solution
In a first iteration we could just stack the channels such that each component will get a column, each channel a row.

As a follow-up, we could add an additional parameter to specify how the time series signals should be grouped.

- [ ] Add each component as a subplot, each row for channel, each column for component

saeub:
## 💥 Breaking Changes

- `share_y` now defaults to `False` (was `True`). Each channel subplot scales its y-axis to its own data. Pass `share_y=True` to restore the previous rendering.
- `zero_centered_yaxis` now defaults to `False` (was `True`). Y-limits follow the data range instead of being symmetric around zero. Pass `zero_centered_yaxis=True` to restore the previous rendering.
- With no `channels` given, `tsplot` now plots all numeric columns, including automatically unnested list columns (e.g. `pixel` as `pixel_x` and `pixel_y`). Previously, list columns were excluded and had to be unnested manually, so default plots will show more channels than before.
- Y-limits are now computed from finite values only and use range-based padding. Plots of channels with a positive minimum (e.g. pixel coordinates) or containing nan/inf will render with different limits than before.

No data or return values change; existing code keeps running, but rendered figures will look different unless the old defaults are passed explicitly.

## Description

Minimal implementation for #450. Automatically unnests list columns and plots each component as a separate channel.

**Question to reviewer:** The default behavior of centering and sharing the y axis (`share_y=True` and `zero_centered_yaxis=True`) doesn't seem very user-friendly. If the default behavior of `tsplot(gaze)` is to plot all numerical columns, they are very likely on completely different scales, so maybe setting the default values to `False` makes more sense?

## Implemented changes

- [x] Unnest columns before plotting
- [x] Include list columns by default when no `channels` are specified
- [x] Make `_unnest_list_column` not private (and move to a utils module?)
- [ ] Adapt tutorial (https://pymovements.readthedocs.io/en/stable/tutorials/plotting.html#plot-raw-samples, unnesting no longer necessary)
- [ ] Adapt tests (unnesting no longer necessary)

## How Has This Been Tested?

- [ ] TODO

## Type of change

- New functionality

## Context

Resolves #450

## Checklist:


dkrako (commented):
Thanks @saeub for this contribution!

I finalized the PR with a few commits on top of yours:

- Fixed the channel handling so selections without list columns work (a plain `str` or a list of non-nested columns previously crashed, which was what broke CI). Selection now happens first, unnesting only when nested columns are present.
- Regarding your question about the defaults: I agree, `share_y` and `zero_centered_yaxis` now default to `False`. This changes how existing `tsplot` calls render (cosmetic only, same data), so the PR is labeled as a breaking change. Detailed reasoning below.
- Extracted the y-limit computation into a helper that ignores non-finite values and uses range-based padding, so channels with a positive minimum (e.g. pixel coordinates) are no longer clipped at the bottom.
- Added tests (default channel selection with automatic unnesting, single `str` channel, hand-computed per-channel y-limits, nan/inf handling) and simplified the plotting tutorials to rely on the new defaults and automatic unnesting.

One thing that came up during review is deferred to a follow-up issue:

- #1683 allows plotting a single component of a nested column (e.g. only the x component of `pixel`)

**Why the defaults changed to <code>share_y=False</code> and <code>zero_centered_yaxis=False</code>**

With `channels=None` now plotting all numeric columns, the default channel set is heterogeneous in unit, scale, and sign (milliseconds, pixels, degrees, deg/s). A shared y-axis gets dictated by the largest-magnitude channel and flattens everything else to a near-constant line. Zero-centering only suits signed, roughly zero-symmetric quantities like velocity, while strictly positive channels like pixel coordinates waste half the panel on an empty negative range. The old combination was the worst case: every subplot got `±max(|all values|)`, so a single large magnitude anywhere dictated all axes.

Per-axes autoscaling also matches what matplotlib users expect, and users plotting homogeneous channels (e.g. only the pixel components) can still opt in with one keyword since they know their channels are comparable.

Also good catch on the pre-existing `share_y=False` bug where per-channel limits were computed from the whole array.

dkrako (approved):

saeub:
Wow, thanks @dkrako! Sorry, I didn't mean to leave it to you to finish this. :sweat_smile:

dkrako:
Ah don't worry, my changes were minor and it was quicker to just do them and merge. The number of open PRs stresses me out.

Issue #1488, opened 2026-02-17
drop data outside of trials


dkrako:
## Description of the problem

Often times there is also data recorded outside of trial boundaries (i.e. shortly before, after or in-between trials).

Usually this data is irrelevant and users will probably like to drop these sample rows completely.

## Description of a solution

As we have our `Gaze.trial_columns` available it should be easy to drop all rows from `Gaze.samples` if any value of the columns is `null`.

PR #1614, opened 2026-07-17
fix: Add `nulls_equal=True` to table joins

### LINKED ISSUE #1613: `null` values in trial columns lead to unexpected behavior
## Current Behavior

My gaze data contains a "page" column, which can be `null` (if the trial does not have multiple pages). I want to add this column as a `trial_column`, but when I do, strange things happen:

```py
>>> gaze.pix2deg()
>>> gaze.detect("idt")
>>> gaze.compute_event_properties(("location", {"position_column": "pixel"}))
shape: (41, 7)
┌─────────────────────────┬──────┬──────────┬─────────┬─────────┬──────────┬───────────┐
│ stage                   ┆ page ┆ name     ┆ onset   ┆ offset  ┆ duration ┆ location  │
│ ---                     ┆ ---  ┆ ---      ┆ ---     ┆ ---     ┆ ---      ┆ ---       │
│ str                     ┆ str  ┆ str      ┆ i64     ┆ i64     ┆ i64      ┆ list[f64] │
╞═════════════════════════╪══════╪══════════╪═════════╪═════════╪══════════╪═══════════╡
│ p1.anno.LiveNLI__75084c ┆ null ┆ fixation ┆ 1321662 ┆ 1321817 ┆ 155      ┆ null      │
│ p1.anno.LiveNLI__75084c ┆ null ┆ fixation ┆ 1321858 ┆ 1321991 ┆ 133      ┆ null      │
│ p1.anno.LiveNLI__75084c ┆ null ┆ fixation ┆ 1322005 ┆ 1322200 ┆ 195      ┆ null      │
│ p1.anno.LiveNLI__75084c ┆ null ┆ fixation ┆ 1322201 ┆ 1322413 ┆ 212      ┆ null      │
│ p1.anno.LiveNLI__75084c ┆ null ┆ fixation ┆ 1322542 ┆ 1322677 ┆ 135      ┆ null      │
│ …                       ┆ …    ┆ …        ┆ …       ┆ …       ┆ …        ┆ …         │
│ p1.anno.LiveNLI__75084c ┆ null ┆ fixation ┆ 1330125 ┆ 1330331 ┆ 206      ┆ null      │
│ p1.anno.LiveNLI__75084c ┆ null ┆ fixation ┆ 1330361 ┆ 1330614 ┆ 253      ┆ null      │
│ p1.anno.LiveNLI__75084c ┆ null ┆ fixation ┆ 1330629 ┆ 1330915 ┆ 286      ┆ null      │
│ p1.anno.LiveNLI__75084c ┆ null ┆ fixation ┆ 1331043 ┆ 1331256 ┆ 213      ┆ null      │
│ p1.anno.LiveNLI__75084c ┆ null ┆ fixation ┆ 1331308 ┆ 1331435 ┆ 127      ┆ null      │
└─────────────────────────┴──────┴──────────┴─────────┴─────────┴──────────┴───────────┘
```

Note how the `location` column is `null`.

I traced the problem back to this line: https://github.com/pymovements/pymovements/blob/c876090a613ee9e55949303ebf15d615881ff026/src/pymovements/events/events.py#L244

Turns out that when doing joins, polars ignores rows with `null` values by default. Adding `nulls_equal=True` to the `join()` call fixes the issue.

Based on a quick search, this could also be a problem here:
https://github.com/pymovements/pymovements/blob/c876090a613ee9e55949303ebf15d615881ff026/src/pymovements/transforms/segmentation.py#L412-L416
There might be more.

## Expected Behavior

Computing properties should work even when there are `null` values in some trial columns.

## Checklist

saeub:
## Description

See #1613.

## Implemented changes

- [x] Add `nulls_equal=True` to all polars `join` calls I could find

## How Has This Been Tested?

Didn't add a test, but checked that it works with some actual gaze data. Let me know if this requires a unit test.

## Type of change

- Bug fix

## Context

Resolves #1613

## Checklist:

### DIFF
--- a/src/pymovements/events/events.py
+++ b/src/pymovements/events/events.py
@@ -241,7 +241,7 @@ class Events:
         join_on: str | list[str]
             Columns to join event properties on.
         """
-        self.frame = self.frame.join(event_properties, on=join_on, how='left')
+        self.frame = self.frame.join(event_properties, on=join_on, how='left', nulls_equal=True)
 
     def drop(
             self,
--- a/src/pymovements/measure/reading/measures.py
+++ b/src/pymovements/measure/reading/measures.py
@@ -307,7 +307,7 @@ def saccade_length_out(fixations: pl.DataFrame) -> pl.DataFrame:
     )
 
     last_fixations = (
-        fixations.join(first_run, on=['trial', 'page', 'word_idx'])
+        fixations.join(first_run, on=['trial', 'page', 'word_idx'], nulls_equal=True)
         .filter(pl.col('run_id') == pl.col('first_run'))
         .group_by(['trial', 'page', 'word_idx'])
         .agg(pl.all().sort_by('onset').last())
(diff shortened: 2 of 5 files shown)


cbueth (changes requested):
The two fixes are correct and directly address the bug in #1613.

I do find one relevant `.join()` call: `measure/reading/measures.py` with inner join in `saccade_length_out()` on ['trial', 'page', 'word_idx']. Both sides can have null page, so fixations with null are silently dropped.

Would you add that one?

As you mentioned, this fix should also unblock the feature request in #1488, since joins now work correctly when trial columns contain nulls. This follow-up can be worked on then. @dkrako

saeub:
@cbueth Thanks, will do :+1:

What about these joins?
https://github.com/pymovements/pymovements/blob/abcffd246f50b8f995d221f6a9165a8a43daffd9/src/pymovements/measure/reading/measures.py#L448-L460

cbueth (approved):
> What about these joins?

These joins aren't strictly affected because left side is words where page is never `null`, but `nulls_equal=True` there doesn't hurt, so I suggest acceptance.

saeub:
> left side is words where page is never `null`

In my case, the page of some words *would* actually be `null`. Because some of my trials are single-page, and `page` is None.

saeub:
@dkrako wanted me to add a test, so I converted this to a draft.

cbueth:
> > left side is words where page is never `null`
> 
> In my case, the page of some words _would_ actually be `null`. Because some of my trials are single-page, and `page` is None.

Then all are needed. The changed ones should be all I found.

cbueth (approved):

PR #1619, opened 2026-07-20
docs: fix argument name error in how to download dataset example

dkrako:
There was an error in the how-to-download example code in the dataset template. There is now `name` argument in `Dataset.__init__()`.


prassepaul (approved):
thx.

PR #1621, opened 2026-07-21
chore!: remove deprecated DatasetDefinitions attributes as scheduled

dkrako:
## :boom: Breaking Change

The deprecated `DatasetDefinition` attributes `filename_format`, `filename_format_overrides` and `has_files` are removed as scheduled (target version v0.28.0). Please use `ResourceDefinition.filename_pattern`, `ResourceDefinition.filename_pattern_schema_overrides` and `ResourceDefinitions.has_content()` instead.

## Implemented changes

- [x] remove `DatasetDefinition.filename_format`
- [x] remove `DatasetDefinition.filename_format_schema_overrides`
- [x] remove `DatasetDefinitions.has_files()`


src/pymovements/dataset/dataset_definition.py
cbueth: Is there a reason why we should import and keep `_HasResourcesIndexer` here?

cbueth (changes requested):
Thanks for the changes.

What I still found was 

https://github.com/pymovements/pymovements/blob/b412952dbce1cb411ef88f182b0639d08b20c15a/src/pymovements/dataset/resources.py#L473

`_HasResourcesIndexer` should be updated or the class removed if it's no longer needed elsewhere (it is still tested).

src/pymovements/dataset/dataset_definition.py
dkrako: good catch, I will remove that one

cbueth (approved):

PR #1623, opened 2026-07-21
feat: add `drop_nulls()` to Dataset, Gaze, and Events

### LINKED ISSUE #1488: drop data outside of trials

saeub:
## Description

Functionality to remove samples and/or events with null values in certain columns. Useful for removing out-of-trial events or lost data samples. See #1488.

## Implemented changes

- [x] Add `Dataset.drop_nulls()`
- [x] Add `Gaze.drop_nulls()`
- [x] Add `Events.drop_nulls()`

## How Has This Been Tested?

- [x] Unit test for `Dataset.drop_nulls()`
- [x] Unit test for `Gaze.drop_nulls()`
- [x] Unit test for `Events.drop_nulls()`

## Type of change

- New functionality

## Context

Resolves #1488

## Checklist:


src/pymovements/events/events.py
```diff
@@ -617,6 +617,20 @@ def split(
             for frame in event_dfs
         ]
 
+    def drop_nulls(self, columns: list[str], criterion: Literal['all', 'any'] = 'all') -> None:
```
dkrako: The default value for `criterion` is inverted to the way how polars/pandas do it: they drop rows if *any* of the columns are null. I would prefer to be consistent with that.

Further we might also allow for `columns: list[str] | None = None`: in case no columns are given the method should be applied to all columns.

Regarding the naming it might make more sense to be in line with polars/pandas to get "free" recognition for users accustomed with that packages. Although the `criterion` name is very descriptive, pandas has a `how` argument that users might more easily recognize. Same with `columns`: polars/pandas use a `subset` argument.

https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.drop_nulls.html
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.dropna.html

tests/unit/gaze/gaze_test.py
dkrako, line 1: please also add tests like these to `events_test.py`

dkrako (changes requested):
Thank you so much for these changes! I have some comments regarding the signature. I think it would be best to be consistent with polars/pandas. Moreover some tests for the `Events` method specifically would be great!

src/pymovements/events/events.py
saeub: > The default value for criterion is inverted to the way how polars/pandas do it

True... My reasoning was that if we want to remove out-of-trial samples, we probably want to use the `all` criterion (related to #1614). But I guess users will notice this easily.

Agree with your suggestions, will change it now.

tests/unit/gaze/gaze_test.py
saeub: done

tests/unit/dataset/dataset_test.py
```diff
@@ -2266,6 +2266,59 @@ def test_compute_event_properties_alias(gaze_dataset_configuration, property_kwa
...
+                'columns': ['onset', 'offset'],
+                'samples': False,
+            },
+            0,
+            1,
+            id='events_only',
+        ),
+    ],
+)
+def test_drop_nulls(
```
dkrako, line 2307: this seems to fail currently

saeub: fixed it

dkrako (approved):
Thank you so much for this contribution @saeub! The API follows the polars/pandas `subset`/`how` naming nicely, and the tests are well parametrized with hand-verifiable literal expectations, including the asserted missing-columns error message. Solid addition that covers both use cases from #1488.

Two semantic gaps surfaced during final review, and I fixed them directly in my finalization commits so we can get this merged:

1. `drop_nulls()` with all-default arguments always raised `ValueError`: `subset=None` expanded to all sample columns, and the `events=True` default then required those columns to exist in the events frame. Settled semantics: with `subset=None` each frame is now checked on its own columns (samples frame on sample columns, events frame on event columns). An explicit `subset` still has to exist in every targeted frame, so typos stay loud.
2. Nested component columns (`pixel`, `position`) were checked with a plain `is_null()`, which is `False` for entries like `[null, 1.0]`, so dropping lost samples silently removed nothing. Rows now count as null component-wise: under `how='any'` if any component is null, under `how='all'` only if all components are null.

Summary of my finalization commits, for transparency:

- fix drop_nulls defaults, nested column null handling and validation: the two semantic fixes above via a new shared helper `row_is_null` in `src/pymovements/_utils/_nulls.py`, runtime validation of `how` (a typo like `'anny'` previously flipped silently to `'all'` semantics), error messages that name the actually missing columns and the frame they are missing from, plus tests for all new paths.
- return self from Dataset.drop_nulls and cover default path: `Dataset.drop_nulls` now returns `self` for method cascading like its neighbors, and the dataset test parametrization covers the all-default `samples=True, events=True` path.
- treat empty subset as no-op in drop_nulls: `drop_nulls(subset=[])` previously raised an opaque polars `ComputeError` from an empty fold; it now keeps all rows, matching `polars.DataFrame.drop_nulls(subset=[])`, with tests pinning the no-op and that an invalid `how` still raises even with an empty subset.
- use uniform event names in drop_nulls example / show any and all semantics in drop_nulls examples: executable doctest examples on `Events.drop_nulls` and `Gaze.drop_nulls` that demonstrate the `how='any'` vs `how='all'` difference (including nested pixel components) and the default per-frame cleaning of samples and events.

Thanks again @saeub, merging once CI is green on the finalized branch!

PR #1625, opened 2026-07-21
feat: add fixation drift correction

### LINKED ISSUE #1615: fixation correction
## Description of the problem
Eye-tracking data, especially during reading tasks, often suffers from vertical drift. This drift causes fixations to appear off-line from the actual text stimuli, making it difficult to accurately assign fixations to specific lines or words.

Currently, `pymovements` lacks a dedicated suite of algorithms to automatically correct for this vertical drift.

## Description of a solution
Implement the fixation correction algorithms described in Carr (2022).
  - `attach`: Attaches fixations to the nearest line.
  - `chain`: Groups fixations into chains and aligns chains to lines.
  - `cluster`: Uses clustering to identify lines.
  - `merge`: Merges fixations into lines based on vertical proximity.
  - `regress`: Uses regression to align fixations to lines.
  - `slice`: Slices the fixation sequence into lines.
  - `split`: Splits the fixation sequence into lines.
  - `stretch`: Stretches or shrinks the fixation sequence to match line positions.
  - `warp`: Dynamic time warping to align fixations to lines. 

## Implementation details
The basic logic should be implemented in `pymovements.transforms.fixation_correction`, ideally only relying on other transforms and polars functions (otherwise we get a circular import). This should then be reused by `Events. Each algorithm should be implemented as a separate function within the `pymovements.transforms.fixation_correction` submodule (or a dedicated directory if the complexity warrants it).

The functions would then be used via an `Events` method like this:

```python
class Events:
   ...
   def correct_fixations(self, method: str, stimulus: TextStimulus, *):
```

It might be worth adding an `inplace: bool = True` argument to make comparisons before/after more convenient.

- Implementation of at least the core algorithms from Carr (2022).
- Integration into the `pymovements` API (e.g., as part of the `gaze` or `measure` modules).
- Comprehensive unit tests for each algorithm.
- Documentation including examples and references.
- High code coverage for new implementations.

## Context

This feature request is based on the following paper:
Carr, J. W. (2022). Algorithms for the automated correction of vertical drift in eye-tracking data. *Behavior Research Methods*, 54(1), 287-310.
https://link.springer.com/article/10.3758/s13428-021-01554-0

There is already some code available in a PR of our MultiplEYE partner project: https://github.com/MultiplEYE-COST/multipleye-preprocessing/pull/62/

bernhardangele:
## Description

This pull request introduces automated vertical drift correction for eye-tracking gaze data recorded during reading tasks, as described in Carr et al. (2022).                                                                              
This change implements 11 primary drift correction algorithms, an ensemble voting method ("Wisdom of the Crowd"), and high-level interface functions integrated into `pymovements.events.correction`.

No new third-party dependencies are required for this change (uses existing NumPy, Polars, SciPy, and Scikit-learn dependencies).

## Implemented changes

- [x] Implemented 11 automated vertical drift correction algorithms in `pymovements.events.correction.drift_algorithms`: `attach`, `chain`, `cluster`, `compare`, `merge`, `regress`, `segment`, `slice`, `split`, `stretch`, and `warp`.
- [x] Added the **Wisdom of the Crowd (WoC)** ensemble correction method (`wisdom_of_the_crowd`), combining predictions across candidate algorithms via majority voting per fixation, and set it as the default algorithm in `fixation_correction.py`.
- [x] Added Right-to-Left (RTL) reading script support (`text_right_to_left=True`) for return sweep detection in `merge`, `segment`, and `split`.
- [x] Added high-level user interface routines `create_corrected_fixations_locations()` and `add_corrected_fixations()` in `pymovements.events.correction.fixation_correction` with automatic `word_XY` calculation for DTW alignment.
- [x] Exported `correction` module in `pymovements.events.__init__.py`.
- [x] Added 20 comprehensive unit tests with 92% code coverage in `tests/unit/events/correction/`.
- [x] Added citations and documentation for all algorithms.

## How Has This Been Tested?

- [x] **Manual testing on example Eyelink data set**: Applied all algorithms to a real multiline-reading data set. All algorithms work and produce the expected results.
- [x] **Unit Tests**: Ran 20 new unit tests covering all 11 drift algorithms, DTW alignment, WoC ensemble voting, Right-to-Left reading parameters, and high-level fixation event dataframe routines (`pytest tests/unit/events/correction/`).
- [x] **Full Event Test Suite**: Ran the full events unit test suite (`pytest tests/unit/events/`) confirming all 375 tests pass.
- [x] **Code Coverage**: Verified 92% code coverage on `src/pymovements/events/correction/`.

## Type of change

- New functionality
- Documentation update

## Context

Resolves #1615

This PR builds on @kasprowski's PR from https://github.com/MultiplEYE-COST/multipleye-preprocessing/pull/62.

#### related issues:
- #1615 

#### required by:
- #

#### requires:
- [ ] #

#### conflicts with:
- #

## Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] Any dependent changes have been merged and published in downstream modules
- [X] I have checked my code and corrected any misspellings


src/pymovements/events/correction/drift_algorithms.py
```diff
@@ -0,0 +1,829 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+"""Algorithms for vertical drift correction in gaze data recorded during reading tasks.
+
+References & Citations:
```
dkrako, line 32: Awesome that you added all of these citations! However, instead please add these as bibtex entries to the bibliography here: https://github.com/pymovements/pymovements/blob/main/docs/source/bibliography.bib

You can then cite them using ``:cite:p:`KEY` `` (or ``:footcite:p:`KEY` `` in conjunction with a ```.. footbibliography::``` at the end of the docstring).

src/pymovements/events/correction/drift_algorithms.py
```diff
@@ -0,0 +1,829 @@
...
+######################################################################
+
+
+def attach(
+    fixation_XY: np.ndarray,
+    line_Y: np.ndarray,
+) -> np.ndarray:
+    """Attach each fixation to the vertically closest text line center.
+
+    Reference: Carr et al. (2022).
```
dkrako: please cite this via `:cite:`/`:footcite` instead

src/pymovements/events/correction/fixation_correction.py
```diff
@@ -0,0 +1,265 @@
...
+        fixationXY = fixations['location'].to_list()
+    elif 'location_x' in fixations.columns and 'location_y' in fixations.columns:
+        fixationXY = fixations.select(['location_x', 'location_y']).to_numpy().tolist()
+    else:
+        raise ValueError('No valid location coordinates found in events dataframe.')
+
+    fixationXY_arr = np.array(fixationXY)
+
+    if algorithm.lower() in {'wisdom_of_the_crowd', 'woc'}:
+        woc_algos = (
```
dkrako: Instead of hardcoding this it would be great to be able to pass a list of algorithms to be used for the ensemble. The default (`None`) should automatically apply the algorithms as in the referenced publication, but having the possiblity to pass a different list would be great. This would be especially useful if we will add more correction algorithms in the future.

src/pymovements/events/correction/fixation_correction.py
```diff
@@ -0,0 +1,265 @@
...
+        AOIs dataframe to calculate line positions from.
+
+    Returns
+    -------
+    list[float]
+        Line center y-coordinates of the text.
+    """
+    y_col = 'start_y' if 'start_y' in aois.columns else 'top_left_y'
+    heights = list(set(aois['height']))
+    height = heights[0]
```
dkrako: this requires AOIs to be of uniform height to work correctly. Please raise an exception in case the set size is larger than 1

src/pymovements/events/correction/fixation_correction.py
```diff
@@ -0,0 +1,265 @@
...
+    return [float(line + height / 2.0) for line in sorted(set(aois[y_col]))]
+
+
+def create_corrected_fixations_locations(
+    events: pl.DataFrame,
+    aois: pl.DataFrame,
+    algorithm: str = 'wisdom_of_the_crowd',
```
dkrako: besides allowing to pass a `str` here it would be nice to also support passing a function directly. However, the input arguments the current functions expect vary so it is probably not feasible in the current state. Not critical for merging.

src/pymovements/events/correction/fixation_correction.py
cbueth: Do they actually need to have the same height?

```python
    return (
        aois.filter(pl.col("line_idx").is_not_null())
        .group_by("line_idx")
        .agg(
            (pl.col("top_left_y") + pl.col("height") / 2)
            .mean()
            .alias("line_center")
        )
        .sort("line_idx")["line_center"]
        .to_list()
    )
```

this could work for this function, without them being all same height.

src/pymovements/events/correction/fixation_correction.py
cbueth: I do also think this might be a follow-up PR, unless you are motivated @bernhardangele :)

pyproject.toml
```diff
@@ -101,7 +101,7 @@ aggressive = 1
 skip = "./src/pymovements/_version.py,./.git/*,./.pytest_cache/*,tests/files/*,*.bib"
 count = true
 quiet-level = 3
-ignore-words-list = 'bund,InstRead'
+ignore-words-list = 'bund,InstRead,istance'
```
cbueth: If you move the citations to the bibtex file, you can remove `istance` from here again.

src/pymovements/events/correction/drift_algorithms.py
bernhardangele: Citations are now added to the bibliography. They use :cite:p: for now.

src/pymovements/events/correction/drift_algorithms.py
bernhardangele: Done (using :cite:)

dkrako:
The issue with the failing docs is fixed in #1593 but needs to be merged first.

src/pymovements/events/correction/fixation_correction.py
bernhardangele: This is now implemented in the function

pyproject.toml
bernhardangele: Done

src/pymovements/events/correction/fixation_correction.py
cbueth: Merci!

src/pymovements/events/correction/fixation_correction.py
bernhardangele: It would be very nice to have this, especially since the algorithms are all already defined as functions. I agree that it is better as a follow-up PR.

src/pymovements/events/correction/fixation_correction.py
bernhardangele: This is now implemented: There is a boolean parameter woc. If woc is True, it allows a list of algorithms to be passed for inclusion in the Wisdom of the Crowd method. If woc is False, a single algorithm name is expected as the parameter

dkrako: Nice! However, I think that the `woc` argument is redundant this way. If a list of algorithms is given, then do the woc ensemble.

bernhardangele: This is now implemented. The warp and compare algorithms are automatically excluded if word x coordinates are missing

dkrako:
Thank you so much for this substantial contribution @bernhardangele! Porting the full Carr et al. (2022) algorithm family plus the Wisdom of the Crowd ensemble into pymovements is a major addition.

**Summary of my finalization commits**: 
- rewrote the algorithms as polars expressions
- updated the API to `correct_fixations` / `correct_fixation_locations` with per-trial application over `trial_columns`
- added `Events.correct_fixations` and `Dataset.correct_fixations` with TextStimulus column mapping
- replaced `text_right_to_left` with `directionality` inferred from the stimulus writing system
- made the ensemble vote on line indices
- added an explicit `character_level` flag for character-level AOI frames
- added clear errors and warnings for malformed AOIs, short trials and unmatched fixation names
- added the reference-validated functional tests
- restructured the module into one file per algorithm under `pymovements.events.correction`
- parameterized the events location column (`location_column`)
- normalized list-form algorithm arguments so `['woc']` behaves identically to `'woc'`
 
As these changes are substantial (full rework of the module, new public API surface), could you please review them, @cbueth? I would prefer to keep the remaining changes to a minimum and address non-blocking comments in a follow-up PR. Since we just released v0.28.0 there'll be more than enough time before the next release.

Thanks again to you both, this is a feature I am really looking forward to shipping.

src/pymovements/events/correction/_utils.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+def map_locations(
+    location: str | pl.Expr,
+    core: Callable[[pl.Series], pl.Series],
+    alias: str,
+) -> pl.Expr:
+    """Return an expression piping the [x, y] locations through core as one batch.
+
+    The core treats its input as a single reading sequence, so the caller must evaluate
```
cbueth, line 88: The per-trial contract is documented, but the core receives the whole batch and treats it as one sequence. An explicit assertion or test that the expression is only evaluated per trial would be good. Silently passing multiple trials as one sequence would produce plausible but wrong corrections.

src/pymovements/events/correction/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+    'correct_fixation_locations',
+
+    'attach',
+    'chain',
+    'cluster',
+    'compare',
+    'merge',
+    'regress',
+    'segment',
+    'slice',
```
cbueth, line 70: This naming shadows https://docs.python.org/3/c-api/slice.html

Consider whether a clearer alias would read better at call sites.

tests/functional/drift_algorithms_reference_test.py
```diff
@@ -0,0 +1,333 @@
...
+
+import numpy as np
+import polars as pl
+import pytest
+
+import pymovements as pm
+from pymovements import WebSource
+from pymovements.events import correction
+
+pytestmark = pytest.mark.network
```
cbueth, line 68: Good that it is gated, MD5-pinned and fails rather than skips. It does download and execute third-party reference code. Just confirm it runs in the required CI job so the validation cannot disappear, and the two pinned sources are ones we want to trust long term.
I think this is a good approach :)

cbueth (approved):
Thank you for reworking and wrapping the drift correction, so we can finally ship it.
I added just minor comments.

dkrako:
Thank you so much @cbueth for the thorough review and approval, and thanks again @bernhardangele for this substantial contribution!

I will address the remaining review comment on adding an explicit assertion/test that the correction expression is only evaluated per trial in a follow-up PR.

Regarding the `slice` naming: I would leave it as is. It shadows the builtin, which is not ideal, but it is the name used by Carr et al. in the original implementation, and keeping the algorithm names consistent is more important I think. But I totally get the point and I stumbled upon this too during implementation.

The reference validation test already runs in this PR's CI, so nothing to be done there.

With that, this is ready to merge. Since this implementation builds on @kasprowski's original work, I will credit him as co-author in the merge commit.

PR #1626, opened 2026-07-21
feat: add metadata field to stimulus classes

saeub:
## Description

Add a metadata dictionary to `TextStimulus` and `ImageStimulus` containing metadata from filename patterns.

## How Has This Been Tested?

- [x] Unit test: `tests/unit/dataset/dataset_files_test.py::test_load_stimulus_file_has_correct_metadata`
- [x] Manually tested with GGTG dataset definition

## Type of change

- New functionality

## Checklist:


src/pymovements/stimulus/image.py
```diff
@@ -30,18 +31,21 @@
...
     images: list[Path]
         Image stimulus list.
+    metadata: dict[str, Any] | None
+        Dictionary containing additional metadata. (default: None)
     """
 
-    def __init__(self, images: list[Path]) -> None:
+    def __init__(self, images: list[Path], metadata: dict[str, Any] | None = None) -> None:
         self.images = images
+        self.metadata = metadata or {}
```
dkrako: this resolves to a boolean value

dkrako: thx for clarification, I didn't know about that syntactic sugar

cbueth: Did I hear sugar? 😋

dkrako (changes requested):
Great, thanks a lot for the changes! Can you also add a test each to `stimulus/image_test.py` and `stimulus/text_test.py`? Maybe parametrize to test an empty dict and a key-value pair? That would be great!

src/pymovements/stimulus/image.py
dkrako: so after checking in tests that the metadata field shares the reference to the passed metadata field I noticed that this fails if an empty dict is passed (it's pointing to another empty dict then). It's very minor but I think we should be consistent here.

Therefore I changed it to ```metadata if metadata is not None else {}```.

@saeub Thanks anyway for exposing this functionality to me!

dkrako (approved):
Thanks again for your changes! I adjusted some of the tests (there were some tests that used absolute paths which break my pycharm workflow so I adjusted these too) and also checked that the metadata dict is guaranteed to share the reference to the passed object.

PR #1633, opened 2026-07-21
feat: add non-AOI fixation ratio reading measures

MirceaMM:
feat: add fixation outside AOI percentage calculations and corresponding tests

## Description

This PR introduces two new summary reading measures:
1. `fixation_outside_aoi_percentage_by_count`: Calculates the proportion of fixations that fall outside any Area of Interest (AOI) relative to the total number of fixations per page (`FOAC`).
2. `fixation_outside_aoi_percentage_by_duration`: Calculates the proportion of total fixation duration spent outside any AOI relative to the total fixation duration per page (`FOAD`).

This can be useful as a data quality measure to determine whether fixation correction is needed, or how much the data improved after applying it.

## Implemented changes

- Added `fixation_outside_aoi_percentage_by_count(fixations: pl.DataFrame)` summary reading measure function in `src/pymovements/measure/reading/measures.py`.
- Added `fixation_outside_aoi_percentage_by_duration(fixations: pl.DataFrame)` summary reading measure function in `src/pymovements/measure/reading/measures.py`.
- Added 7 unit tests in `tests/unit/measure/reading/reading_measures_test.py` testing all-inside fixations, mixed fixations, multi-trial, multi-page, and zero-duration edge cases.

## How Has This Been Tested?

Ran unit tests covering both functions across various DataFrame configurations:
- [x] Tested `fixation_outside_aoi_percentage_by_count` with all fixations inside AOI (`word_idx` populated), mixed fixations (`word_idx=None` for half of them), and grouping by trial for multi-trial Event frames.
- [x] Tested `fixation_outside_aoi_percentage_by_duration` with duration-weighted ratios, and grouping by trial for multi-trial Event frames.


src/pymovements/measure/reading/measures.py
```diff
@@ -396,6 +396,71 @@ def per_group(df: pl.DataFrame) -> pl.DataFrame:
...
+    return (
+        fixations.group_by(['trial', 'page'])
+        .agg(
+            [
+                (pl.col('word_idx').is_null()).sum().alias('fix_outside_aoi'),
+                pl.len().alias('total_fixations'),
+            ],
+        )
+        .with_columns(
+            (pl.col('fix_outside_aoi') / pl.col('total_fixations')).alias('FOAC'),
```
cbueth: The zero denominator needs to be handled. If a group has no fixations or `total_duration == 0`, the output will be NaN/inf. The by-duration variant is more susceptible since durations can legitimately be zero.

This might be solved with
```python
pl.when(pl.col('total_fixations') > 0)
.then(pl.col('fix_outside_aoi') / pl.col('total_fixations'))
.otherwise(None)
.alias('FOAC'),
```

Please add atest case for this.

src/pymovements/measure/reading/measures.py
```diff
@@ -396,6 +396,71 @@ def per_group(df: pl.DataFrame) -> pl.DataFrame:
...
+                pl.when(pl.col('word_idx').is_null())
+                .then(pl.col('duration'))
+                .otherwise(0)
+                .sum()
+                .alias('duration_outside_aoi'),
+                pl.col('duration').sum().alias('total_duration'),
+            ],
+        )
+        .with_columns(
+            (pl.col('duration_outside_aoi') / pl.col('total_duration')).alias('FOAD'),
```
cbueth: Same division by zero as above, could be replaced by 

```python
pl.when(pl.col('total_duration') > 0)
.then(pl.col('duration_outside_aoi') / pl.col('total_duration'))
.otherwise(None)
.alias('FOAD'),
```

And also get an own test case.

src/pymovements/measure/reading/measures.py
```diff
@@ -396,6 +396,71 @@ def per_group(df: pl.DataFrame) -> pl.DataFrame:
...
+        Fixation table containing at least ``trial``, ``page``,
+        and ``word_idx`` columns.
+
+    Returns
+    -------
+    pl.DataFrame
+        DataFrame with columns ``trial``, ``page``, and ``FOAC``.
+    """
+    return (
+        fixations.group_by(['trial', 'page'])
```
cbueth: Maybe we should use maintain order, in case elements are equal. This is not as important, but might be worth a thought.
```suggestion
        fixations.group_by(['trial', 'page'], maintain_order=True)
```

src/pymovements/measure/reading/measures.py
```diff
@@ -396,6 +396,71 @@ def per_group(df: pl.DataFrame) -> pl.DataFrame:
...
+        Fixation table containing at least ``trial``, ``page``,
+        ``word_idx``, and ``duration`` columns.
+
+    Returns
+    -------
+    pl.DataFrame
+        DataFrame with columns ``trial``, ``page``, and ``FOAD``.
+    """
+    return (
+        fixations.group_by(['trial', 'page'])
```
cbueth: Maybe we should use maintain order, in case elements are equal. This is not as important, but might be worth a thought.
```suggestion
        fixations.group_by(['trial', 'page'], maintain_order=True)
```

tests/unit/measure/reading/reading_measures_test.py
```diff
@@ -237,3 +239,102 @@ def test_regression_path_duration_no_first_pass():
...
+    df = pl.DataFrame({
+        'trial': ['1', '1', '2', '2', '2'],
+        'page': ['1', '1', '1', '1', '1'],
+        'word_idx': [0, None, None, None, 0],
+    }).cast({'word_idx': pl.Int64})
+    result = fixation_outside_aoi_percentage_by_count(df).sort('trial')
+    assert result.height == 2
+    assert result.filter(pl.col('trial') == '1')['FOAC'][0] == pytest.approx(0.5)
+    assert result.filter(pl.col('trial') == '2')['FOAC'][0] == pytest.approx(2 / 3)
+
```
cbueth, line 332: Add something like this:

```
def test_fixation_outside_aoi_percentage_by_count_all_outside():
    df = pl.DataFrame({
        'trial': ['1', '1'],
        'page': ['1', '1'],
        'word_idx': [None, None],
    }).cast({'word_idx': pl.Int64})
    result = fixation_outside_aoi_percentage_by_count(df)
    assert result['FOAC'][0] == pytest.approx(1.0)
```

Please use `@pytest.mark.parametrize` to save some repeated code to summarise this and earlier tests. Maybe each section.

tests/unit/measure/reading/reading_measures_test.py
```diff
@@ -237,3 +239,102 @@ def test_regression_path_duration_no_first_pass():
...
+    df = pl.DataFrame({
+        'trial': ['1', '1', '1', '1'],
+        'page': ['p1', 'p1', 'p2', 'p2'],
+        'word_idx': [0, None, None, None],
+        'duration': [100, 100, 200, 200],
+    }).cast({'word_idx': pl.Int64})
+    result = fixation_outside_aoi_percentage_by_duration(df).sort('page')
+    assert result.height == 2
+    assert result.filter(pl.col('page') == 'p1')['FOAD'][0] == pytest.approx(0.5)
+    assert result.filter(pl.col('page') == 'p2')['FOAD'][0] == pytest.approx(1.0)
```
cbueth: Also add a case here along the lines of:

```python
def test_fixation_outside_aoi_percentage_by_duration_all_outside():
    df = pl.DataFrame({
        'trial': ['1', '1'],
        'page': ['1', '1'],
        'word_idx': [None, None],
        'duration': [50, 150],
    }).cast({'word_idx': pl.Int64})
    result = fixation_outside_aoi_percentage_by_duration(df)
    assert result['FOAD'][0] == pytest.approx(1.0)
```

Please use parametrised tests.

---

Finally also test fixations with zero duration to ensure they don't break the division.

```python
def test_fixation_outside_aoi_percentage_by_duration_zero_total():
    df = pl.DataFrame({
        'trial': ['1', '1'],
        'page': ['1', '1'],
        'word_idx': [0, 1],
        'duration': [0, 0],
    })
    result = fixation_outside_aoi_percentage_by_duration(df)
    assert result['FOAD'][0] is None
```

cbueth (changes requested):
Thanks for the contribution, the implementation is solid overall, uses Polars correctly, has the page/trial semantics and the duration weighted calculation looks correct to me. I recommend fixing the division-by-zero and adding the missing edge case tests before merge.

The other two points have less of a priority.

You use the acronyms FOAC and FOAD. While I think I know where they come from, it would be great to have them defined in the docstrings. Consider adding to the Returns section of each:
```txt
DataFrame with columns ``trial``, ``page``, and ``FOAC``
(Fixation Outside AOI by Count: proportion of fixations
without a mapped word, 0.0 to 1.0).
```

The new measures should be imported and included in `__all__`.
```python
from pymovements.measure.reading.measures import fixation_outside_aoi_percentage_by_count
from pymovements.measure.reading.measures import fixation_outside_aoi_percentage_by_duration
```
In this file:
https://github.com/pymovements/pymovements/blob/1ba37a7f7f8f9c71ce416e0abb4b1bee64b85c52/src/pymovements/measure/reading/__init__.py#L48-L55

tests/unit/measure/reading/reading_measures_test.py
dkrako: for parametrization it's probably easiest to have an `expected_result` dataframe to check the complete output dataframe via `pytest.testing.assert_frame_equal()`. In case there are floating point rounding issues you can use the `abs_tol`/`rel_tol` argument there.

cbueth (approved):
Thank you for integrating the feedback. To me it looks good now.

@dkrako: If you think that it is prominently enough documented that these functions write the columns 'FOAC' and 'FOAD', it can be merged. Also if you think the names of the functions are fitting. I think it only comes down to usability now, the functionality looks great to me.

src/pymovements/measure/reading/measures.py
```diff
@@ -396,6 +396,81 @@ def per_group(df: pl.DataFrame) -> pl.DataFrame:
 
     return fixations.group_by('trial', 'page', maintain_order=True).map_groups(per_group)
 
+# ---------------------------
+# Summary measures
+# ---------------------------
+
+
+def fixation_outside_aoi_percentage_by_count(fixations: pl.DataFrame) -> pl.DataFrame:
+    """Compute the percentage of fixations outside any AOI, by count.
```
dkrako: The output of your functions are ratios not percentages. Please also update your docstrings and replace percentage with ratio.

src/pymovements/measure/reading/__init__.py
```diff
@@ -35,6 +35,8 @@
 from pymovements.measure.reading.measures import first_pass_fixation_count
 from pymovements.measure.reading.measures import first_pass_reading_time
 from pymovements.measure.reading.measures import first_reading_time
+from pymovements.measure.reading.measures import fixation_outside_aoi_percentage_by_count
```
dkrako: Please add your functions to the docs here: https://github.com/pymovements/pymovements/blob/main/docs/source/reference/measures.rst

dkrako (commented):
Thank you so much @MirceaMM and @stremoka for working though this contribution! I made some minor changes to finalize this:

- **Renamed the functions and output columns** to `non_aoi_fixation_count_ratio` (column `NAFCR`) and `non_aoi_fixation_duration_ratio` (column `NAFDR`). The previous acronyms `NAFC`/`NAFD` read a bit like raw counts next to columns such as `TFC`, although the values are ratios in 0 to 1. The new form puts the head noun last ("fixation count ratio"), makes the ratio visible in the acronym, and keeps name, expansion and column consistent.
- **Documented the outside-AOI convention**: a fixation counts as outside all AOIs if and only if its `word_idx` is null, which is what `map_to_aois` produces. The docstrings now warn that datasets encoding this with a sentinel such as `-1` must convert those values to null first, otherwise the ratio is silently underestimated. They also note that the input must contain only fixation events, since non-fixation rows carry null AOI columns after `map_to_aois` and would inflate the ratio.
- **Removed an unreachable guard** in the count variant: `group_by().agg()` never emits empty groups, so the `total_fixations > 0` branch was dead code. The corresponding guard in the duration variant is reachable (all-zero durations) and stays.
- **Added both of you to `CITATION.cff`**.

One heads-up for context: #1671 rewrites the reading measures as declarative expressions, and these two functions will be ported to that form there (they become one-liners). A configurable outside-AOI predicate for sentinel-based datasets is planned as part of that port, so it was intentionally left out here.

PR #1634, opened 2026-07-21
docs: fix deprecated directive formatting

### LINKED ISSUE #1404: broken formatting in deprecated arguments
Currently the notes on deprecated arguments are malformatted in our docs:

See this example on the [`DatasetDefinition`](https://pymovements.readthedocs.io/en/latest/reference/api/pymovements.DatasetDefinition.html) page:

Image

The problem is actually that the decorator is not used correctly, for example here:

https://github.com/pymovements/pymovements/blob/8e1da50f0a44b8c0f5dab31140058c9d8bcae8bb/src/pymovements/dataset/dataset_definition.py#L74-L77

The following is malformatted:

- a new line is missing before the deprecated instruction
- the indentation is missing for the additional text.

The above example wouzld be correctly rendered if it would be formatted like this:

```
    extract: dict[str, bool] | None
        Decide whether to extract the data. (default: None)

        .. deprecated:: v0.22.1
           This field will be removed in v0.27.0.
```

Sanjays2402:
## Description

Fixes malformed reStructuredText deprecation directives so their notices render correctly in the generated API documentation.

## Implemented changes

- [x] Separate directives from parameter descriptions with a blank line
- [x] Indent directive content consistently
- [x] Add a regression test for the affected docstrings

## How Has This Been Tested?

- [x] `python -m pytest -q --confcutdir=tests/unit tests/unit/test_docstring_formatting.py`
- [x] Ruff check and format check on the new test

## Type of change

- Bug fix
- Documentation update

## Context

Resolves #1404

#### related issues:
- N/A

#### required by:
- N/A

#### requires:
- [ ] N/A

#### conflicts with:
- N/A

## Checklist:


tests/unit/test_docstring_formatting.py
```diff
@@ -0,0 +1,50 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+# Copyright (c) 2024-2026, pymovements developers.
+# Distributed under the terms of the MIT License.
```
cbueth: The copyright string is duplicated, would you choose the one pre-commit added above?

tests/unit/test_docstring_formatting.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def test_deprecated_directives_are_separated_and_indented() -> None:
+    """Require valid reStructuredText formatting for deprecation directives."""
+    source_directory = Path(__file__).parents[2] / 'src' / 'pymovements'
+    source_files = (
+        source_directory / 'dataset' / 'dataset_definition.py',
+        source_directory / 'gaze' / 'gaze.py',
+        source_directory / 'gaze' / 'integration.py',
+    )
```
cbueth: I think this case is overspecified, as it only checks the files changed in this PR. The benefit is not carrying over to the full codebase.

cbueth (commented):
Thanks for the PR, @Sanjays2402! The docstring fixes look correct, blank lines and proper indentation will make the `.. deprecated::` directives render properly in Sphinx. However, the regression test only checks the three files that were fixed here, so it won't catch new malformed directives added elsewhere in the future. I'd suggest scanning recursively instead.

Also, the assertion message says "malformed deprecation directives". You could drop the word "deprecated" from the function name to match. Happy to discuss alternatives like using a Sphinx build check if that's simpler.

I think @dkrako should comment on the necessity of such programmative check. The corrections should be shipped for sure.

Also, see the failing pre-commit hooks. If you rin them before committing, they should make the needed linting.

tests/unit/test_docstring_formatting.py
Sanjays2402: Dropped it, kept the pre-commit one.

tests/unit/test_docstring_formatting.py
Sanjays2402: Agreed. It now walks all of src/pymovements with rglob instead of the three files I touched, and that immediately turned up four more malformed directives in dataset/resources.py (the url/filename/... properties), which I fixed in the same commit. Verified by checking out src at the pre-PR commit: the test reports every original violation, and it passes on 31a818a.

dkrako (approved):
Thank you so much for the fix @Sanjays2402, and congrats on your first contribution to pymovements!

The directive formatting fixes look good, and the regression test guarding against future malformed `.. deprecated::` directives is a nice addition. The rendered docs on the RTD preview confirm the notices now show up as proper deprecation admonitions.

I pushed a few finishing touches directly to your branch:

- moved the test to `tests/docs/docstring_formatting_test.py`: our pre-commit `name-tests-test` hook requires the `*_test.py` suffix  (this was the pre-commit.ci failure), and docs-related tests live in `tests/docs`
- added the missing module docstring (this was the pylint CI failure)
- located the sources via `request.config.rootpath` instead of `Path(__file__).parents[2]`, so the test won't break if the file is moved again
- allowed a directive directly after a section underline (valid rST), so the extra blank line in the `Notes` section of `dataset_definition.py` wasn't needed
- moved the `(default: None)` marker in `from_pandas` out of the directive content onto the description line

One side note: feel free to add yourself to the `CITATION.cff` file, so you get credited as a contributor.

PR #1635, opened 2026-07-21
feat: add recording duration measure

### LINKED ISSUE #1585: Recording duration sample measure
## Description of the problem
- Currently recording duraition is inferred at parsing time using
  https://github.com/pymovements/pymovements/blob/95a30e2cdfa21bf1e0b53fe955390137da70e63c/src/pymovements/gaze/_utils/_parsing_eyelink.py#L691-L702
- this can be removed, in favor of a sample measure

## Description of a solution
- sample measure can work aggregated on trial level, but also on Gaze level
- Remove the `pre_processed_metadata['total_recording_duration_ms']` field. (blocked by #1663)
- Optional counting of null/nan/infs for duration. Currently it is end_ts - start_ts and not offered.

- New recording duration sample measure
- A viable replacement of accessing `pre_processed_metadata['total_recording_duration_ms']`.

## Context

- #893

ColumbusLabs:
## Description

Add a registered sample-level duration measure that works for a complete `Gaze` or per trial. The measure uses the observed timestamp span (`max - min`) and replaces the EyeLink-specific `total_recording_duration_ms` metadata calculation.

The private metadata key is removed as requested by #1585. Downstream consumers that read it directly will need to migrate to `Gaze.measure_samples('duration')`. This is a breaking change: the new measure spans observed samples, while the removed field summed EyeLink `START`-to-`END` blocks, so gaps, blinks, or filtered samples can make the new value shorter.

## Implemented changes

- [x] Add and register `pymovements.measure.samples.duration`
- [x] Cover direct, whole-Gaze, per-trial, and `polars.Duration` behavior
- [x] Remove the legacy EyeLink duration metadata field
- [x] Preserve explicit warnings for unmatched EyeLink recording blocks
- [x] Publish the sample measure in the API documentation

## How Has This Been Tested?

- [x] Focused affected suite: 170 passed
- [x] Full warning-as-error documentation build: `tox -e docs -- -aE`
- [x] All changed-file pre-commit hooks, including mypy and pylint
- [x] `git diff --check`

A broader unit run was sampled without failures but was not completed; all affected measure, registry, Gaze, and EyeLink parser tests completed successfully.

## Type of change

- New functionality
- Breaking change
- Documentation update

## Context

Resolves partially #1585 (fully resolving the issue is blocked by #1663)

#### related issues:
- #1608 touches adjacent EyeLink parser code but deliberately retains this duration field for the present follow-up.
- #1637 changes the `time` column to `polars.Duration`; this PR includes direct duration-typed coverage without depending on that PR.

## Checklist:

- [x] I have added tests that prove the change is effective

### DIFF
--- a/src/pymovements/gaze/_utils/_parsing_eyelink.py
+++ b/src/pymovements/gaze/_utils/_parsing_eyelink.py
@@ -608,8 +608,9 @@ def parse_eyelink(
     calibrations = []
     recording_config: list[dict[str, Any]] = []
     samples_config: list[dict[str, Any]] = []
-
+    start_recording_timestamp: str | None = None
     total_recording_duration = 0.0
+
     is_binocular = False
 
     # Single pass: collect events, patterns, samples, samples config, and metadata
@@ -692,16 +693,17 @@ def parse_eyelink(
             start_recording_timestamp = match.groupdict()['timestamp']
 
         elif match := _match_regex(STOP_RECORDING_REGEX, line):
-            stop_recording_timestamp = match.groupdict()['timestamp']
-
-            try:
-                float(stop_recording_timestamp) - float(start_recording_timestamp)
-            except UnboundLocalError:
+            if start_recording_timestamp is None:
                 warnings.warn(
                     'END recording message without associated START recording message. '
                     f"File '{filepath}' may be corrupted. "
-                    'Total recording duration may be incorrect.',
+                    'Recording intervals may be incomplete.',
                 )
+            else:
+                total_recording_duration += (
+                    float(match.groupdict()['timestamp']) - float(start_recording_timestamp)
+                )
+            start_recording_timestamp = None
         if messages and (match := _match_regex(MSG_REGEX, line)):
             messages_list.append([match.groupdict()['timestamp'], match.groupdict()['content']])
 
@@ -838,6 +840,7 @@ def parse_eyelink(
     pre_processed_metadata['calibrations'] = calibrations
     pre_processed_metadata['validations'] = validations
     pre_processed_metadata['recording_config'] = recording_config
+    pre_processed_metadata['total_recording_duration_ms'] = total_recording_duration
 
     gaze_schema_overrides = {
         'time': pl.Float64,
--- a/src/pymovements/gaze/_utils/_parsing_eyelink.py
+++ b/src/pymovements/gaze/_utils/_parsing_eyelink.py
@@ -695,16 +695,13 @@ def parse_eyelink(
             stop_recording_timestamp = match.groupdict()['timestamp']
 
             try:
-                block_duration = float(stop_recording_timestamp) - float(start_recording_timestamp)
+                float(stop_recording_timestamp) - float(start_recording_timestamp)
             except UnboundLocalError:
                 warnings.warn(
                     'END recording message without associated START recording message. '
                     f"File '{filepath}' may be corrupted. "
                     'Total recording duration may be incorrect.',
                 )
-            else:  # this will only be executed if no exception was raised in the try block.
-                total_recording_duration += block_duration
-
         if messages and (match := _match_regex(MSG_REGEX, line)):
             messages_list.append([match.groupdict()['timestamp'], match.groupdict()['content']])
 
@@ -841,7 +838,6 @@ def parse_eyelink(
     pre_processed_metadata['calibrations'] = calibrations
     pre_processed_metadata['validations'] = validations
     pre_processed_metadata['recording_config'] = recording_config
-    pre_processed_metadata['total_recording_duration_ms'] = total_recording_duration
 
     gaze_schema_overrides = {
         'time': pl.Float64,
--- a/src/pymovements/gaze/_utils/_parsing_eyelink.py
+++ b/src/pymovements/gaze/_utils/_parsing_eyelink.py
@@ -581,7 +581,6 @@ def parse_eyelink(
     recording_config: list[dict[str, Any]] = []
     samples_config: list[dict[str, Any]] = []
 
-    total_recording_duration = 0.0
     # First pass: collect SAMPLES config for binocular detection only
     is_binocular = False
     for line in lines:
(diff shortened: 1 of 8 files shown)


cbueth:
Dear @ColumbusLabs, thanks for the PR and wanting to contribute. I have been looking into the timing last week and will review your suggestions in that context. Please give me a few days to come back to it.

src/pymovements/gaze/_utils/_parsing_eyelink.py
```diff
@@ -696,16 +695,13 @@ def parse_eyelink(
...
 
             try:
-                block_duration = float(stop_recording_timestamp) - float(start_recording_timestamp)
+                float(stop_recording_timestamp) - float(start_recording_timestamp)
             except UnboundLocalError:
                 warnings.warn(
                     'END recording message without associated START recording message. '
                     f"File '{filepath}' may be corrupted. "
                     'Total recording duration may be incorrect.',
                 )
```
cbueth: After removing `total_recording_duration_ms`, this `try` block now computes `float(stop) - float(start)` and discards the result. It only exists to raise `UnboundLocalError` when an END appears without a START. This is not needed.
Detecting a corrupted file could still be valuable: e.g. initialize `start_recording_timestamp = None` and warn on `None`. But as of now, computing a discarded expression is not a good approach.

The warning text "Total recording duration may be incorrect" references the field you just removed.

src/pymovements/gaze/_utils/_parsing_eyelink.py
```diff
@@ -842,7 +838,6 @@ def parse_eyelink(
     pre_processed_metadata['calibrations'] = calibrations
     pre_processed_metadata['validations'] = validations
     pre_processed_metadata['recording_config'] = recording_config
-    pre_processed_metadata['total_recording_duration_ms'] = total_recording_duration
 
     gaze_schema_overrides = {
```
cbueth, line 845: Removing `total_recording_duration_ms` from `pre_processed_metadata` is breaking for anyone reading it directly. We mark breaking changes with `feat!:`. consider whether this deserves a `!` or at least an explicit migration note.

src/pymovements/measure/samples/measures.py
```diff
@@ -198,6 +198,29 @@ def dispersion(
...
+    time_column: str
+        Name of the timestamp column. (default: 'time')
+
+    Returns
+    -------
+    pl.Expr
+        The duration of the sample group.
+    """
+    timestamps = pl.col(time_column)
+    return (timestamps.max() - timestamps.min()).alias('duration')
```
cbueth, line 227: Worth documenting somewhere: `max − min` of sample timestamps =/= the old START->END block span, especially when blinks/gaps shrink the observed sample span, so the "replacement" won't be value-identical for users migrating.

tests/unit/measure/samples/measures/duration_test.py
```diff
@@ -0,0 +1,67 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+"""Test duration sample measure."""
+import polars as pl
+import pytest
+from polars.testing import assert_frame_equal
+
+from pymovements.measure.samples import duration
+
+
```
cbueth, line 27: Just fyi, once we #1637 is merged, the `time` column becomes `Duration('us')`. A test case with a `polars.Duration` time column would be nice then. Also depends what get's merged first.

cbueth (commented):
Hi @ColumbusLabs, thank you for waiting on my review and thansk for the PR. Welcome to pymovements! The new `duration` measure is clean and good you included tests with proper coverage.

As always, there is some feedback:
- now that `total_recording_duration_ms` is gone, the `try` block computes `float(stop) - float(start)` and discards the result, existing only to raise `UnboundLocalError` for an END-without-START. I you really want to keep it, please detect that case more directly and with a fitting warning text.
- Removing `total_recording_duration_ms` from the metadata is a breaking change for direct. Since we mark breaking changes with `feat!:`, it may deserve a `!` or at least a note in the changelog.
- Minor: the new max−min sample span isn't numerically identical to the old block-span figure when blinks/gaps shorten the observed span. This is worth a sentence so migrating users aren't surprised.
- Optional: once #1637 (Duration time columns) merges, `time` will be `Duration('us')`. You could test for it (test with a Duration column), or wait for merge.

Thanks again!

src/pymovements/gaze/_utils/_parsing_eyelink.py
ColumbusLabs: Removed the discarded subtraction. The parser now tracks START explicitly with a None sentinel, warns on an unmatched END, and resets the state after every END so a second END cannot reuse an earlier START. Added regression coverage for that second-unmatched-END case.

src/pymovements/gaze/_utils/_parsing_eyelink.py
ColumbusLabs: Agreed. I changed the PR title to feat(measure)!: add recording duration measure and expanded the migration note in the description. It now calls out both the removed metadata key and the semantic difference users should expect.

src/pymovements/measure/samples/measures.py
ColumbusLabs: Documented this in the public duration docstring and the PR migration note: the measure spans observed sample timestamps and can be shorter than EyeLink START-to-END blocks when samples are absent because of gaps, blinks, or filtering.

tests/unit/measure/samples/measures/duration_test.py
ColumbusLabs: Added direct coverage with a polars.Duration microseconds time column. The result preserves Duration microseconds with the expected span, so this behavior is covered without waiting for #1637 to merge.

cbueth:
Thanks for the changes @ColumbusLabs. I suggest @dkrako looks over this and decides how to proceed.

dkrako (approved):
Thank you for this contribution @ColumbusLabs, and welcome to pymovements!

The new `duration` measure is a fine addition and the reworked unmatched-END handling is better than the old try/except. Good catch on the second END reusing a stale START, and including a regression test for that.

On the open question of `total_recording_duration_ms`: we keep the field for now, so your second commit had it right. The new measure spans the observed samples (`max - min`), while the metadata field sums the START to END recording blocks, so it is not a drop-in replacement, and the recording-span quantity would otherwise be lost. Its removal is deferred to the Session/Recording data model (#1663), which will carry per-block durations. Handling of missing samples within the span is tracked in #1694. I have updated the title accordingly: this PR is now a non-breaking feature addition.

I added a small commit on top with two docs touch-ups: the recording-duration test docstring now explains that the expected 1000.0 can only come from the START/END message timestamps (the file contains no samples), and the `duration` docstring now states that the result carries the unit of the time column.

Feel encouraged to add yourself to `CITATION.cff` so you are fully attributed in the next release. Thanks again!

PR #1636, opened 2026-07-22
chore!: remove deprecated ResourceDefinitions.from_dict as scheduled

dkrako:
## :boom: Breaking Change

The deprecated function `ResourceDefinitions.from_dict()` is removed as scheduled. Please use `ResourceDefinitions.from_dicts()` instead. This means passing a `dict[str, list[dict[str, Any]]]` as `resources` to `DatasetDefinition()` is not supported anymore. Please pass a `list[dict[str, Any]]` instead.

## Implemented changes

- [x] remove `ResourceDefinitions.from_dict()`
- [x] remove usage in `DatasetDefinition`
- [x] remove relevant tests from `dataset_definition_test.py` and `resources_test.py`


cbueth (approved):
The removal looks good to me, thanks. I also removed the `ResourcesLike` from nitpicks.

PR #1637, opened 2026-07-22
feat!: use polars `Duration` for time columns in Gaze and Events

### LINKED ISSUE #1111: use `polars.Duration` for time-related columns
## Description of the problem
Currently the time column of a `GazeDataFrame` will have the type `pl.Float64` after parsing a 2000Hz eyelink file with `from_asc()`.

This is inefficient, as float comparisons are much slower than integer comparisons.

## Description of a solution

The [`polars.Duration`](https://docs.pola.rs/api/python/stable/reference/api/polars.datatypes.Duration.html#polars.datatypes.Duration) dtype would be the perfect fit for time related columns, as it is actually just an integer.

Additionally, the column dtype would specify the time unit which would be very nice.

- [ ] change `time` column dtype in `GazeDataFrame.frame` to `polars.Duration`
- [ ] change `onset` column dtype in `EventDataFrame.frame` to `polars.Duration`
- [ ] change `offset` column dtype in `EventDataFrame.frame` to `polars.Duration`
- [ ] change `duration` column dtype in `EventDataFrame.frame` to `polars.Duration`
- [ ] add a benchmark test for evaluating the performance improvement

cbueth:
## Description

This PR implements the suggestion of #1111. Right now time data is in `Float64` columns: `time` in `Gaze`, `onset`/`offset`/`duration` in `Events`. The units are "milliseconds" by convention only, but nothing in the type system carries that information across boundaries. This PR switches those columns to `polars.Duration('ms')`, an `Int64`-backed type, so we can carry the time unit using the feature `polars` offers.

A central design question is: when converting `Float64` to  `Duration('ms')`, should we truncate or round? The answer might differ base on input unit:
- If we have millisecond without arithmetic: `.cast(pl.Duration('ms'))`leads to truncation. We would assume sub missilecond is not relevant, rounding would reduce precision.
- at second (s/us/step) input we get multiply/divide: `.round().cast(pl.Duration('ms'))` before casting, this rounds first. The motivation is floating-point arithmetic introduces errors like 1.001 x 1000 = 1000.999... So the `.round()` should kind of recover the intended value.

### The Events constructor
The current Events post-processing solution is a bit, lets say colorful. It uses `.cast(pl.Float64).round().cast(pl.Duration('ms'))`. The `.cast(pl.Float64)` is needed because `pl.read_csv` on empty files infers all columns as `pl.Utf8`, and `.round()` chokes on strings.

Alternatives for this would be:
1. Cast known time columns from String to Float64 in `load_event_files` already.
2. Override the schema from the start using `{'onset': pl.Float64, 'offset': pl.Float64, 'duration': pl.Float64}` in`pl.read_csv`.
3. Keep `.cast(pl.Float64)` no-op on Float64, a cheap widen on Int64, and needed for String.

### Questions for reviewers
1. The events constructor works but I find it a bit weird and long, but I think it is the best solution in our case. Otherwise, should we move the String to Float64 cast to `load_event_files` instead?
2. What do you think about the truncation / rounding choice. Truncation might be assymetric across paths, but simpler. Is the solution consistent enough?
3. Breaking(?): When you access individual Duration values from a polars DataFrame in Python, like `events.onset[0]` or iterate over the column, polars returns `datetime.timedelta(milliseconds=123)` instead of the raw number `123.0`. Any user code doing arithmetic or comparisons with those values now gets a timedelta, not a float.

## Implemented changes

- [x] `Gaze._convert_time_units`: `ms` -> `.cast(Duration('ms'))` and  `s`/`us`/`step` -> `.round().cast(Duration('ms'))`.
- [x] `Events.__init__`: numeric time cols are `.cast(Float64).round().cast(Duration('ms'))`
- [x] save path converts Duration -> numeric ms via `dt.total_milliseconds()` before write.
- [x] Internal: reading measures, resample, scanpath plot, segmentation updated to handle Duration
- [x] Added `to_duration()` helper to fixtures. I first added it to `tests/unit/helpers.py`, but hook `name-tests-test` did not like that.

This change would allow measures to support delivering a time unit.

## How Has This Been Tested?

- [x] Running all tests locally

## Type of change

- New functionality
- Breaking change

## Context

Resolves #1111

#### related issues:
- #1398 

#### required by:
- #

#### requires:
- [ ] #

#### conflicts with:
- #

## Checklist:


cbueth:
Example where it propagates. Before
```
    >>> from pymovements.gaze.io import from_asc
    >>> gaze = from_asc(file='tests/files/eyelink_monocular_example.asc')
    >>> gaze.samples
    shape: (16, 3)
    ┌─────────┬───────┬────────────────┐
    │ time    ┆ pupil ┆ pixel          │
    │ ---     ┆ ---   ┆ ---            │
    │ i64     ┆ f64   ┆ list[f64]      │
    ╞═════════╪═══════╪════════════════╡
    │ 2154556 ┆ 778.0 ┆ [138.1, 132.8] │
    │ 2154557 ┆ 778.0 ┆ [138.2, 132.7] │
    │ 2154560 ┆ 777.0 ┆ [137.9, 131.6] │
    │ 2154564 ┆ 778.0 ┆ [138.1, 131.0] │
    │ 2154596 ┆ 784.0 ┆ [139.6, 132.1] │
    │ …       ┆ …     ┆ …              │
    │ 2339246 ┆ 622.0 ┆ [629.9, 531.9] │
    │ 2339271 ┆ 617.0 ┆ [639.4, 531.9] │
    │ 2339272 ┆ 617.0 ┆ [639.0, 531.9] │
    │ 2339290 ┆ 618.0 ┆ [637.6, 531.4] │
    │ 2339291 ┆ 618.0 ┆ [637.3, 531.2] │
    └─────────┴───────┴────────────────┘
```

After:

```
    >>> from pymovements.gaze.io import from_asc
    >>> gaze = from_asc(file='tests/files/eyelink_monocular_example.asc')
    >>> gaze.samples
    shape: (16, 3)
    ┌───────────────┬───────┬────────────────┐
    │ time          ┆ pupil ┆ pixel          │
    │ ---           ┆ ---   ┆ ---            │
    │ duration[ms]  ┆ f64   ┆ list[f64]      │
    ╞═══════════════╪═══════╪════════════════╡
    │ 35m 54s 556ms ┆ 778.0 ┆ [138.1, 132.8] │
    │ 35m 54s 557ms ┆ 778.0 ┆ [138.2, 132.7] │
    │ 35m 54s 560ms ┆ 777.0 ┆ [137.9, 131.6] │
    │ 35m 54s 564ms ┆ 778.0 ┆ [138.1, 131.0] │
    │ 35m 54s 596ms ┆ 784.0 ┆ [139.6, 132.1] │
    │ …             ┆ …     ┆ …              │
    │ 38m 59s 246ms ┆ 622.0 ┆ [629.9, 531.9] │
    │ 38m 59s 271ms ┆ 617.0 ┆ [639.4, 531.9] │
    │ 38m 59s 272ms ┆ 617.0 ┆ [639.0, 531.9] │
    │ 38m 59s 290ms ┆ 618.0 ┆ [637.6, 531.4] │
    │ 38m 59s 291ms ┆ 618.0 ┆ [637.3, 531.2] │
    └───────────────┴───────┴────────────────┘

```

cbueth:
Do we have a test with 2000hz data?

dkrako (commented):
@cbueth thank you so much for getting this migration started. I reviewed the original PR and then reworked it fairly heavily, so this comment is my review of your submission followed by a summary of the changes I made on top, and ideas for remaining follow-ups.

## Review of the original PR

The overall approach was right: migrate `time`, `onset`, `offset` and `duration` to a polars temporal dtype with conversion shims at each numeric boundary, and you touched essentially all the relevant files. The reasons I reworked it rather than merging as-is:

- The central problem was storing `Duration('ms')`. Eye-tracking data may have sub-millisecond resolution (2 kHz EyeLink half-ms steps, microsecond timestamps), and the `ms`/`us`/`step` conversion branches truncated it. The `us` branch did `(col / 1000).round().cast(Duration('ms'))`, collapsing microsecond timestamps to whole milliseconds, and the `ms` branch cast float time straight to `Duration('ms')`, dropping the `.5` on every half-millisecond 2 kHz sample.
- The tests could not catch that loss: the `to_duration()` fixture and the "cast expected to Duration" assertions rounded the expected values the same way as the output, so truncation was compared against truncation.

## Summary of the changes I made

- Changed the storage dtype from `Duration('ms')` to `Duration('us')`, so sub-millisecond timing is preserved instead of truncated.
- Added a shared helper module `src/pymovements/_utils/_time.py` as the single source of truth for every conversion (`duration_to_ms`, `numeric_to_duration_us`, `normalize_duration_to_us`, `time_column_to_duration_us`, `timesteps_to_numpy`, `durations_to_ms`), replacing the roughly six ad-hoc idioms that had drifted apart.
- `Events.__init__`: added a `time_unit` parameter and made the constructor keyword-only. Numeric onset/offset/duration is interpreted by that unit, `timedelta`/`timedelta64` input is converted by its own physical unit, missing values (NaN) are stored as null rather than raising, infinite values are rejected with a clear error, and sub-microsecond Duration input is rounded rather than truncated. The unit is validated up front so a typo is rejected even for Duration input.
- `Gaze._convert_time_units`: rewritten to store the time column as `Duration('us')`, validate the unit up front, and normalize any Duration input to microseconds. The existing `'us'` unit now yields microsecond-precision Duration rather than being collapsed to milliseconds.
- `Gaze` `calibrations`, `validations` and `messages` now store their `time` column as `Duration('us')` too, consistent with the samples time column, via the shared `time_column_to_duration_us` helper (also reused by `Gaze._convert_time_units`); their CSV exports convert back to numeric milliseconds.
- `io.from_csv`: documented that a `Duration` schema override on the time column reinterprets the raw values in that Duration unit (the `time_unit` parameter itself already existed).
- Detectors `idt`/`ivt`/`microsaccades`/`fill`/`blink`/`out_of_screen` now accept Duration timesteps via `timesteps_to_numpy`, so they behave the same whether called directly or through `Gaze.detect`. The `ihmm` detector that landed on `main` in the meantime was integrated the same way (Duration timesteps accepted; doctests and tests use `Duration('us')`).
- Rewrote `fill` to mark samples by half-open interval membership `[onset, offset)` instead of exact-index arithmetic, which is robust to float timesteps and non-unit spacing.
- `merge_subsequent_close_events` compares gaps in fractional milliseconds so sub-ms gaps are not truncated.
- The event-vs-sample measures (`events2segmentation`, `events2timeratio`, `EventSamplesProcessor`) coerce both sides to numeric milliseconds, so a Duration/numeric dtype mismatch produces correct results instead of silently all-False.
- The data-quality checks, `quality.py`, `data_loss_histogram`, `tsplot`, reading measures, `resample`, `segmentation` and the CSV save paths all route through the shared helpers. CSV save narrows whole-millisecond Duration columns to `Int64` (reload-safe); feather preserves the Duration dtype.
- Along the way this fixed a few latent bugs that predated the migration: the sampling-rate consistency check was off by 1000x on a Duration time column, `resample` lost sub-millisecond precision, and `resample` left NaN in interpolated pixel/position list columns.
- Made `normalize_duration_to_us` overflow-safe (it no longer builds an intermediate nanosecond count that wraps past ~292 years) while preserving the rounding exactly.
- Broad test additions across the suite; doctests and tutorials updated to `duration[us]`

## Possible follow-up work (out of scope here, to track as separate issues)

- Expose public conversion helpers for users who manage their own dataframe export. `polars.write_csv` cannot serialize a `Duration` column, and while pymovements' own save paths convert internally, a user who calls `df.write_csv(...)` on e.g. `gaze.samples` or `events.frame` hits the error with no supported way to convert the time/onset/offset/duration columns to numeric milliseconds first. The conversion helpers currently live in `_utils._time` and are private.
- Let detector duration thresholds accept Duration values, not just int milliseconds (`minimum_duration` in idt/ivt/microsaccades/fill and `minimum_duration`/`maximum_duration`/`minimum_gap` in blink), with a clear error when `timesteps=None` (sample-based), so the "remember it is milliseconds" footgun is gone for thresholds too.
- Return time-valued measure results as Duration so results self-document their unit (event `duration` already is). Scope is only additive-duration measures such as the reading measures; rate/ratio/kinematic measures stay numeric because they divide by time. This is a further breaking change and needs its own note.
- Interpolate non-time Duration columns in `resample` (currently excluded and documented as a limitation): convert them to Int64 microseconds, interpolate, and cast back with Duration-aware null handling.

@cbueth because I changed the code considerably from your original submission, please review all of the changes end to end and leave a comment once you have finished.

cbueth:
@dkrako thank you for reviewing and integrating it. I think it looks good and also good that the test fixtures are not needed anymore.
I also prefer using `Duration('us')` over `Duration('ms')` out of the technical reasons. So we even break the current behaviour as until now, we did not have this accuracy. A good moment to break, allowing better precision. Having all in `_utils._time` is great. Few points:
- Could we have a dedicated 2 kHz end-to-end test, like `from_asc` with 0.5 ms steps -> `Duration('us')` -> detection?
- Ties-to-even rounding is now the convention for all time conversions, which is a fine choice, but let's make sure the changelog calls it out since "half-up" is the more common expectation. If we breka that now, we should be sure about the decision.
- Since `polars.write_csv` can't serialize Duration columns and the helpers are private, the "expose public conversion helpers" follow-up is the one that should be a prio.

cbueth:
As I am author of this PR, I cannot formally approve.

dkrako:
@cbueth Since your last approve I added two additions on top:

- a merge of current main (post v0.28.0). I audited all conflict resolutions; most notably, `reading/measures.py` was resolved wholesale to the #1671 rewrite, with this branch's contribution reduced to a single Duration-to-ms coercion in `compute_reading_measures`.
- one test commit (178f00b1) covering the numeric-time branch in `tsplot.py` that the merge resolution introduced. Without it the 100% coverage gate fails.

CI is green and I will merge as is.

After this merges, I plan a follow-up PR with:

- fractional (float) timesteps support in the I-DT detector, plus 2 kHz detection tests based on `eyelink_monocular_2khz_example.asc`
- documentation on the `Gaze` class that the time columns of `messages`, `calibrations` and `validations` are stored as `Duration('us')` and that numeric time input is interpreted as milliseconds
- calibration/validation save tests that actually exercise the Duration-to-ms CSV conversion (they use a `timestamp` column, which the conversion ignores) plus a round-trip content assertion
- restoring the regression test for keyword-only `Events.__init__` that was dropped as collateral in 8b52e426
- rewording a stale guard comment in `Gaze._convert_time_units`

dkrako (approved):

PR #1639, opened 2026-07-22
fix: Normalize line endings for test files to match .gitattributes eol=lf rule

cbueth:
`.gitattributes` enforces LF line endings for all test files, but `raccoons.asc` and `potsdam_binge_pvt_example.csv` were committed with CRLF before the rule was added. This caused Git to flag them as modified on every checkout. The `.xlsx` file also got wrongly matched as text.

This converts the two text files to LF so they match the `eol=lf` rule, and marks `Sentences.xlsx` as binary since spreadsheets must not undergo text normalization.


cbueth:
DIscarding this PR, I was still basing on an old main.

PR #1602, opened 2026-07-06
docs: reduce nitpicky ignores

cbueth:
## Description

Fix cross-references to pymovements internal objects and remove the corresponding `nitpick_ignore_regex` entries.

**Question for the reviewer:** The class template generates individual pages for each method (e.g. `Gaze.pos2acc`) via an autosummary toctree, but the `[source]` button on those pages is broken because they reference an autoclass directive rather than the actual implementation. I understand that the intermediate pages do serve a purpose: methods like `Gaze.smooth` differ from `transforms.smooth` since they add `**kwargs` passthrough and class-level logic; but the broken `[source]` link is misleading. A possible fix should be removing `:toctree:` from the methods section in the class template to eliminate these intermediate pages and have methods link directly to the class page, where the `[source]` link resolves correctly.

## Implemented changes
- [x] Fix `:py:cls:` typos -> `:py:class:` in docstrings
- [x] Fix cross-refs using wrong submodule paths (e.g. `pymovements.dataset.DatasetPaths` -> `pymovements.DatasetPaths`)
- [x] Add `compute_threshold` to API reference (missing function page)
- [x] Remove 4 nitpick entries: `dataset.(Dataset|DatasetDefinition|DatasetPaths)`, `datasets.Dataset`, `gaze.Experiment`, `events.engbert.compute_threshold|_decompress`
- [x] Add `UnknownMeasure` to API docs, remove its nitpick ignore
- [x] Remove transforms/`pm`/events-module/`deprecated` nitpick ignores
- [x] Add `clip`, `smooth`, `resample` to transforms API docs
- [x] Narrow broad `py:(obj|attr|meth)` catch-all to `py:obj` only (attr/meth refs in docstrings are now fixed with plain text)
- [x] Fix some attr/meth cross-refs in docstrings that lacked API pages by using `:py:class:`~pymovements.ClassName`.attribute pattern
- [x] Fix `smooth.py` docstring bullet list indentation

## How Has This Been Tested?

- [x] Repeatedly:
  - Removing the nitpick rule, fresh docs build run, fix erroring warnings
- [x] Final doc build
- [x] Ignoring `-D nb_execution_mode=off -D "suppress_warnings=myst-nb.lexer,toc.not_included,myst.header"` fresh builds went quite fast.

## Type of change

- Documentation update

## Context

Resolves #1599, #1579, #1567

#### related issues:
- #

#### required by:
- #

#### requires:
- [ ] #

#### conflicts with:
- #

## Checklist:

[...]

PR #1640, opened 2026-07-22
docs: Reduce nitpicky ignores continued

### LINKED ISSUE #1599: reduce nitpicky ignores in sphinx config by using correct references
## Current Problem

Currently there are lots of nitpicky ignores defined in our sphinx `conf.py` (introduced by #1344):

https://github.com/pymovements/pymovements/blob/63fb9fceee04801b0340e271547cff4d9c151acd/docs/source/conf.py#L149-L210

The problem with these is that it's just a bandaid and the references aren't linking correctly. For example on the [Dataset page](https://pymovements.readthedocs.io/en/stable/reference/api/pymovements.Dataset.html) the link to `DatasetPaths` does not work correctly because it is specified as ```:py:class:`~pymovements.dataset.DatasetPaths` ``` in [`dataset.py`](https://github.com/pymovements/pymovements/blob/63fb9fceee04801b0340e271547cff4d9c151acd/src/pymovements/dataset/dataset.py#L73). However the correct sphinx reference is ```:py:class:`~pymovements.DatasetPaths` ```.

## Proposed Solution

I guess most of the ignores can be avoided by using the correct sphinx reference.

For example, 

https://github.com/pymovements/pymovements/blob/63fb9fceee04801b0340e271547cff4d9c151acd/docs/source/conf.py#L199

can be avoided by replacing all occurences of ```:py:class:`~pymovements.datasets.Dataset` ``` with ```:py:class:`~pymovements.Dataset` ```.

#### Related issues

- [ ] #1567

cbueth:
## Description

Enable `numpydoc_show_class_members` so Napoleon documents all attributes listed in a class's `Attributes` docstring section as static attribute entries. This creates link targets that `:py:attr:` cross-references can resolve to, for both regular attributes and `@property`-based ones.

With these targets now available, the broad `nitpick_ignore_regex` entries that suppressed all `py:attr` cross-reference warnings for our high-level classes can be removed entirely.

## Implemented changes

- Enable `numpydoc_show_class_members` to auto-document class attributes
- Remove the `py:attr` and `py:(attr|meth)` nitpick ignore patterns
- Add missing `Attributes` sections to DatasetPaths, Experiment, and Gaze so all properties get documented
- Add schema, columns, frame to Gaze's Attributes; add url, filename, md5, mirrors to ResourceDefinition's Attributes
- Add class-level type annotations where Napoleon needs them
- Fix broken backtick syntax in attribute cross-references
- Remove cross-references to non-existent Dataset attributes (events_rootpath, preprocessed_rootpath)
- Turn Dataset.experiment cross-references into plain text
- Fix Dataset.paths.raw reference (was :py:meth: for a property)
- Remove the template's attribute display loop (Napoleon handles visual rendering now, no duplication)

## Type of change

- Documentation update

## Context

Resolves #1599, builds on #1602


cbueth:
The class-level type annotations were added alongside existing `@property` decorators to make these attributes detectable by pydoclint (DOC602/DOC603) and Napoleon for automatic documentation. The `# type: ignore[no-redef]` suppresses mypy's redefinition warning since annotations and properties with the same name intentionally coexist. For the `ResourceDefinition` dataclass, `field(init=False, repr=False)` prevents the properties from being treated as dataclass fields while keeping them visible to pydoclint.

cbueth:
Example `@property` attribute being properly documented now: https://pymovements--1640.org.readthedocs.build/en/1640/reference/api/pymovements.Gaze.html#pymovements.Gaze.schema.

dkrako (approved):
Thanks a lot @cbueth! Removing those broad ignore patterns in favor of real link targets is a solid improvement, and the pydoclint enforcement will keep the Attributes sections honest from here on.

To finalize, I stacked some commits on top of your work:

- documented all property-backed attributes as class attributes and enforced consistency via pydoclint (`treat-property-methods-as-class-attributes = true`), including the `Dataset.fileinfo` init fix from empty DataFrame to `{}` (6e33d87b)
- dropped the stale `frame` entry from the `Gaze` Attributes section (the property was removed on main by #1660, which is what broke pre-commit.ci)
- fixed a leftover `:py:meth:` role for `Dataset.path`, two stray backticks before `:py:attr:` roles in dataset_files.py, and the section headings in the `DatasetPaths.stimuli` docstring
- switched events.py to a plain `import polars` so the `schema` annotation and docstring can both use the full `polars.type_aliases.SchemaDict` path (pydoclint pins the docstring type to the module's import spelling)

One design note: I would have liked to keep the attributes table at the bottom of the class pages, but it only listed the attribute names with a one-line summary, which the rendered `Attributes` sections now show in full on the same page, so keeping it would just duplicate information. Unrelated but worth noting: the `Example` sections in the property docstrings (for instance on the `DatasetPaths` properties) are not rendered anywhere in the API reference. That was already the case before this PR (the old table had no `:toctree:`, unlike the methods block), so it is out of scope here, I opened #1702 for it but I don't think that has any priority now.

Thanks again, merging once CI is green.

cbueth:
Thanks for the review and merge. I saw you changed many `pl` to `polars`, can you explain what the induced change is / motivation for it? Or is it about this?

>     * switched events.py to a plain `import polars` so the `schema` annotation and docstring can both use the full `polars.type_aliases.SchemaDict` path (pydoclint pins the docstring type to the module's import spelling)

dkrako:
Good catch, thanks for checking! I was actually hoping that this would somehow fix the intersphinx problem but that didnt really work and later on I saw your comment in the conf.py:

https://github.com/pymovements/pymovements/blob/fe3f8f94886ba293fc7cf36bf03c20474402b325/docs/source/conf.py#L170-L172

As gaze.py was formatted the same I left my changes in. I think for the signatures in the reference API docs it would be actually helpful to see `polars.DataFrame` vs `pandas.DataFrame` fully written out but now I also realize that this is probably what `napoleon_type_aliases` is actually for.  It might be worth creating a follow-up issue to discuss how we want to treat these consistently in the future.

cbueth:
Yes, polars interphinx is sadly not working. Generally, there is no guarantee to what a package delivers in their intersphinx. With the nitpicky warns we at least find out when one stops working. I am find with how it is now, but I do not think we are bringing much value to the core features of pymovements discussing this even more in another follow-up.

PR #1644, opened 2026-07-23
fix: validate single-column array shapes in from_numpy

### LINKED ISSUE #640: make passing time array in gaze.from_numpy() more robust by flattening it
## Description of the problem
#634 has exposed an issue with the way we pass the time numpy array to `polars.from_numpy()`.
Up to now we supported 1d and 2d array for passing timesteps.
But the latest polars version now doesn't seem to support passing a 2d array to `from_numpy` (on looking closely it is somehow recognized as an array of objects instead of a 2d array). 

## Description of a solution
I can think of three solutions.

1. It could be that this gets simply resolved by a new polars version. Then we wouldn't need to do much work actually and just wait until it gets resolved upstream.

2. Stop supporting passing 2d arrays for the timesteps. As a user would still get a polars error it would be better to at least raise an exception in pymovements after checking the shape of the time array.

3. Flatten time array to a 1d array. This should be done in this code block before passing it to `pl.from_numpy()`.
https://github.com/aeye-lab/pymovements/blob/a9edc9dada4bf13ab13f034429570cb942dce3be/src/pymovements/gaze/integration.py#L231-L234
Before flattening the time array, we must make sure that all dimensions have a size of 1, except for a single dimension which is the actual timeseries axis, i.e. a shape of `(1, 10)` or even `(10, 1, 1)` should be excepted, but a shape of `(2, 10)` should raise an error.

I personally prefer solution 3, but as this usecase is so obscure and can be resolved by the user themselves, I think solution 2 is also viable. Solution 1 is our default until then.

- [ ] either raise a more helpful error message on 2+d-arrays or flatten nd arrays to a 1d array.

kevinmanase:
## Summary

- flatten singleton dimensions in explicitly supplied `time` NumPy arrays
- reject zero-dimensional arrays and arrays with multiple non-singleton dimensions with a clear shape-specific error
- document the accepted shape contract and add regression coverage for leading, trailing, and surrounding singleton dimensions

Closes #640.

## Verification

- `uv run --python 3.13 --extra dev pytest tests/unit/gaze/integration_numpy_test.py -q` — 36 passed
- `uv run --python 3.13 --extra dev pre-commit run --files src/pymovements/gaze/integration.py tests/unit/gaze/integration_numpy_test.py` — all hooks passed
- `uv run --python 3.13 --extra docs sphinx-build -W --keep-going -b html docs/source /tmp/pymovements-docs-640` — build succeeded


dkrako (approved):
Excellent, thank you so much @kevinmanase! This is a clean implementation of the preferred solution 3 from #640. The shape guard is exactly right: allowing at most one non-singleton dimension means `reshape(-1)` can never silently reorder or merge samples, and the explicit `ndim == 0` check correctly catches scalars that would otherwise slip through the singleton count.

I pushed one commit on top: `trial` and `distance` are passed to `pl.from_numpy()` the same way as `time`, so they were susceptible to the same upstream polars issue. I moved your validation into a shared helper and applied it to all three inputs, parametrized the new tests over all three arguments, and added a row count assertion. Error message and behavior for `time` are unchanged from your version.

Issue #1612, opened 2026-07-17
Petition to improve some default argument values


saeub:
I think some choices for default values could be improved. I constantly run into these, so I decided to start a list.

1. `measures.samples.location()`: `position_column` is currently `'position'` as default. I think the main use case for this property is to find the location of a fixation on screen, so `'pixel'` would be more appropriate. It is also the only use case that is mentioned in [the tutorials](https://pymovements.readthedocs.io/en/stable/tutorials/plotting.html#calculating-fixation-properties)). Changing the value to `'pixel'` requires a nested data structure `("location", {"position_column": "pixel"})` that is not very transparent to beginners.
2. ~`plotting.tsplot()`: `share_y` and ` zero_centered_yaxis` are currently `True` as default. Since the default behavior of that function is to plot all columns in the dataframe, and those are almost certainly not on the same scale, this leads to very confusing results, where some of the channels just appear to be a straight line at y=0.~ ✔️
3. `Dataset.split_gaze_data()`: `by` is currently a required argument, while `Gaze.split()` uses `trial_columns` by default. Since splitting by `trial_columns` is probably the most common use case, I think using it as the default value in `split_gaze_data()` too would be consistent and make sense.
4. `plotting.scanpathplot()`: `arrow_rad` is currently 0.25. Unless there are only very few fixations, this makes the plot look quite difficult to read, I think. Compare to `arrow_rad=0`, which looks less busy and much more understandable to me:  
   <img width="300" alt="Image" src="https://github.com/user-attachments/assets/0c70f52a-3ca7-443f-8ec2-066685fef411" /><img width="300" alt="Image" src="https://github.com/user-attachments/assets/ccca3186-efce-455d-9e28-2a1ac39f877b" />
   I imagine for a simple single-line reading experiment, `arrow_rad=0.25` might make sense because it's easier to distinguish progression from regression, but for any other use case, I think it looks strange.
5. The default parameters for the `ivt` algorithm appear to be suboptimal -- it misses *a lot* (like, more than half) of fixations. The defaults for `idt` seem to work much better (left: I-VT, right: I-DT):  
   <img width="300" alt="Image" src="https://github.com/user-attachments/assets/a5426471-0c28-4a3c-b971-5b82a69c8051" /><img width="300" alt="Image" src="https://github.com/user-attachments/assets/22289216-756f-43b7-919f-b05c9301a992" />  
   This seems quite important, as I-VT with defaults is recommended as the detection algorithm all over the documentation.

[...]

PR #1646, opened 2026-07-27
feature: plot event overlays in `tsplot`

saeub:
## Description

Optionally plot events as shaded rectangles in `tsplot`. Events are colored based on name and added to the `fig.legend()`:

```python
fig, ax = pm.plotting.tsplot(gaze, ["position_x", "position_y", "velocity_x", "velocity_y"], plot_events=True)
fig.legend()
```

image

## Implemented changes

- [x] Add `plot_events` argument to `tsplot`

## How Has This Been Tested?

- [x] Unit test to check that rectangles are added, correctly colored, and added to legend

## Type of change

- New functionality

## Checklist:


cbueth:
That looks nice! If one has more types of events they would get displayed, and with what color? Are the axe limits set by the screen resolution?
I think it would be nice to have a little space between the top and bottom row, but still using sharex axis.

saeub:
@cbueth

> If one has more types of events they would get displayed, and with what color?

Yes, see screenshot. It just uses the `tab10` color map from matplotlib.

> Are the axe limits set by the screen resolution?

The ylim is currently a bit buggy, and also not very intuitive (see #1612). Here ylim is set based on the min/max of the entire recording, but I'm just showing a small excerpt, so that's also why it looks a bit weird.

> I think it would be nice to have a little space between the top and bottom row, but still using sharex axis.

Agree. I think the layout generally needs an overhaul (see #1612). For the example I applied `fig.tight_layout()`, without that it's even worse (labels overlapping plots etc.). But I'm assuming that should be a separate PR.

saeub:
Not sure why the test is failing (Python/matplotlib version issue?)

dkrako (approved):
Great, thank you so much @saeub! The event overlay is a solid addition to `tsplot`. To get this mergeable I pushed a few commits on top of your work; summary so you can see what changed:

- `tsplot` crashed for any `Gaze` without a `time` column (a plain-DataFrame gaze without an experiment), even without event plotting. It now falls back to the sample index, with a regression test.
- More than 10 unique event names raised `KeyError` (the `zip` with the tab10 colors truncated silently), so colors now cycle via modulo, with a test for 11 names.
- The red minimum-requirements job was the test assuming `axvspan` returns a `Rectangle`; on matplotlib 3.8 it returns a `Polygon`. The test now asserts on `ax.patches` directly, which works on both.
- Renamed `plot_events` to `show_events` for consistency with `show_grid` and `show_yticks`.
- Events access and color computation are now gated behind `show_events` and computed once per figure instead of per channel.

Selecting which event types to shade (e.g. only fixations) is captured as a follow-up in #1682.

One behavior change to be aware of: the x-axis now shows the `time` column (usually ms) when present, sample index otherwise. Previously it was always the sample index, so existing plots change scale for sampling rates other than 1000 Hz. This is documented in the docstring, and it partially addresses #1684 (the x-axis criterion; gaps in the data are no longer compressed, though the remaining criteria there stay open for a follow-up).

Thanks again, merging once CI is green.

PR #1647, opened 2026-07-27
docs: fix Savitzki-Golay citation usage

saeub:
Not sure what the `|_|` means, but it prevents the citation from being rendered properly.


cbueth (approved):
Thank you for finding this 👍

PR #1648, opened 2026-07-28
fix: loading stimuli with subsets and optional patterns

saeub:
## Description

This PR covers two errors that happen when calling `dataset.load()`:

- When stimulus filepath patterns are defined with optional capture groups (i.e., a metadata field is `None` only for some of the files), polars complained because it only uses the first row to determine the dataframe schema. I removed the `infer_schema_length=1` option to let polars use the default (i.e., 100 rows). **I'm not sure why this argument was added -- if there is a reason, let me know.**
- When `dataset.load()` was called with a `subset` argument, the stimulus files were not loaded. This is due to an incorrect check of the `content` attribute. Now it is checked the same way as in other places.

These are both tiny and related fixes, so I thought it's okay to bundle them into a single PR.

## Type of change

- Bug fix

## Checklist:


dkrako (approved):
Great, thank you so much @saeub! Both diagnoses were correct, and the `content` check fix is right: it now matches how stimulus content is checked everywhere else in the codebase.

To your question about `infer_schema_length=1`: it dates back to my dataset_files refactor in #308 and I honestly can't say why I changed it. At the time all fileinfo values were plain regex capture strings, so one row sufficed; it only became a problem once optional capture groups appeared. No reason to keep it.

I finalized the PR with a few small commits on top of your work:

- Changed your `pl.from_dicts(data=filepaths)` to `pl.from_dicts(data=filepaths, infer_schema_length=None)`. The polars default only infers from the first 100 rows, so the same `ComputeError` would still occur when the optional field is missing for the first 100 files and present only later, which is realistic for larger datasets. Inferring over all rows is cheap for file lists.
- Added two regression tests: `test_scan_dataset_optional_pattern_field_missing_in_first_rows`  and `test_load_subset_loads_stimuli`.
- Sorted the matched filepaths before building the fileinfo dataframe (replacing the dataframe sort afterwards), so schema inference sees a deterministic file order regardless of filesystem iteration order (helped me with `test_scan_dataset_optional_pattern_field_missing_in_first_rows` and is no actual behavior change from the caller side).

Thanks again @saeub, merging once CI is green.

dkrako:
pre-commit.ci run

Issue #1136, opened 2025-04-29
extract before remove


SiQube:
## Current Behavior

> The UCL dataset fails the tests in #1049 if `remove_finished=True`.
> 
> Reason: resource filename is duplicated, hence archive is already removed before being extracted (again).
> 
> This is just a workaround before fixing the underlying issue of handling filename duplicates in resources.
> 
> Here's one of the failed jobs: https://github.com/aeye-lab/pymovements/actions/runs/13903514974/job/38903529945?pr=1049

https://github.com/aeye-lab/pymovements/pull/1050#issue-2925756369

## Expected Behavior

first extract all necessary data, then remove the archive

## Minimum acceptance criteria

- [ ] UCL dataset has a single download and passes `remove_finished`

if merged:
- [ ] CoLAGaze dataset has a single download and passes `remove_finished`
- [ ] ETDD70 dataset has a single download and passes `remove_finished`

## Checklist

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed
- [x] I have provided sufficient information for the team

PR #1652, opened 2026-07-30
feat: add name-keyed DatasetDefinition.sources

dkrako:
## ⚠️ Behavior changes

* **Source validation now happens at construction time.** `DatasetDefinition` validates its sources on init: dangling source references, unused named sources, duplicate source filenames, and conflicting resolved sources. Fully identical duplicate inline sources remain valid and are deduplicated silently. However, custom definitions whose duplicated sources have drifted apart (same url and filename but differing `md5` or `mirrors`, or the same filename under different urls) previously constructed and downloaded without error and will now fail at construction, including via `from_yaml` and in load-only workflows. Fix by deduplicating into one named source.
* **One-time re-download for ETDD70 and UCL.** The shared-source migration renames the downloaded archives (`edd_raw.zip`/`edd_fix.zip` → `etdd70.zip`, `UCL_events.zip`/`UCL_measures.zip` → `UCL.zip`), so existing local copies of these two datasets will be re-downloaded once. As a side effect, ETDD70 no longer downloads the same archive twice. MCFW-Gaze and CERVINO keep their archive filenames and are not affected.

## Description

This adds a shared, name-keyed `DatasetDefinition.sources` mapping and lets a
`ResourceDefinition.source` be either an inline `WebSource` (as before) or a
string referencing an entry in `sources`. Multiple resources can now point at a
single shared source, which is deduplicated so each file is downloaded only once.
The new `DatasetDefinition.resolve_source()` and `DatasetDefinition.resolved_sources()`
methods expose the resolved and deduplicated sources.

This is a step towards BIDS compatibility (mirroring `DatasetLinks` behaviour) and
towards allowing selection of which sources to download in `Dataset.download()`.
Unlike `DatasetLinks`, sources and resources here are pre-harmonization.

Old dataset definitions keep working unchanged: inline `WebSource` sources, the
deprecated `url`/`filename`/`md5`/`mirrors` fields, and YAML with inline `source:`
dicts all behave as before.

## Implemented changes

- [x] Add `DatasetDefinition.sources: dict[str, WebSource]` (accepts dicts and
      converts them via `WebSource.from_dict`)
- [x] Allow `ResourceDefinition.source` to be a `str` reference into `sources`
- [x] Validate at construction: dangling references, unused named sources,
      duplicate source filenames, and conflicting resolved sources (same
      `(url, filename)` with differing `md5`/`mirrors`, or the same filename
      claimed by different urls)
- [x] Add `DatasetDefinition.resolve_source()` and
      `DatasetDefinition.resolved_sources()`; download iterates the resolved
      sources deduplicated by `(url, filename)` so each file is downloaded only
      once; both re-validate on call, so dangling references or conflicts
      introduced by post-init mutation raise `ValueError`
- [x] Skip extraction for resources without a source; raise a clear
      `AttributeError` if a resource resolves to a source without a filename
      (such a file can never have been downloaded)
- [x] Guard the deprecated `url`/`filename`/`md5`/`mirrors` properties so they
      raise a clear error when the source is a named reference
- [x] Migrate the shipped `ETDD70`, `UCL`, `MCFW-Gaze` and `CERVINO` definitions
      (including their docs meta files) to a single shared named source each
- [x] Serialize named sources in `to_dict`/`to_yaml`, round-trip them via
      `from_yaml`, and show `sources` before `resources` in the HTML repr
- [x] Document named shared sources in the dataset contributing guide

## How Has This Been Tested?

- [x] New unit tests in `resources_test.py`, `dataset_definition_test.py`,
      `dataset_download_test.py`, and a new `shipped_datasets_test.py`
      (references resolve, unique filenames, no duplicate `(url, filename)`)

## Type of change

- New functionality
- Breaking change (duplicated sources that disagree on `md5`/`mirrors`, or share
  a filename under different urls, now raise `ValueError` at construction time)
- Documentation update

#### related issues:
- #1136

## Checklist:


src/pymovements/dataset/dataset_definition.py
```diff
@@ -403,6 +433,204 @@ def __init__(
...
+                    raise ValueError(
+                        f"Dangling source reference: '{resource.source}' "
+                        f"in resource '{resource.content}'",
+                    )
+                referenced_names.add(resource.source)
+
+        unused_names = sources.keys() - referenced_names
+        if unused_names:
+            name = sorted(unused_names)[0]
+            raise ValueError(f"Unused source: '{name}' is not referenced by any resource.")
```
cbueth, line 483: Requiring every named source to be referenced by some resource contradicts the stated goal of later selecting which sources to download, since you cannot pre-declare an optional source. It also only reports one unused name at a time. I would make this a warning, or drop it and let download decide what to fetch.

This might be the most important point.

src/pymovements/dataset/dataset_definition.py
```diff
@@ -403,6 +433,204 @@ def __init__(
...
+        ----------
+        sources: dict[str, WebSource]
+            Mapping of source names to sources to validate.
+        """
+        filenames: set[str] = set()
+        for source in sources.values():
+            if source.filename is not None:
+                if source.filename in filenames:
+                    raise ValueError(f"Duplicate source filename: '{source.filename}'")
+                filenames.add(source.filename)
```
cbueth, line 452: Either deduplicate identical named sources here or remove this method and rely on the resolved check.

`_validate_sources` rejects duplicate filenames ignoring url, and rejects even two fully identical named sources, while identical inline sources are silently deduplicated. It also runs first and pre-empts the more descriptive same-filename-different-urls error. Since every named source must be referenced, resolved validation already covers all named sources, so this check is largely redundant and inconsistent.

src/pymovements/dataset/dataset_definition.py
```diff
@@ -403,6 +433,204 @@ def __init__(
...
+                )
+            # A repeated filename with a different url would silently overwrite the first
+            # download.
+            if source.filename is not None:
+                if source.filename in filename_urls:
+                    raise ValueError(
+                        f"Conflicting sources for filename '{source.filename}': the same "
+                        f"filename is claimed by different urls "
+                        f"('{filename_urls[source.filename]}' != '{source.url}').",
+                    )
```
cbueth, line 596: See comment on `_validate_sources`.

src/pymovements/dataset/dataset_definition.py
```diff
@@ -403,6 +433,204 @@ def __init__(
...
+            sources: dict[str, WebSource | dict[str, Any]] | None,
+    ) -> dict[str, WebSource]:
+        """Initialize sources mapping, converting dict values to ``WebSource`` if necessary."""
+        if sources is None:
+            return {}
+
+        return {
+            name: source if isinstance(source, WebSource) else WebSource.from_dict(source)
+            for name, source in sources.items()
+        }
```
cbueth, line 496: There should be a guard for symmetry.

No value type validation. `sources={'main': 'http://...'}` falls into `WebSource.from_dict` and fails with an unclear `AttributeError` as `str` has no `get`, while `ResourceDefinition.from_dict` raises a clear `TypeError`.

src/pymovements/dataset/resources.py
```diff
@@ -163,6 +167,29 @@ def __init__(
...
+        source = self._inline_websource(attr)
+        return getattr(source, attr) if source else None
+
+    def _set_websource_attribute(self, attr: str, value: Any) -> None:
+        """Set an attribute of the inline ``WebSource``, creating one without a source."""
+        source = self._inline_websource(attr)
+        if source is None:
+            self.source = WebSource(**{'url': None, attr: value})
+        else:
+            self.source = replace(source, **{attr: value})
```
cbueth, line 191: Not too grave, but building `WebSource(**{'url': None, attr: value})` relies on duplicate-key collapsing when `attr` is `url`, and always injects `url=None` for the `filename`/`md5`/`mirrors` setters. It matches the old behavior but is obscure. the old explicit branches were clearer and the removed type ignores are no longer justified by the dynamic dict.

src/pymovements/dataset/dataset_download.py
```diff
@@ -73,23 +73,20 @@ def download_dataset(
...
+    sources = definition.resolved_sources()
 
-    downloadable_resources = [resource for resource in definition.resources if resource.source]
-
-    if not downloadable_resources:
+    if not sources:
         raise AttributeError(
             'No downloadable resources found in DatasetDefinition. '
             'ResourceDefinition.source must be specified to download a dataset.',
         )
```
cbueth, line 86: Should be reworded as the condition changed ->  no downloadable sources to match the new model

src/pymovements/dataset/dataset_download.py
```diff
@@ -149,7 +152,16 @@ def extract_dataset(
...
             for resource in definition.resources.filter(content):
-                source_path = paths.downloads / resource.source.filename
+                source = definition.resolve_source(resource)
+                if source is None:
+                    continue
+                if source.filename is None:
+                    raise AttributeError(
+                        'WebSource.filename must not be None for source of resource '
+                        f"with content '{resource.content}'",
+                    )
```
cbueth, line 162: the PR advertises re-validation on call

Good addition, but note `extract_dataset` only calls `resolve_source` and never `_validate_resolved_sources`, so a filename/url conflict introduced by post init mutation is caught by download but not by extract.

tests/unit/dataset/shipped_datasets_test.py
```diff
@@ -0,0 +1,49 @@
...
+@pytest.fixture(name='definition')
+def fixture_definition(name):
+    """Get the dataset definition for a given name."""
+    return DatasetLibrary.get(name)
+
+
+@pytest.mark.parametrize('name', DatasetLibrary.names())
+def test_shipped_definition_constructs(definition, name):
+    """Test that the shipped definition constructs and passes init validation."""
+    assert definition.name == name
```
cbueth, line 35: no change needed, just commenting that `test_shipped_definition_sources_no_duplicate_url_filename` partly duplicates the new init validation, but it is still useful as a guard over shipped data

cbueth (changes requested):
The design is solid and the tests are quite detailed. See my comments inline.

There are two blocking points. First, the description validates unused named sources but also says this is a step towards selecting which sources to download. If selection is at the source level, unused sources must be allowed. if at the resource level, the error is fine. Description and code need to agree, pick one
Second, the description says fully identical duplicates are deduplicated silently, but two identical named sources with the same filename raise. Either deduplicate identical named sources too, or state that named sources must be unique by filename.

Furthermore, few points
- Add a clear TypeError in `_initialize_sources` for non-dict/non-WebSource values, matching `ResourceDefinition.from_dict`
- the download error message ("resources" -> "sources") and the docstring in dataset.py need to match
- `extract` only calls `resolve_source`, not `resolved_sources`, so the "re-validate conflicts on call" claim is not satisfied on the extract path
- Archive rename for ETDD70/UCL forces a one-time re-download. I think this is fine

src/pymovements/dataset/dataset_definition.py
dkrako: Agreed. The init-time error is dropped entirely in fdd77f23, unused named sources are now legal (no warning either). Selection will happen at the source level via a `Dataset.download(sources=...)` parameter in a follow-up PR.

src/pymovements/dataset/dataset_definition.py
dkrako: Fixed in fdd77f23. Fully identical named sources now deduplicate silently, same as inline ones. The descriptive same-filename-different-urls error surfaces first instead of the filename-only error.

Sources that share url and filename but differ in md5 or mirrors only raise at the resolved level, i.e. when they are actually co-referenced by resources.

I kept `_validate_sources` itself: now that unused named sources are legal it is no longer redundant with the resolved check, and it also enforces that every named source defines a url (d9001e1d, see the setter thread for context).

src/pymovements/dataset/dataset_definition.py
dkrako: Addressed together with the `_validate_sources` thread above.

src/pymovements/dataset/dataset_definition.py
dkrako: Added in fdd77f23: `_initialize_sources` now raises a `TypeError` for non-dict/non-WebSource values, matching the `ResourceDefinition.from_dict` wording, with a test covering the plain-string case.

src/pymovements/dataset/dataset_download.py
dkrako: Reworded to "No downloadable sources found in DatasetDefinition." in 7eb3ec24.

src/pymovements/dataset/dataset_download.py
dkrako: Good catch. `extract_dataset` now calls `definition.resolved_sources()` up front (7eb3ec24), so a conflict introduced by post-init mutation raises on the extract path too. There is a test that mutates a resource to a conflicting md5 after init and asserts `extract()` raises, and the ValueError is documented on both `Dataset.extract` and `extract_dataset`.

tests/unit/dataset/shipped_datasets_test.py
dkrako: Agreed, keeping it as a guard over the shipped definitions.

src/pymovements/dataset/resources.py
dkrako: Restored the explicit branches per setter in d9001e1d.

While doing that I also widened `WebSource.url` to `str | None`, which removes all four `type: ignore[arg-type]` comments. The annotation now matches what these setters actually do: setting `filename`, `md5` or `mirrors` before `url` creates a `WebSource` without a `url`.

`url` stays a required argument without a default, `download()` keeps its `None` guard (which now does the real type narrowing), and named sources must define a `url` at `DatasetDefinition` init.

Once the deprecated properties are removed we may discuss if it's worth narrowing back the annotation to `str`.

dkrako:
Thanks a lot @cbueth for your thorough review! All your comments are addressed in my last commits and replied to.

Ready for another look.

PR #1653, opened 2026-07-31
fix: skip OS metadata files during archive extraction

### LINKED ISSUE #1346: do not try to extract MAXOSX metadata files
## Current Behavior

During working on #1332 I encountered the following error:

If there's a MAXOSX metadata file in an archive that ends with `.zip` (or any other supported archive extension), `Dataset.extract()` will treat it as a zip file and will try to extract the content.

However, these MACOSX files are just metadata files and cannot be treated as an archive to be extracted.

## Expected Behavior

MACOSX metadata files should be excluded from nested extraction.

Moreover, it would probably be great to completely exclude these files when extracting archives.

- [ ] exclude MACOSX metadata files from nested extraction

Sanjays2402:
## Description

`Dataset.extract()` fails on archives created on macOS. Such archives carry resource forks in a `__MACOSX` directory as `._` prefixed files that mirror the names of the files they belong to, so an archive containing `inner.zip` also contains `__MACOSX/._inner.zip`. The recursive branch of `extract_archive` collects nested archives by extension, so that metadata twin is picked up and handed to the zip extractor, which raises `BadZipFile: File is not a zip file`.

## Implemented changes

- [x] Filter `__MACOSX` entries and `._` prefixed files out of `archive_filepaths` before recursing in `extract_archive`
- [x] Add a regression test covering a nested archive alongside its `__MACOSX` metadata twin

## How Has This Been Tested?

- [x] New `test_extract_archive_skips_macosx_metadata_files` in `tests/unit/dataset/_utils/_archives_test.py`: fails on the unpatched source with the reported `BadZipFile`, passes with the fix (verified by stashing the source change)
- [x] `pytest tests/unit/dataset/_utils/_archives_test.py` — 236 passed
- [x] `flake8` clean on both changed files

## Type of change

- Bug fix

## Context

Resolves #1346

## Checklist:

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.


src/pymovements/dataset/_utils/_archives.py
```diff
@@ -133,6 +133,14 @@ def extract_archive(
         archive_filepaths = get_filepaths(path=destination_path, extension=archive_extensions)
         archive_filepaths = [filepath for filepath in archive_filepaths if filepath != source_path]
 
+        # macOS stores resource forks in a __MACOSX directory as ._ prefixed metadata files.
+        # They mirror the names of the files they belong to, so a ._archive.zip is not an
+        # archive and must not be passed to an extractor.
+        archive_filepaths = [
+            filepath for filepath in archive_filepaths
+            if '__MACOSX' not in filepath.parts and not filepath.name.startswith('._')
```
cbueth: the `._` prefix filter is broader than strictly needed. the `__MACOSX` guard alone already covers the reported case (AppleDouble files in zips always live in `__MACOSX/`). Keeping both is fine as defense-in-depth. Just noting it also skips any legitimately-named `._foo.zip` (not that I would name my files like that).

tests/unit/dataset/_utils/_archives_test.py
```diff
@@ -586,3 +587,20 @@ def test_extract_archive_destination_path_not_None_no_remove_top_level_no_remove
...
+    archive_path = tmp_path / 'test.zip'
+    with zipfile.ZipFile(archive_path, 'w') as zip_open:
+        zip_open.writestr('inner.zip', inner_buffer.getvalue())
+        # macOS stores resource forks as ._ prefixed files inside a __MACOSX directory.
+        zip_open.writestr(os.path.join('__MACOSX', '._inner.zip'), b'\x00\x05\x16\x07')
+
+    destination_path = tmp_path / 'extracted'
+    extract_archive(source_path=archive_path, destination_path=destination_path, recursive=True)
+
+    assert (destination_path / 'inner' / 'test.file').is_file()
```
cbueth, line 647: the single assertion is sufficient since a `BadZipFile` would have raised pre-fix. optionally you could add `assert not (destination_path / '__MACOSX' / '._inner.zip').exists()` as an explicit check that the twin is untouched.

cbueth (approved):
Hi, this is a nice, focused fix for #1346. The filter in the recursive branch of `extract_archive` correctly prevents macOS resource-fork twins (`__MACOSX/._inner.zip`) from being passed to the zip extractor, and the regression test faithfully reproduces the reported `BadZipFile`. The metadata twin raises `BadZipFile: File is not a zip file` when passed to the zip extractor.
All CI checks pass 👍 
There are two optional nits, neither blocking, see them below. Otherwise this is good to merge. Thanks!

tests/unit/dataset/_utils/_archives_test.py
```diff
@@ -586,3 +625,55 @@ def test_extract_archive_destination_path_not_None_no_remove_top_level_no_remove
...
+        # macOS metadata that must be skipped during extraction.
+        pytest.param('test.zip', '.DS_Store', True, id='zip_dsstore'),
+        pytest.param('test.zip', os.path.join('__MACOSX', '._data.txt'), True, id='zip_macosx'),
+        pytest.param('test.tar', '.DS_Store', True, id='tar_dsstore'),
+        # ._ prefixed files outside a __MACOSX directory may be legitimate and must be kept.
+        pytest.param('test.zip', '._data.txt', False, id='zip_underscore_outside_macosx'),
+        pytest.param('test.tar', '._data.txt', False, id='tar_underscore_outside_macosx'),
+        # A __MACOSX directory is macOS cruft at any depth, not only at the top level.
+        pytest.param(
+            'test.zip', os.path.join('data', '__MACOSX', 'file.txt'), True,
```
dkrako, line 664: I'm not 100% sure with this one but I tend to think it's safe to skip the __MACOSX directory at any level

dkrako (commented):
Thank you so much for your initiative @Sanjays2402, solid fix. I verified it against the actual GazeOnFaces archive from OSF: the nested `database3_sciencemuseum.zip` ships `database3_sciencemuseum/gaze_csv.zip` next to its `__MACOSX/database3_sciencemuseum/._gaze_csv.zip` twin, which is exactly what raised `BadZipFile`.

I pushed a few follow-up commits to take it a bit further:

- Skip the metadata at extraction time rather than only during nested extraction. The original filter kept the twins out of the recursive branch, but `_extract_zip`/`_extract_tar` still wrote them to disk (the real dataset leaves 27 `__MACOSX` entries and their `._` twins behind). Filtering inside the extractors means the metadata never lands on disk, which is the "completely exclude these files" part of #1346. The recursive-branch filter is dropped since it is now redundant.
- Define metadata as a `__MACOSX` directory (at any depth) or a `.DS_Store` filename. The bare `._` prefix is intentionally not matched, so a legitimately named `._foo` file outside `__MACOSX` is preserved. This also addresses @cbueth's note that the `._` filter was broader than needed.
- Also skip `.DS_Store`, the same class of Finder cruft (slightly beyond the original #1346 scope, flagging it explicitly).
- Drop the "cannot be fully processed" warning from `gaze_on_faces.yaml`, since the dataset now extracts end to end (the generated `docs/.../GazeOnFaces.yaml` was regenerated to match).
- Rework the tests into a parametrized case (zip/tar, skip/keep, top-level and nested) plus a small local `make_archive` fixture, keeping the nested-twin case as the dedicated #1346 regression.

@cbueth Please review my changes as they extended the original PR by some bit.

dkrako:
@Sanjays2402 one more thing: feel free to add yourself to our [`CITATION.cff`](https://github.com/pymovements/pymovements/blob/main/CITATION.cff) so you're properly attributed in the next release. I'll hold the PR open a little longer to give you the chance; if you'd rather not, just react with a 👎 and I'll merge as is.

dkrako:
The benchmark failure is unrelated. The flagged benchmarks are the I-HMM and import ones, which this PR doesn't touch. Looks like temporary runner variance, re-runs don't clear it.

cbueth:
@dkrako would you also add a windows hidden files ignore, specifically `thumbs.db`?

Sanjays2402:
Yes — this is already covered on the current head. Commit `dcb7231` skips `Thumbs.db` case-insensitively in both ZIP and TAR extraction, including nested paths, and adds `zip_thumbsdb`, `tar_thumbsdb`, and nested lowercase coverage in `_archives_test.py`. No additional change is needed from my side.

cbueth (approved):

PR #1654, opened 2026-08-04
fix: improve data quality reporting

dkrako:
## Description

Please include a clear and concise description of the change and which issue is fixed.
Please also include excepted improvements, relevant motivation and context.
List any dependencies that are required for this change.

## Implemented changes

Insert a description of the changes implemented in the pull request.

- [ ] Change 1
- [ ] Change 2
- [ ] Change 3

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes.
Provide instructions so we can reproduce.
Please also list any relevant details for your test configuration

- [ ] Test A
- [ ] Test B

## Type of change

Remove irrelevant items:
- Bug fix
- New functionality
- Performance improvement
- New dataset
- Breaking change
- Documentation update

## Context

Follow up to #1579 

Resolves #

#### related issues:
- #

#### required by:
- #

#### requires:
- [ ] #

#### conflicts with:
- #

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] I have checked my code and corrected any misspellings

PR #1655, opened 2026-08-05
feat: add BIDS-style sources metadata to loaded and derived objects

dkrako:
## Description

This PR adds a `sources` entry to the `metadata` dictionary of `Gaze`, `Events`, `PrecomputedEventDataFrame`, `ReadingMeasures`, `TextStimulus` and `ImageStimulus`. It is populated automatically by all loading paths and propagated to derived objects, so every object can report which files it was generated from, independent of how it was loaded.

BIDS defines a `Sources` metadata field that lists the source files a derived file was generated from. pymovements had no equivalent: file provenance was only available through `Dataset.fileinfo`, which requires an initialized and scanned dataset and is unavailable when loading single files.

Design decisions:
- Files loaded standalone (e.g. `gaze.from_csv`) are recorded with their resolved absolute path; files loaded via `Dataset` are recorded relative to the dataset root.
- On reload of a previously saved file (e.g. `from_ipc`), `sources` lists the file that was actually read. Alternative would be reading out the source from an optional metadata file. Currently only the `Gaze` class writes a metadata json file and there's no functionality to load it so I chose that for now. Worth discussing.
- A `sources` entry already present in user-provided metadata is respected and never overwritten. It is validated to be a list of path strings; a malformed entry raises a `TypeError`.
- Writing sidecar files and BIDS URI support (`bids::`) are out of scope for this PR.

## Implemented changes

- [x] New private helpers `add_source`, `relativize_sources` and `merge_sources` in `pymovements._utils._sources`; `merge_sources` deduplicates entries while preserving order, comparing `str` and `Path` entries as POSIX-style path strings
- [x] `gaze.from_csv`, `from_asc`, `from_ipc` and `from_begaze` record the loaded file in `metadata['sources']` and document the entry as path-input only
- [x] `TextStimulus.from_csv`/`from_file` and `ImageStimulus.from_file`/`from_files` record the loaded stimulus files; `from_files` with an empty file list omits the `sources` entry
- [x] Dataset loading records sources for gaze, event, precomputed event, precomputed reading measure and stimulus files, relativized to the dataset root
- [x] New `metadata` attribute on `Events`, `PrecomputedEventDataFrame` and `ReadingMeasures`
- [x] Propagation to derived objects: gaze to detected events (sources merged after detection method resolution), `Events.map_to_aois` and `Gaze.map_to_aois` merge stimulus sources, `compute_reading_measures` unions event and AOI sources, `clone`/`split` on `Gaze`, `Events`, `TextStimulus` and `split_precomputed_events` preserve metadata

## How Has This Been Tested?

- [x] New unit test modules: `tests/unit/gaze/io/sources_test.py`, `tests/unit/events/events_metadata_test.py`, `tests/unit/events/precomputed_test.py`
- [x] Sources behaviour of the public API is covered in the feature test modules it belongs to (gaze io, detection, AOI mapping, stimuli, dataset file loading, reading measures) rather than in dedicated private-helper test modules
- [x] Full unit test suite passes locally, 100% coverage on new and changed code

## Type of change

- New functionality

## Context

Required by #1654

BIDS spec:
- [`Sources` metadata field](https://bids-specification.readthedocs.io/en/stable/derivatives/common-data-types.html#common-file-level-metadata-fields) (derivatives, common file level metadata fields)
- [BIDS URI](https://bids-specification.readthedocs.io/en/stable/common-principles.html#bids-uri) (common principles; out of scope here)

## Checklist:


src/pymovements/_utils/_sources.py
```diff
@@ -0,0 +1,158 @@
...
+    file: Any
+        Path of the loaded source file. Values that are not :py:class:`str` or
+        :py:class:`~pathlib.Path` are ignored.
+
+    Returns
+    -------
+    dict[str, Any]
+        Copy of the metadata dictionary with a ``sources`` entry.
+    """
+    metadata = dict(metadata) if metadata else {}
```
cbueth, line 60: The dict is copied shallowly. An existing `sources` list stays shared. `relativize_sources` then edits that list in place. So a caller's list can be rewritten.

Please fix by building a new list in `relativize_sources`.

src/pymovements/_utils/_sources.py
```diff
@@ -0,0 +1,158 @@
...
+    Raises
+    ------
+    TypeError
+        If the ``sources`` entry is not a list of path strings.
+    """
+    _check_sources(metadata)
+
+    resolved_root = root.resolve()
+    sources = (metadata or {}).get('sources') or []
+    sources[:] = [_relativize_source(source, resolved_root) for source in sources]
```
cbueth, line 120: See above, fix here.

src/pymovements/_utils/_sources.py
```diff
@@ -0,0 +1,158 @@
...
+absolute for standalone loading, relative to the dataset root when loaded via
+:py:class:`~pymovements.Dataset`.
+"""
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Any
+
+
+def add_source(metadata: dict[str, Any] | None, file: Any) -> dict[str, Any]:
```
cbueth, line 37: The description promises a TypeError for malformed sources. But the validation only runs in merge and relativize. Directly built `Events`, `ReadingMeasures` or stimuli skip it. check in `add_source` or in the constructors.

src/pymovements/_utils/_sources.py
```diff
@@ -0,0 +1,158 @@
...
+    dict[str, Any]
+        Copy of the metadata dictionary with a ``sources`` entry.
+    """
+    metadata = dict(metadata) if metadata else {}
+    if 'sources' not in metadata and isinstance(file, (str, Path)):
+        metadata['sources'] = [Path(file).resolve().as_posix()]
+    return metadata
+
+
+def _check_sources(metadata: dict[str, Any] | None) -> None:
```
cbueth, line 66: Can be checked here

src/pymovements/dataset/dataset.py
```diff
@@ -1176,6 +1186,12 @@ def measure_reading(
 
             events.map_to_aois(aoi_text_stimulus)
 
+            # The reading measures are derived from the events and their AOI stimulus files.
+            # AOI files below the dataset root are recorded relative to it, like all other
+            # dataset sources.
+            relativize_sources(events.metadata, self.paths.dataset)
```
cbueth, line 1192: `compute_reading_measures` relativises `self.events` metadata in place. It returns a new object but rewrites the dataset's events. Relativise a copy instead.

cbueth (changes requested):
This adds a `sources` list to metadata, mirroring BIDS `Sources`. New helpers are in `pymovements._utils._sources`. Loaders record the file they read. Dataset loading stores paths relative to the dataset root. Sources propagate to events, AOIs, reading measures and splits/clones.

Three loader functions now require a `paths` argument. Loaded objects obtain a `sources` key -> metadata assertions change. `map_to_aois` and `compute_reading_measures` now mutate metadata. `Gaze.__eq__` and `Events.__eq__` do not care about metadata.

src/pymovements/_utils/_sources.py
dkrako: Good catch. I went one step further than a new list: relativize_sources is now a pure function in 652cc3b4. It returns a copy of the metadata dict with a freshly built sources list and callers assign the result back.

Also added a regression test that a user-supplied list survives dataset loading unchanged.

src/pymovements/_utils/_sources.py
dkrako: Done in 652cc3b4. The function returns a new dict with a new list now. I also made merge_sources a pure function for consistency, so none of the helpers mutate their arguments anymore.

src/pymovements/_utils/_sources.py
dkrako: Good catch again! `add_source()` now runs `_check_sources()` on the passed metadata in 652cc3b4, so every loader raises the promised `TypeError` on a malformed sources entry.

Added parametrized tests over the four gaze loaders for both malformed shapes.

src/pymovements/dataset/dataset.py
dkrako: Fixed in 652cc3b4 with `relativize_sources()` being pure function.

dkrako:
Thanks @cbueth  for the thorough review, all your findings are addressed in 652cc3b4.

As #1625 was merged in the mean time I added 7ea99644 that implements the sources similar to how it's done for reading measures.

Ready for another look.

PR #1662, opened 2026-08-05
fix: skip all-NaN windows in numpy filter helpers

Randomidous:
## Description
`filter_candidates_remove_nans` in `pymovements/events/_utils/_filters.py` raised an `IndexError` when a fixation candidate window contained only NaN gaze values. The leading-NaN stripping loop incremented `start_id` past the end of the array, then the subsequent end-stripping loop accessed `cand_values[start_id]` with an out-of-bounds index.

This surfaces whenever gaze data has a gap long enough to fill an entire candidate window (e.g. during an extended blink or tracker dropout)

## Implemented changes

- [x] Added an early-exit guard in `filter_candidates_remove_nans`: after the leading-NaN loop, if start_id >= len(cand_values) the candidate is fully NaN and is skipped via continue instead of proceeding to the end-stripping loop

## How Has This Been Tested?
- new parametrized test case `test_all_nan_candidate_skipped` added to `tests/unit/events/_utils/_filters_test.py`. It passes a single 7-sample candidate where every sample is (nan, nan) and asserts the result is an empty list (candidate dropped, no crash).

## Type of change

Remove irrelevant items:
- Bug fix
- Performance improvement

## Checklist:


dkrako (approved):
Thank you so much for this fix @Randomidous, and welcome to pymovements! Good catch on the `IndexError` for all-NaN candidate windows.

The fix is correct: bounding the leading-NaN strip loop and dropping fully-NaN candidates is the right semantics here. A candidate with zero valid samples carries no information, and all call sites (`ivt.py`, `microsaccades.py`, `idt.py`) tolerate fewer or empty candidates, so the behavior change is crash-to-drop only.

### Maintainer finalization

I pushed a commit (35c5c74) addressing the review findings:

- Applied the same drop-empty guard to the sibling function `events_split_nans`, which crashed on the identical all-NaN edge case (currently unreachable through the public API because `idt.py` filters first, but the two functions now behave consistently), with a test param covering it.
- Tightened the test assertions: the previous `np.all` comparison was vacuously true for empty results, so a regression that appends an empty array instead of skipping the candidate would have passed. The tests now assert the result length and compare candidates element-wise with `np.array_equal`, split into one parametrized test per function.
- Added a boundary test param for a single valid sample surrounded by NaNs, dropped the now-redundant `end_id > start_id` loop condition, and documented the drop behavior in both docstrings.

Thanks again, merging once CI is green!

Randomidous:
Amazing! Happy to help and thanks for the additions! 🚀

PR #1665, opened 2026-08-05
test: add doctest_tmp_cwd fixture for avoiding stateful doctests

cbueth:
## Description

tldr: Doctests were failing due to missing cleanup.

The `Dataset.apply` doctest downloads the ToyDataset into `toy_dataset/` in the repository root. Its expected output only matches a fresh download (`Downloading ... to toy_dataset...downloads...`). If a previous run left `toy_dataset/` behind, `Dataset.download()` prints `Verifying existing file: ...` / `Using existing verified file: ...` instead, causing a `DocTestFailure`.
This makes tests stateful and dependent on state left behind by previous runs. This PR adds an autouse fixture that removes the `toy_dataset/` directory before and after every test, so test outcomes no longer depend on prior state. No new dependencies are required.

## Implemented changes

- [x] Add autouse fixture `fixture_cleanup_test_artifacts` that removes `toy_dataset/` before and after every test
- [x] Register the fixture
- [x] Load the fixture globally via `addopts = "-p tests.fixtures.cleanup_fixtures"`. This way it also applies to `src/` doctests. Otherwise it does not apply usually.
- [x] Add the repository root (`.`) to `pythonpath` so the plugin module is importable at pytest startup

## How Has This Been Tested?

Locally
- [x] pytest src
- [x] pytest tests/unit/dataset/dataset_test.py tests/unit/dataset/resources_test.py
- [x] pre-commit

## Type of change

- Bug fix

## Context

This has been around for a long while.

## Checklist:


dkrako (commented):
Thanks for the your initiative @cbueth, the stateful `Dataset.apply` doctest was bugging me too.

I finalized the PR with a slightly different approach that fixes the cause instead of cleaning up afterwards: the two artifact-writing doctests (`Dataset.apply`, `Gaze.save`) now opt into a `doctest_tmp_cwd` fixture via a hidden `.. testsetup::` block, which chdirs into pytest's tmp dir, so nothing is written into the repo root in the first place and no global cleanup is needed. In detail:

- new `tests/doctest/doctest_fixtures.py` with the opt-in fixture, loaded via `-p doctest_fixtures` with `pythonpath = ['tests/doctest']` (instead of putting the repo root on `sys.path`)
- bumped the dev requirement to `pytest>=8.4.0`, the first version that applies `pythonpath` before `-p` plugin loading (pytest-dev/pytest#11118); with older pytest the plugin approach in this PR would fail at session startup
- added `sphinx.ext.doctest` to the docs config so the testsetup blocks stay hidden in the rendered docs
- documented the pattern in CONTRIBUTING.md and the contributing docs

Could you take a look at my finalization commit? If you are fine with it I will accept and merge.

cbueth (commented):
Thanks your your addition. I did not know about `.. testsetup::` yet. That is a cool solution. See this as accepted by me.

dkrako (approved):

PR #1668, opened 2026-08-09
fix: preserve zero-based word indices in reading measures

### LINKED ISSUE #1667: Reading Measure computation changes word index and could delete first word
## Current Behavior
The compute_reading_measures function sets all word indices -1 (thus word with original index=1 get index=0 and so on). The function also inserts a dummy word at index -1, thus deleting any word that originally had index=0.

## Expected Behavior
Word indices should stay constant dependent on the input and no words should be deleted.

Muhtasim-Munif-Fahim:
## Problem


dkrako (approved):
Thanks a lot for this contribution, @Muhtasim-Munif-Fahim, and welcome to pymovements! 🎉

Good catch: the unconditional `- 1` shift silently corrupted all word indices for zero-based AOI data, and your fix preserves the input indexing while keeping one-based data working. I also appreciate that you added a regression test that pins the zero-based case.

One heads-up rather than a change request: the `eq(0).any()` base detection is a heuristic, so it can misfire on edge cases (e.g. a one-based text where word 0 was never fixated is fine, but a zero-based AOI table whose first word got clipped would be misread as one-based). That's acceptable here, because the underlying problem is structural: the index shifting shouldn't exist in this function at all. This is addressed in #1671, a major rework of the reading measure code that removes the offset machinery entirely and preserves input indices by construction. Your test case carries over there unchanged, so your fix stays protected.

One more thing: feel free to add yourself to our [`CITATION.cff`](https://github.com/pymovements/pymovements/blob/main/CITATION.cff) so you're properly attributed in the next release. I'll hold the PR open a little longer to give you the chance; if you'd rather not, just react with a 👎 and I'll merge as is.

PR #1670, opened 2026-08-11
docs: remove OSF recommendation

saeub:
## Description

Starting in February 2027, OSF will [drop support for sharing public datasets](https://www.cos.io/blog/osf-changes-a-note-to-users). Existing public datasets will continue to be available.


dkrako (approved):
Good to know, thanks for the heads-up!

PR #1666, opened 2026-08-06
feat: add first pass fixation count to reading measures

othersideofthelake:
Added first pass fixation count (FPFC) to reading measure computation as well as to the explanatory table in the reading measure tutorial

#### Depends on

- [ ] #1641

PR #1671, opened 2026-08-11
feat: rewrite reading measures as declarative expression engine

dkrako:
## Description

This PR is a major rework of the reading measure computation. It replaces the imperative loop in `compute_reading_measures` with a declarative engine in which every reading measure and every fixation annotation is a polars expression, composed by a thin orchestrator. Grouping into independent reading sequences is no longer hardcoded anywhere: the caller chooses the partitioning, and the engine validates it.

The rework also removes the word index offset machinery that caused the zero-based index corruption fixed in #1668: word indices from the AOI table are now preserved by construction (zero-based, one-based, or any other start).

The engine is validated cell by cell against the published PoTeC reading measure files, which were produced by the reference implementation this code descends from.

## Implemented changes

- [x] all word-level reading measures are polars aggregation expressions (`total_fixation_count()`, `first_fixation_duration()`, ...); the caller supplies the `group_by(...).agg(...)`
- [x] all fixation annotations are polars expressions (`run_id()`, `prev_word_idx()`, `is_first_pass()`, ...); `annotate_fixations` is the consuming function that applies them in dependency order
- [x] regression path measures (`RPD_inc`, `RPD_exc`, `RBRT`) reformulated via running-maximum attribution (`rpd_target`): every fixation belongs to the regression path of exactly one word, removing the last Python loop
- [x] `is_first_pass` vectorized: first run of the word AND no higher word fixated before the run starts
- [x] robust `trial`/`page` group handling: `aois` accepts a dict keyed by trial, a plain AOI table broadcasts across fixation trials with a warning, and any other asymmetric or dtype-mismatched group column raises an informative error instead of silently producing all-zero measures
- [x] `group_columns`, `event_name`, and all input columns (`str | pl.Expr`) are parameters instead of hardcoded values; group columns are preserved in the output
- [x] word table dedupes inconsistent labels, drops null word indices; fixation word indices without an AOI entry (e.g. `-1` sentinels) are excluded from sequence logic
- [x] new measures: within-word landing position (`LP`, one-based as in PoTeC, 0 for never-fixated words, null when the position is not determinable from the data) and `skipped` (1 if the word was never fixated, `TFC == 0`)
- [x] `FPFC` (#1666) and the non-AOI fixation ratios `NAFCR`/`NAFDR` (#1633) are ported to expression form; the ratios are sequence-level standalone expressions, not part of the per-word default pipeline
- [x] removed the spurious `TRC_out`/negative `SL_out` on each trial's last fixated word and the `SL_in` start sentinel artifact of the old implementation
- [x] removed `build_word_level_table` and `mark_skipped_tokens` (redundant with the engine output); count columns are `u64`
- [x] shared `as_expr` helper in `pymovements/_utils/_expressions.py` for `str | pl.Expr` inputs
- [x] the `compute_reading_measures` docstring documents every output column in output order as a table, cross-referencing each measure's expression function
- [x] the reading measures tutorial links to that API reference instead of duplicating the measure table, and gained a section computing measures for several stimuli at once via `group_columns` and a per-stimulus `aois` dict

## How Has This Been Tested?

- [x] 125 unit tests in `tests/unit/measure/reading/` covering the engine, all expressions, group handling errors, and edge cases, including pins for every known divergence from the PoTeC reference implementation with the reference-expected values as commented-out lines (`pytest tests/unit/measure/reading/`)
- [x] differential fuzzing: 3000 randomized fixation sequences compared against the previous imperative implementation, zero divergence outside the documented artifact fixes
- [x] gold-standard validation against PoTeC: all 900 published reading measure files reproduced exactly (142,125 words, ~2.7M measure cells, 0 mismatches), with only the documented reference implementation artifacts masked; whether these end-of-sequence behaviors of the reference were deliberate could not be reconstructed (clarified with Deborah), so pymovements keeps its consistent behavior with every divergence pinned by tests; runtime for the full sweep dropped from over five minutes to under one minute
- [x] new functional test `tests/functional/reading_measures_potec_test.py` (marked `network`) samples every 60th PoTeC file and re-validates against the published measures, including `LP` cell by cell with word start characters from PoTeC's word limits resource (~30 s including download)

## Type of change

- Bug fix
- New functionality
- Performance improvement

#### related issues:
- #1667
- #1668 (the index preservation fix, superseded structurally; the regression test carries over unchanged)

## Follow-up work (out of scope here)

- measure registry with user-facing measure selection: one entry per measure (expression factory, grouping key, fill value) replacing the hardcoded lists in `processing.py`, exposed through the `measure_reading` API
- expose `measure_reading` on `Gaze` and/or `Events`, not only on `Dataset`
- streamline multi-stimulus preparation: the tutorial's "several stimuli at once" section needs a hand-written loop because `map_to_aois` is single-stimulus (one `TextStimulus` per call), AOI/text stimulus loading is manual `fileinfo` plumbing, and the per-group `aois` dict must be assembled by hand. Group-aware mapping (e.g. `map_to_aois` accepting a dict of `TextStimulus` keyed by group value), dataset-level text stimulus accessors, and `measure_reading` on `Gaze`/`Events` (see above) would collapse the section to a few lines
- CI handling of the `network` marker: deselect network tests in the default `py*` tox envs and run them in a scheduled job (until then, the PoTeC functional test downloads ~47 MB per CI job)
- character-level output tables (per-character measures) on top of the now char-aware engine
- let the PoTeC functional test consume the PoTeC dataset definition instead of defining its own `WebSource`s, so URLs and checksums live in one place. This requires adding the merged scanpaths as a resource to `potec.yaml` (e.g. `precomputed_scanpaths`; the merged reading measures already exist as `precomputed_reading_measures`) and a way to download selected resources only, since `Dataset.download()` currently always fetches everything including the raw gaze archive

## Checklist:


docs/source/tutorials/reading-measures.ipynb
```diff
@@ -189,34 +190,83 @@
...
+   "id": "rm-groups-build",
+   "metadata": {},
+   "execution_count": null,
+   "outputs": [],
+   "source": [
+    "stimuli = ['goldfish-pos.text.0', 'goldfish-pos.text.1']\n",
+    "\n",
+    "multi_fixations = []\n",
+    "aois = {}\n",
+    "for stim in stimuli:\n",
```
dkrako, line 219: this block is a bit extensive currently but that's due to our current limitations handling stimulus collections. this will be improved in the future but out of scope of this PR.

docs/source/tutorials/reading-measures.ipynb
```diff
@@ -189,34 +190,83 @@
    "id": "8af2afbc-10ec-41b5-a7ec-0c3ef676997d",
    "metadata": {},
    "source": [
-    "The resulting DataFrame contains one row per word (Area of Interest, AOI), along with a range of eye-tracking reading measures.\n",
-    "\n",
-    "#### Description of computed reading measures\n",
-    "\n",
-    "| Column | Description |\n",
```
dkrako, line 196: I moved the table to the compute_reading_measures docstring

src/pymovements/measure/reading/processing.py
```diff
@@ -20,160 +20,733 @@
...
-            dummy_fixation_dict[col] = ['']
-        elif col == word_index_column:
-            dummy_fixation_dict[col] = [index_offset - 1]
-        else:
-            dummy_fixation_dict[col] = [0]
-    dummy_fixation = pl.DataFrame(
-        dummy_fixation_dict,
-        schema=fixations.schema,
+    if group_columns is None:
+        group_columns = ['trial', 'page']
```
cbueth: This default is also picked up by `Dataset.measure_reading`, which doesn't set `group_columns`. When fixations carry `trial`/`page` but the AOI table is a flat frame, callers get silent per-trial grouping and a broadcast warning. Consider having `Dataset.measure_reading` pass `group_columns=[]` to keep its historical single-sequence behavior, or document the change.

src/pymovements/measure/reading/measures.py
```diff
@@ -17,575 +17,535 @@
...
-        DataFrame with columns ``trial``, ``page``, ``word_idx``, and
-        ``LP``.
+    pl.Expr
+        Aggregation expression producing the ``LP`` column.
     """
-    return fixations.group_by(['trial', 'page', 'word_idx']).agg(
-        pl.col('char_idx').sort_by('onset').first().alias('LP'),
-    )
+    char_idx_expr = as_expr(char_idx)
+    return (char_idx_expr.first() + 1).alias('LP')
```
cbueth, line 283: `.first()` here depends on the row order inside the `group_by` window matching onset order. It works as `annotate_fixations` sort, but this might break easily in the future. The docstring notes it must be onset-sorted, but aggregation does not enforce it.

src/pymovements/measure/reading/processing.py
```diff
@@ -20,160 +20,733 @@
...
+    table = (
+        words
+        .join(word_level, on=on, how='left', nulls_equal=True)
+        .join(regression_paths, on=on, how='left', nulls_equal=True)
+    )
+
+    if 'LP' in table.columns:
+        # Landing position within the word, one-based; unfixated words are filled 0 like every
+        # other measure. The zero-fill is keyed on the join miss (TFC is null exactly for
+        # unfixated words at this point, before the shared fill below), so a fixated word whose
```
cbueth: `TFC.is_null()` guard is correct only because the `fill_null(0)` on `joined_measure_columns` runs after this block. If the fill order ever changes, unfixated words would silently get a wrong LP.

The unfixated mask could be calculated once to get rid of the hidden ordering dependency.

src/pymovements/measure/reading/annotation.py
```diff
@@ -17,277 +17,306 @@
...
-                    entered_from_left
-                    and no_higher_word_seen
-                    and first_time_entering_word
-                )
+    first_run_of_word = (
+        run_id_expr == run_id_expr.min().over(group_columns + [word_idx_expr])
+    )
 
-                words_ever_entered.add(w)
+    return (no_higher_word_seen & first_run_of_word).alias('is_first_pass')
```
cbueth, line 296: It seems to have the right logic, but the docstring
"The no-higher-word condition is constant across a run (within a run the running maximum either already exceeded the word or is the word itself), so it needs no run-level broadcast."
is a bit brief. Maybe an example would make the case of refixating on the rightmost word more easy to understand.

cbueth (approved):
This PR was various changes, is well tested and does not change the API.
I suggest this to be merged, but think that the `group_columns` default can be reconsidered before.

src/pymovements/measure/reading/measures.py
dkrako: The dependency is real, but it is rather a package-wide idiom: FFD and SL_in rely on the same assumption, and the order is produced centrally by the sort in `annotate_fixations`. Instead of enforcing order in every aggregation, I pinned the contract in three places:

- a regression test feeds a scrambled fixation frame through `compute_reading_measures` and asserts identical output to onset-ordered input (d3b06cdd263b05eb94a3088557d1972d24f1e83b),
- `annotate_fixations` warns when onsets are not sorted within a sequence and suggests checking group columns (328accdf4fa7ae3b7ff332ca8125b71f50435790),
- each function that requires sorted onsets now has a warning block in  the docs (0327c9c943e4e50c30e79519cbc8de94d3a582fe).

A general temporal-order validation helper on `Events` is planned as a follow-up, I will write the issue.

src/pymovements/measure/reading/processing.py
dkrako: Good catch, the hidden ordering dependency is gone, though via a slightly different route than a transient mask: the zero-fill and the `skipped` derivation are moved above the LP block, and LP is keyed on `skipped == 1` (7e1b7adb28abc75298959f6dbb66c1c8857b2649). Since `skipped` is a real output column anyway, no helper column is needed, and LP stays excluded from the fill so its null states survive.

We will discuss refactoring the `skipped` out as additional measure, I will write an issue for that.

src/pymovements/measure/reading/annotation.py
dkrako: Agreed, I think all of the annotation and measure functions should have at least one example. To keep the re-review surface of this PR small, it lands in a small docs follow-up PR right after this one, together with doctests for the remaining annotation and measure functions.

src/pymovements/measure/reading/processing.py
dkrako: Agreed, and I went one step further: `compute_reading_measures` itself now defaults to a single reading sequence as well (dbcf07fb0ade46023b45011263050569be8c657c), since an implicit `['trial', 'page']` default was inconsistent with every other place we work with default lists.

Deriving an informed default from `Events.trial_columns` plus a grouping attribute on `TextStimulus` will be worked out soon.

src/pymovements/measure/reading/measures.py
cbueth: fancy, thanks

cbueth (approved):

PR #1674, opened 2026-08-12
feat: allow overriding trial_columns in measure_events_ratio

### LINKED ISSUE #1673: `measure_events_ratio`: allow overriding `trial_columns` for session-level aggregation
## Description of the problem
`Gaze.measure_events_ratio()` always passes `self.trial_columns` to `transforms.events2timeratio()`. When trial columns are set, the returned expression is per-row (one value per sample, matching each sample's trial ratio) rather than a session-level scalar. To compute a single session-wide ratio, the caller has two suboptimal options:
1. Use `gaze.samples.select(expr).item()`, which crashes with `ValueError` because the DataFrame has one row per sample, not one row total
2. `gaze.samples.select(expr.first()).item()` gives the ratio only of the first trial, not the session
There is no way to tell `measure_events_ratio` "ignore trial columns, I want the aggregate over the whole session."

## Description of a solution
Add a `trial_columns` parameter to `Gaze.measure_events_ratio()` that overrides `self.trial_columns`:
```python
def measure_events_ratio(
    self,
    name: str,
    ...,
    trial_columns: list[str] | None = _MISSING,
) -> pl.Expr:
```
- `trial_columns=None` -> session-level scalar (no trial grouping)
- `trial_columns=["trial"]` -> overwrite column to use
- `trial_columns` not passed -> use `self.trial_columns` (default, current behavior)
The parameter is forwarded to `transforms.events2timeratio()`, which already supports it.

- Calling `gaze.measure_events_ratio("blink", trial_columns=None)` returns an expression that evaluates to a single session-level scalar. NOT weighted by trials.
- Calling without `trial_columns` preserves existing behavior
- `.item()` on the resulting scalar DataFrame works without error
- Updated docstring examples showing both session-level and per-trial usage

## Context
Encountered in the MultiplEYE preprocessing pipeline. Some session crashed because the expression returned 3 million rows via `.select()` and `.item()` failed. The user just did not blink, apparently? We work around it by calling `transforms.events2timeratio()` directly with `trial_columns=None`, but the Gaze method should support this directly.

cbueth:
## Description

`Gaze.measure_events_ratio()` always forwarded `self.trial_columns` to `transforms.events2timeratio()`. When trial columns are set, the returned expression is one value per sample, so there was no way to obtain a single session-level ratio without calling the transform directly. This adds a `trial_columns` parameter to override the gaze's trial columns.

## Implemented changes

- [x] Add keyword-only `trial_columns` parameter to `Gaze.measure_events_ratio()` (defaults to `self.trial_columns` when not passed)
- [x] `trial_columns=None` now yields a single session-level scalar, so `.select(...).item()` works without error
- [x] Passing a list of columns overrides the gaze's trial columns
- [x] Docstring examples

## How Has This Been Tested?

- [x] unit tests for session-level, default, and override behavior
- [x] Doctests pass

## Type of change

- New functionality

## Context

Resolves #1673

#### related issues:
- #1673

#### requires:
- [ ] -

## Checklist:


tests/unit/measure/samples/measures/event_ratio_test.py
```diff
@@ -323,3 +323,182 @@ def test_measure_events_ratio_sampling_rate_fallback(
...
+    def test_event_ratio_trial_columns_none(
+        self,
+        samples,
+        gaze_trial_columns,
+        events_data,
+        expected_ratio,
+    ):
+        """trial_columns=None yields a single session-level scalar and item() works."""
+        events = pm.Events(pl.DataFrame(events_data))
+        events.trial_columns = gaze_trial_columns
```
dkrako: I prefer just passing the `trial_columns` directly to the `Events()` constructor

dkrako (changes requested):
Thanks a lot for your changes @cbueth, this is a well-tested implementation and the doctests are a nice touch.

Before merging I'd like to change one design point because I'd like to drop the `_MISSING` sentinel default.

The sentinel exists because the PR makes `None` mean "no grouping", while elsewhere in the API (e.g. `Gaze.split()`) `None` means "fall back to `self.trial_columns`". I'd rather keep that convention consistent:

- `trial_columns=None` (default): fall back to the gaze's `trial_columns`, as `split()` does
- `trial_columns=[]`: session-level scalar (your use case from #1673, "group by zero columns" gives exactly one group)
- `trial_columns=['trial']`: override

An empty list already flows through both the method body and `events2timeratio` as "no grouping" (both test truthiness), so this needs no extra logic, the signature becomes a plain `list[str] | None = None`. I know this deviates from the wording in #1673, but since you authored both I hope you're fine with adjusting it here. The doctests and the `trial_columns=None` tests would switch to `[]` accordingly.

We can discuss this further in our next meeting.

dkrako (approved):
Excellent, thank you so much for your changes!

PR #1688, opened 2026-08-19
feat: add 'first' and 'last' as methods to location measure

saeub:
## Description

Knowing the start and end of a saccade (or maybe blink) could be useful. This basically uses the first and last samples in the event (assuming that the samples are already ordered by timestamp -- is this an assumption we make generally?).

## Implemented changes

- [x] Add `first` and `last` as methods for location measure.

## How Has This Been Tested?

- [x] Add `first` and `last` methods to existing unit test parameters

## Type of change

- New functionality

## Checklist:

### DIFF
--- a/src/pymovements/measure/samples/measures.py
+++ b/src/pymovements/measure/samples/measures.py
@@ -266,11 +266,15 @@ def location(
         \text{Location} = \text{median} \left(\text{position}_1, \ldots,
          \text{position}_n \right)
 
+    For methods ``start`` and ``end`` the location is the position of the first and last sample in
+    the event, respectively.
+
 
     Parameters
     ----------
     method: str
-        The centroid method to be used for calculation. Supported methods are ``mean``, ``median``.
+        The centroid method to be used for calculation. Supported methods are ``mean``, ``median``,
+        ``first``, and ``last``.
         (default: 'mean')
     position_column: str
         The column name of the position tuples. (default: 'position')
@@ -288,10 +292,10 @@ def location(
     ValueError
         If method is not one of the supported methods.
     """
-    if method not in {'mean', 'median'}:
+    if method not in {'mean', 'median', 'first', 'last'}:
         raise ValueError(
             f"Method '{method}' not supported. "
-            f"Please choose one of the following: ['mean', 'median'].",
+            f"Please choose one of the following: ['mean', 'median', 'first', 'last'].",
         )
 
     component_expressions = []
@@ -304,8 +308,12 @@ def location(
 
         if method == 'mean':
             expression_component = position_component.mean()
-        else:  # by exclusion this must be median
+        elif method == 'median':
             expression_component = position_component.median()
+        elif method == 'first':
+            expression_component = position_component.first()
+        else:  # by exclusion this must be last
+            expression_component = position_component.last()
 
         component_expressions.append(expression_component)
 
--- a/tests/unit/measure/samples/measures/location_test.py
+++ b/tests/unit/measure/samples/measures/location_test.py
@@ -82,6 +82,32 @@ def test_location_exceptions(init_kwargs, exception, message):
             ),
             id='position_three_samples_median',
         ),
+
+        pytest.param(
+            {'method': 'first'},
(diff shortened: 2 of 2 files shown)


dkrako (changes requested):
Thank you four initiative @saeub, start and end positions of saccades are definitely a very useful addition!

One design request before merge though: I would prefer this as two separate measures (`location_start` and `location_end`) instead of new `method` values on `location`:

There are really two orthogonal axes here: *what* is measured (the event's central location vs. its start and end position) and *how* it is estimated (`mean` and `median` are two estimators of the same central location; a start position could likewise be estimated in different ways). A flat `method` enum collapses both axes into one list, while separate measures keep the *what* in the measure name and leave `method` free for the *how*.

Concretely on that *how* axis: each measure can grow its own `method` argument later, e.g. estimating the start position from the mean of the first n samples or via clustering, without overloading `location` further.

One more thought for the future, out of scope for this PR: for saccades, the start and end position could eventually be estimated from the whole adjacent fixation window (preceding fixation for start, following for end) instead of a single boundary sample, which would be much more robust to noise. That needs cross-event context in the processor, so it is clearly out of scope here, but it is another reason to have dedicated measures with their own `method` argument. Alternatively it might be an alternative method that operates exclusively on event data.

On your question about sample ordering: yes, pymovements assumes samples are in recording order throughout the event pipeline (`disposition` relies on this in the same way), so your change adds no new assumption. However it might make sense to put some additional guards for that in the future, for example in the `EventSampleProcessor`.

One smaller thing that carries over to the new functions: null handling should match `mean`/`median`: those skip nulls, while the current implementation returns null if the first or last sample happens to be null. Please use the first/last valid (non-null) sample instead, document that in the docstring, and pin it with a test that has a leading/trailing null sample.

Happy to merge once this is restructured, thanks again!

saeub:
@dkrako What do you think about naming them `location_onset` and `location_offset` (or `onset_location` and `offset_location`) instead of "start" and "end", to keep it consistent with the event column names?

> the start and end position could eventually be estimated from the whole adjacent fixation window (preceding fixation for start, following for end) instead of a single boundary sample

That would be nice, but only works if fixations have been detected. If the user is only interested in saccades, being able to quickly plot their start and end locations is still helpful, I think.

dkrako:
Thanks @saeub, and consistency with the event columns is a fair argument. I will need to think about this a bit more until I can really decide. Let me share my current thinking, I'm happy to hear your take.

I'd rather keep `onset` and `offset` reserved for the time domain. They name instants in time, and in the positional domain offset actually means something like a distance between two locations.

There is also a what/how split similar to the one in my first review: `location_start` names the quantity we're after (where the event started), while `location_onset` names the current mechanism (the value at the onset instant). Once we add more robust estimation, e.g. estimating a saccade's start position from the preceding fixation window, the result is no longer the location at the onset timestamp at all (and it might not even be a pure sample measure anymore).

A mechanism-free name lets the estimation evolve behind a `method` argument (or a successor implementation) without renaming, whereas `location_onset`/`location_offset` would either become misnomers or force a second, near-synonymous measure pair later. `onset`/`offset` as `method` values would directly translate to using the location at that exact time sample. If we later provide a method that uses the location of the first/last N samples, that is no longer the onset in the strict time-domain sense.

On your second point: fully agreed, the boundary-sample version is useful on its own, saccades without fixation detection being exactly the case. The fixation-window estimation is future work and doesn't block anything here.

saeub:
Yeah, naming things is hard... :sweat_smile:

I understand the point that "offset" could be reserved for a single time point/sample. But I think that "start" and "end" still refer to the time domain -- we want to know the location at the **time** of the start of the event.

- Naming it `location_start` would make it more explicit that it is not necessarily the exact location of the sample at the `offset` timestamp. (But note that even if we take the first sample in the event time span, this is not necessarily the sample at that exact timestamp -- for example, there might be a missing sample there.)
- Naming it `location_onset` would make it clearer to the user that it refers to the time domain. ("start" could also be understood locally -- e.g., the top-most or left-most location of the event)

Issue #522, opened 2023-09-11
Infer the correct number of components for empty dataframes.


dkrako:
From a polars Dataframe like `pl.DataFrame(schema={'pixel': pl.List(pl.Float64)})` we cannot infer the number of components,
as the number of components is inferred from the list lengths.

This can probably only be solved by resolving #453 

The priority for fixing this is very low, as there are probably close to no use cases with empty dataframe initialzitations.

dkrako:
fixed in #521

PR #1689, opened 2026-08-19
feat: align Events.unnest() with Gaze.unnest()

saeub:
## Description

`Events.unnest()` did not support the same arguments as `Gaze.unnest()` and hardcoded the columns to be unnested. This PR reuses the code used in `Gaze.unnest()`.

## Implemented changes

- [x] Use the same implementation for `Events.unnest()` as for `Gaze.unnest()`
- [x] Move `pm.gaze._utils._column_nesting` to `pm._utils._column_nesting` to avoid circular import
- [x] Ignore `null` values when inferring number of components to allow unnesting event frames with multiple event types. (Necessary because event properties that were only computed for one event type are `null` for other event types.)

## How Has This Been Tested?

Existing tests involving `Events.unnest()` pass. Added a test with multiple event properties that include `null` values.

## Type of change

- New functionality

## Checklist:


dkrako (approved):
Great, thank you so much @saeub! Clean, well-scoped PR.

On that behavior change: `events.unnest()` previously only ever unnested `location`, silently doing nothing when it was absent and ignoring all other list columns. Now the no-argument call unnests every list column and only warns when the frame has no list columns at all. We ship this as-is without a deprecation cycle, since the old method was very rudimentary.

I added a few small finalization commits on top:

- Component inference used to fail on columns without inferable values with a confusing "number of components inconsistent" error containing an empty Series repr. It now distinguishes the two cases and raises a clear `cannot infer number of components in all-null column '<name>'` or `cannot infer number of components in empty column '<name>'` for length-0 frames. Both messages are pinned by parametrized tests through both public surfaces, `Events.unnest()` and `Gaze.unnest()`, and documented in the `Raises` sections. Empty frames succeeding at unnesting altogether is tracked separately in #522.
- Test additions: the new `input_columns`/`output_suffixes`/`output_columns` arguments are now exercised through `Events.unnest()` with explicit expected columns, and the stale `test_unnest_location_absent_is_noop` was renamed to `test_unnest_no_list_columns_warns` to match the warning behavior.

Thanks again, merging once CI is green!

PR #1690, opened 2026-08-19
feat: specify column names in `compute_event_properties()`

### LINKED ISSUE #1043: specify target columns for event properties
## Description of the problem
Currently columns that are created by `compute_event_properties()` will always match the names of the respective property.

For example, if someone wants to compare mean and median dispersion, this is not possible.

## Description of a solution
Instead it should be possible to specify the target columns for the calculated event property.

## Context

Related issues:

- #1041 
- #1042

saeub:
## Description

Adds an argument `column_names` to `Gaze.compute_event_properties()` through which the user can specify the names of the columns that will be added to `Event.frame`. This allows the user to compute the same property several times (e.g., with different parameters, or on different events) by giving each a different column name. When the argumen tis not provided, the default behavior is still the same.

## Implemented changes

- [x] Add `column_names` argument to `Gaze.compute_event_properties()`
- [x] Add `column_names` argument to `EventSamplesProcessor.__init__()`
- [ ] Add `column_names` argument to `EventProcessor.__init__()`

## How Has This Been Tested?

- [ ] TODO

## Type of change

- New functionality

## Context

Resolves #1043

## Checklist:

### DIFF
--- a/src/pymovements/gaze/gaze.py
+++ b/src/pymovements/gaze/gaze.py
@@ -1298,6 +1298,7 @@ class Gaze:
             self,
             event_properties: str | tuple[str, dict[str, Any]]
             | list[str | tuple[str, dict[str, Any]]],
+            column_names: str | list[str] | None = None,
             name: str | None = None,
     ) -> None:
         """Calculate event properties for given events.
@@ -1309,6 +1310,9 @@ class Gaze:
         ----------
         event_properties: str | tuple[str, dict[str, Any]] | list[str | tuple[str, dict[str, Any]]]
             The event properties to compute.
+        column_names: str | list[str] | None
+            The name(s) of the column(s) to be added to the event dataframe. If None, columns will
+            be named after the event properties. (default: None)
         name: str | None
             Process only events that match the name. (default: None)
 
@@ -1319,6 +1323,9 @@ class Gaze:
             :ref:`event-measures` for an overview of supported measures.
         RuntimeError
             If specified event name ``name`` is missing from ``events``.
+        ValueError
+            If ``column_names`` has a different length than ``event_properties``, or if there are
+            duplicate column names.
         """
         if len(self.events) == 0:
             warn(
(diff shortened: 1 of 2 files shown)


dkrako (changes requested):
Thank you so much for taking this on @saeub, and good timing to discuss direction before you invest in tests!

Reading the implementation, I would prefer a different API for this: an `alias` or `output_name` parameter on each measure function itself, instead of the parallel `column_names` list. The name then travels with the measure spec through the existing kwargs mechanism:

```python
gaze.compute_event_properties([
    ('dispersion', {'method': 'mean', 'output_name': 'dispersion_mean'}),
    ('dispersion', {'method': 'median', 'output_name': 'dispersion_median'}),
])
```

Reasons:

- No parallel list `column_names` list to keep positionally in sync with `event_properties` that has the risk of silent misalignment on reorder
- `SampleMeasureLibrary.get(name)(**kwargs)` already forwards kwargs, so the feature reaches `Dataset.compute_event_properties` and direct functional calls automatically, and custom registered measures follow the same convention.

The cost is one `output_name` parameter plus docstring entry per measure function. Your duplicate-name check is worth keeping, just moved to operate on `expr.meta.output_name()`.

Happy to hear your thoughts before you continue, and thanks again for picking up #1043!

saeub:
The `data_loss` measure aliases its output to `data_loss_ratio` instead of `data_loss` -- is this intended?
https://github.com/pymovements/pymovements/blob/dc3dd9c4ab9de923144323a95c8e8f60b2576e3c/src/pymovements/measure/samples/measures.py#L886

Right now my code assumes that measures without a `output_name` argument should be named like the measure itself (overriding the alias given in the measure function itself). Maybe we could even remove the redundant aliasing within the measure function definition?

saeub:
Conclusion from meeting discussion: Either put the `output_name` logic in the measure registry, or add an argument to every measure function definition. Don't put the logic in the `Processor` class.

Issue #1424, opened 2025-12-18
retry download before failing


dkrako:
## Description of the problem

Currently a single server time out may lead to a failed download.

This is also annoying in our CI, as github workers sometimes have connectivity issues and then tests fail randomly.

## Description of a solution

We could handle a failed download more gracefully. Instead of instantly raising an exception, a number of retries could lead to a more robust download performance in more difficult conditions.

A simple wait time between the retries would probably help with connectivity issues.

## Minimum acceptance criteria
- [ ] add retry argument and logic to download function

PR #1701, opened 2026-08-26
ci: retry flaky network tests with pytest-rerunfailures

dkrako:
## Description

Tests marked with `network` download files from real remote hosts (mostly OSF). On every PR commit they run in three parallel matrix jobs, so each push fires near-simultaneous download bursts at the same hosts. Transient connectivity issues and rate limiting then fail CI for reasons unrelated to the change under review, most recently the PoTeC reference test introduced in #1671.

This PR makes network tests retry on failure with a delay, giving throttling hosts room to recover. It is the minimal intervention: no coverage loss, no workflow changes, no contributor-facing policy. If network tests still fail after retries, stronger measures (skipping network tests on PR commits by default with a label/dispatch opt-in, serving download-logic test fixtures from a local HTTP server) can be evaluated later based on CI logs: reruns show up as `RERUN` in the pytest summary.

The retry policy also covers doctests: the `Dataset.apply` docstring example really downloads ToyDataset on every run (since #1665 made it runnable). Doctests cannot carry pytest markers, so the downloading example is declared explicitly with the `REMOTE_DATA` doctest option flag from `pytest-doctestplus`, and the plugin maps that declaration to the `network` marker at collection time. As a side effect, `-m "not network"` now truly excludes all network access including the downloading doctest (previously the documented offline run still downloaded ToyDataset). `--remote-data=any` is set in addopts, so all runs (local and CI) keep executing the downloading example as before; going offline is done via `-m "not network"`. Explicitly passing `--remote-data=none` would skip only the flagged `download()` example and break the dependent examples, which is accepted as an unsupported invocation.

Adds `pytest-rerunfailures`, `pytest-doctestplus`, and `pytest-remotedata` to the `dev` extra.

## Implemented changes

- [x] add a local pytest plugin `tests/plugins/rerun_network.py`, registered via the existing `pytest_plugins` list in `tests/conftest.py`, whose `pytest_collection_modifyitems` hook applies `flaky(reruns=2, reruns_delay=30)` to all `network`-marked tests
- [x] the same hook adds the `network` marker to doctests that declare `REMOTE_DATA` examples, so they are retried and filterable via `-m` like any other network test
- [x] flag the `dataset.download()` example in `Dataset.apply`'s docstring with `# doctest: +REMOTE_DATA` as the explicit network declaration
- [x] enable `pytest-doctestplus` via `--doctest-plus` in addopts (it was previously installed in some environments but inactive) and set `--remote-data=any` in addopts so all runs keep executing the downloading example
- [x] add `pytest-rerunfailures>=16.0`, `pytest-doctestplus>=1.7.0`, and `pytest-remotedata>=0.4.1` to the `dev` dependencies (`pytest-remotedata` provides the `--remote-data` option that doctestplus reads)
- [x] add a marker meta-test `tests/plugins/rerun_network_test.py` guarding that `network` tests receive the retry marker and others do not
- [x] extend the `name-tests-test` pre-commit exclude for non-test modules in `tests/plugins/` and ignore `tests/plugins` in the benchmark tox env, both consistent with the existing `tests/fixtures/` handling

## How Has This Been Tested?

Verified locally, partly with a temporary test file (not committed):

- [x] a failing `network`-marked test is rerun twice with a 30 second delay and reported as `RERUN` in the pytest summary
- [x] a download failure inside a module-scoped fixture re-executes the fixture on rerun instead of re-raising the cached error (relevant for the PoTeC test in #1671, which downloads inside a module-scoped fixture)
- [x] the `Dataset.apply` doctest is selected by `-m network` and deselected by `-m "not network"`
- [x] with the default configuration the `Dataset.apply` doctest downloads ToyDataset and passes fully
- [x] all 70 src doctests pass with `--doctest-plus` enabled
- [x] the meta-tests in `tests/plugins/rerun_network_test.py` pass; non-network tests are unaffected

CI wall time is unaffected in the happy path; the 30 second delay only applies when a network test actually fails.

## Type of change

- Performance improvement (CI robustness)

## Context

The `pytest_collection_modifyitems` hook is the mechanism the pytest documentation itself uses for attaching marker-conditional behavior suite-wide, rather than a workaround: see [Automatically adding markers based on test names](https://docs.pytest.org/en/stable/example/markers.html#automatically-adding-markers-based-on-test-names) and [Control skipping of tests according to command line option](https://docs.pytest.org/en/stable/example/simple.html#control-skipping-of-tests-according-to-command-line-option), both implementing the same shape of hook. Registering it from `tests/plugins/` mirrors the existing `tests/fixtures/` pattern and keeps `tests/conftest.py` a pure plugin registry. Scoping the reruns via the pytest invocation instead was considered and rejected: `pytest-rerunfailures` offers no marker-based CLI scoping, so it would require either global reruns (masking real failures suite-wide), error-message regex matching via `--only-rerun` (fragile), or split pytest invocations (report and coverage plumbing).

One caveat: partial runs that collect only `src/` (e.g. `pytest src/pymovements/dataset/dataset.py`) do not load `tests/conftest.py`, so the plugin and its retry behavior are absent there; CI and tox always collect both `tests` and `src`.

#### related issues:
- #1671
- #1424

## Checklist:


pyproject.toml
```diff
@@ -157,7 +160,7 @@ skip-checking-raises = true
 treat-property-methods-as-class-attributes = true
 
 [tool.pytest.ini_options]
-addopts = "-ra -l --tb=long --doctest-modules --import-mode=importlib -p doctest_fixtures"
+addopts = "-ra -l --tb=long --doctest-modules --doctest-plus --remote-data=any --import-mode=importlib -p doctest_fixtures"
```
cbueth, line 163: So `--remote-data=any` makes every network/remote-data test (including `Dataset.download()` doctest) run in all default runs, not just scheduled CI, right? With the `network` marker one cannot deselect, I think. Every CI does the download then. One could have one CI for network and one with `not network`. But we already have so many different CI jobs.

tests/plugins/rerun_network.py
```diff
@@ -0,0 +1,42 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+"""Retry network tests to absorb transient connectivity issues and rate limiting."""
+from __future__ import annotations
+
+import doctest
+
+import pytest
+
+REMOTE_DATA = doctest.register_optionflag('REMOTE_DATA')
```
cbueth: This re-registers an optionflag that `pytest-doctestplus` already owns. So this does not change the outcome, but maybe across versions it might be different. The flag could be imported from `pytest_doctestplus` directly.

.pre-commit-config.yaml
```diff
@@ -93,7 +93,7 @@ repos:
       - id: double-quote-string-fixer
       - id: end-of-file-fixer
       - id: name-tests-test
-        exclude: .*_fixtures\.py
+        exclude: .*_fixtures\.py|tests/plugins/.*(?<!_test)\.py
```
cbueth: The `exclude` regex works, but the `(?<!_test)` lookbehind is not so obvious to understand I suggest add an explanatory comment above: What is included and what not.

cbueth (approved):
I can accept this PR, again with some grains of salt that you can add by your likes.

About enabling `--remote-data=any` in the default I am not so sure. This would make it always-on.

tests/plugins/rerun_network.py
dkrako: Good point, the import makes the coupling explicit. applied in 802c754e

Note that `output_checker` is not documented public API, but the meta-test in `tests/plugins/rerun_network_test.py` would catch a moved import immediately.

.pre-commit-config.yaml
dkrako: Revisiting this, the lookbehind was not just non-obvious but redundant: the check passes exactly the files ending in `_test.py` which would trivially pass.

A plain `tests/plugins/` exclusion behaves identically in every case, including mis-named tests. So instead of explaining the regex I dropped it in 8e99710d and added a one-line comment stating the intent.

pyproject.toml
dkrako: Yes, all default runs still execute the download doctest, but that is our package-wide default not touched by this PR. `--remote-data` is only an execute-or-skip switch for the statements flagged with `+REMOTE_DATA`: `any` runs them, `none` skips them.

That is why the default `--remote-data=none` is not an option here: it skips only the flagged `>>> dataset.download()` statement but not the later statements in the doctest that depend on the downloaded dataset, so the doctest would fail in every default run.

Deselecting does work though: our plugin maps doctests containing `+REMOTE_DATA` statements to the `network` marker at collection time, so `-m "not network"` now really excludes the download (before this PR the `not network` run still downloaded ToyDataset in the doctest).

I agree splitting CI into a network and a not-network job is worth discussing, but also agree that more CI jobs will add clutter that may be unnecessary. Let's revisit it once we see how the retries do in practice.

PR #1712, opened 2026-09-01
tests: reach 100% test coverage

dkrako:
## Description

Codecov's target is 100% for both `project` and `patch`, but `main` sat at 99.13% (54 missed lines + 8 partial branches across 10 files). This closes every gap so coverage reaches a true 100%.

The largest gap (both `_scripts` modules at 0%) turned out to be a coverage **attribution** bug, not missing tests: in CI the wheel-installed copy is measured under `.tox/.../site-packages/...` and remapped via `[tool.coverage.paths]` whose
canonical entry was the literal `{toxworkdir}/src/pymovements` (a path that does not exist). Pointing the canonical path at `src/pymovements` makes the remap resolve to real files.

The remaining gaps were closed with targeted tests, or by removing genuinely dead code.

## Implemented changes

- [x] Fix `[tool.coverage.paths]` canonical path (`{toxworkdir}/src/pymovements` -> `src/pymovements`) so wheel-install coverage is attributed to the repo files (fixes the `_scripts` 0% on Codecov)
- [x] Remove dead code:  `Screen.__post_init__`, unreachable guard in `tsplot`, always-true guard in `Gaze.__str__`, and the obsolete boolean-repr branch in `_html` (its only trigger, `_HasResourcesIndexer`, was removed in #1621)
- [x] Add tests: `from_ipc` `add_columns` / `column_schema_overrides`, `_column_nesting` inconsistent-component-count error, `download(extract=False)`, `Gaze.schema`, and `idt` detection without a `position` column
- [x] Split `test_heatmap_deprecated_parameters` by file dependency and add `test_heatmap_deprecated_add_stimulus_renders_stimulus`

## Checklist:


cbueth (changes requested):
Thank you for adding the coverage.
https://app.codecov.io/gh/pymovements/pymovements/tree/tests%2Ffull-coverage states 99.79%

I see that the only line is https://app.codecov.io/gh/pymovements/pymovements/blob/tests%2Ffull-coverage/src%2Fpymovements%2Fdataset%2F_utils%2F_archives.py#L228 where each version can only cover one branch.
If we want the codecov badge to be 100% it would need to say 100% on the linked branch. it has some `pragma` comments, but it does not seem to be effective.

Other than that, I would accept.

cbueth (approved):
We have reached 💯% flat!

PR #1713, opened 2026-09-07
fix: merge overlapping events before summing durations in `events2timeratio`

### LINKED ISSUE #1661: events2timeratio double-counts overlapping events, ratio can exceed 1.0
## Bug

`events2timeratio` (and therefore `Gaze.measure_events_ratio`) sums event durations without merging overlapping intervals. When two same-name events overlap, the overlap is counted twice, and the resulting ratio can exceed 1.0.

This affects all binocular data: events are recorded per eye, so same-name events from the left and right eye (blinks, fixations, saccades) typically overlap in time. Any per-name ratio computed on binocular events is silently overestimated, up to ratios above 1.0.

## Reproduction

```python
import polars as pl
from pymovements.transforms import events2timeratio

# left-eye blink [0, 96] fully contains right-eye blink [6, 85]
events = pl.DataFrame({
    'name': ['blink', 'blink'],
    'onset': [0.0, 6.0],
    'offset': [96.0, 85.0],
})
samples = pl.DataFrame({'time': [float(t) for t in range(105)]})

samples.select(events2timeratio(events, samples, 'blink', sampling_rate=1000.0))
```

| | blink ratio |
|---|---|
| actual: durations summed, `((96 - 0 + 1) + (85 - 6 + 1)) / 105` | 177 / 105 ≈ 1.686 |
| expected: overlap merged first, `(96 - 0 + 1) / 105` | 97 / 105 ≈ 0.924 |

## Existing tests

PR #1608 added parametrized tests pinning this down on two levels, each with a passing case asserting the current double-counted value and a `strict` xfail case asserting the correct merged value:

- `tests/unit/transforms/segmentation_test.py::test_events2timeratio_overlapping_events`
- `tests/functional/data_loss_measures_test.py::test_from_asc_binocular_overlapping_blinks_events_ratio`

Fixing this issue means the xfail cases start passing: remove the xfail marks and drop the `current_behavior_overlap_double_counted` parametrizations.

## Suggested fix

Merge overlapping intervals of the matching events before summing durations in `events2timeratio`. Notes:

- The per-trial branch (`trial_columns`) needs the same treatment; merging must happen within each trial group.
- `events2segmentation` already detects overlaps via the `_has_overlap` helper in `src/pymovements/transforms/segmentation.py` and warns about them, so detection machinery exists in the module; `events2timeratio` needs actual merging.
- Merging should probably be the default (a ratio above 1.0 is never meaningful), so no opt-in parameter seems necessary.

## Tasks

- [ ] Merge overlapping same-name event intervals in `events2timeratio`, including the `trial_columns` branch
- [ ] Remove the xfail marks and current-behavior parametrizations from the two tests above
- [ ] Check whether the docstrings of `events2timeratio` and `Gaze.measure_events_ratio` need updating (formula currently documents a plain sum)

#### Context

- Follow-up to #1608, see review discussion
- Related: #1584

cbueth:
Fixes #1661

### Problem
`events2timeratio` (and `Gaze.measure_events_ratio`) summed event durations without merging overlapping intervals. Binocular data records same-name events per eye that typically overlap in time, so the overlap was counted twice and ratios could exceed 1.0.

### Changes
- `_merged_event_durations`: vectorized polars merge of overlapping same-name intervals via sort + running-max offset + `shift`/`cum_sum` interval detection. Applies `.over(trial_columns)` when grouping so merging happens within each trial.
- `events2timeratio` now uses this helper in both the session-level and per-trial branches, so overlapping intervals are merged before durations are summed.
- Updated the `events2timeratio` and `Gaze.measure_events_ratio` docstrings to document the merge.
- Dropped the `current_behavior_overlap_double_counted` parametrizations and removed the xfail marks from the two tests added in #1608
- updated the `event_ratio_test.py` overlapping case to the corrected value

### Example
For left-eye blink `[0, 96]` containing right-eye blink `[6, 85]`, the ratio is now `(96-0+1)/105 approx 0.924` instead of `(96+1 + 85-6+1)/105 approx 1.686`.


dkrako (approved):
Thank you so much for this fix @cbueth!

The merging approach is exactly right and handles all the tricky cases correctly: overlapping, touching, and fully contained events each end up counted only once.

I added a few small things on top rather than bouncing the PR back:

- a three-event overlap test case (`[0, 96]`, `[6, 85]`, `[90, 100]` merging into `[0, 100]`) that pins the running-max logic, since all existing overlap cases had only two events,
- a per-trial test case (trial 1 `[0, 6]`; trial 2 `[1, 2]` and `[4, 5]`) that pins the `.over(trial_columns)` isolation of the merge,
- inlined the leftover single-case parametrize wrappers from the dropped `current_behavior_overlap_double_counted` cases,
- a tiny pre-existing docstring correction in `Gaze.measure_events_ratio`: the no-`sampling_rate` formula now reflects that the sampling interval is always applied (estimated from the data when no rate is available).

Thanks again for the clean fix, merging once CI is green.

PR #1726, opened 2026-09-09
fix(plotting): break tsplot lines at temporal gaps; don't plot time as a channel

### LINKED ISSUE #1684: tsplot draws a continuous line across temporal gaps because it ignores the time column
## Current Behavior

`tsplot()` plots every channel against the sample index (`t = np.arange(n_samples)`) and never
uses the `time` column. As a consequence, missing samples are handled inconsistently:

- If missing samples are present as rows with null values, the nulls become NaN via
  `DataFrame.to_numpy()` and matplotlib correctly renders a gap in the line.
- If missing samples are absent rows (e.g. the tracker omits samples during blinks, or the user
  cleaned the data with `Gaze.drop_nulls()`), the samples on either side of the temporal gap sit
  at consecutive indices. The plot then draws a continuous line straight across the gap and the
  x-axis is silently compressed, so the temporal structure of the signal is misrepresented.

Additionally, when `channels=None`, the `time` column is plotted as a channel of its own
(a diagonal line), instead of serving as the x-axis.

## Expected Behavior

Temporal gaps in the recording should be visible as gaps in the plotted lines, regardless of
whether missing samples are encoded as null rows or as absent rows. The x-axis should reflect
time, not the sample index.

- `tsplot()` uses the `time` column as x-axis if present (falling back to sample index otherwise).
- Absent rows produce visible line breaks, e.g. by inserting NaN at time discontinuities.
- `time` is not plotted as a channel when `channels=None`.
- Tests cover both missing-data encodings (null rows and absent rows).

### Steps to Reproduce

```python
import polars as pl
import pymovements as pm

# samples 30..59 missing as absent rows (tracker gap, or after drop_nulls())
df = pl.DataFrame({
    'time': [i for i in range(100) if not (30 <= i < 60)],
    'pixel': [[float(i), 2.0 * i] for i in range(100) if not (30 <= i < 60)],
})
gaze = pm.Gaze(samples=df)
pm.plotting.tsplot(gaze, channels='pixel', savepath='tsplot_gap.png')

itzzdev09:
Closes #1684

## What

Two fixes for `plotting.tsplot()`:

### 1. Lines no longer drawn across temporal gaps

`tsplot()` connected consecutive samples with a straight line regardless of their `time` values. When missing samples are encoded as **absent rows** (a tracker dropping samples during a blink, or data cleaned with `Gaze.drop_nulls()`), the samples on either side of the gap sit at adjacent positions, so the plot drew a continuous line straight across the gap and the temporal structure was misrepresented.

`_insert_gap_breaks()` runs after the time axis is built: a step larger than `1.5 x` the median positive sample step is treated as a gap, and a `NaN` sample is inserted at that discontinuity so matplotlib stops the line. Regularly-sampled data and gaps already encoded as null rows are unaffected.

### 2. `time` is no longer auto-selected as a channel

Since #1637 the `time` column is a `Duration` column, and the `channels=None` selector (`is_numeric() or isinstance(dtype, pl.Duration) or ...`) began picking it up as a plottable channel — producing a spurious diagonal subplot. `time` is the x-axis, so it is now excluded from auto-selection. Passing `channels=['time']` explicitly still works.

## Tests

`pytest tests/unit/plotting/tsplot_test.py` -> **86 passed** (82 existing + 4 new):

- `test_tsplot_breaks_line_at_gap_from_absent_rows` - one NaN break inside the gap, between the last pre-gap and first post-gap sample
- `test_tsplot_keeps_gap_from_null_rows_visible` - null-row gaps stay visible (unchanged path)
- `test_tsplot_no_gap_leaves_samples_untouched` - regular grid: no inserted points, no NaN
- `test_tsplot_does_not_plot_time_column_as_channel`

One existing test (`test_tsplot_default_channels_unnest_list_columns`) asserted the old behaviour of `time` appearing as the first channel; updated to match.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/pymovements/plotting/tsplot.py
+++ b/src/pymovements/plotting/tsplot.py
@@ -112,13 +112,20 @@ def tsplot(
         If array has more than two dimensions.
     """
     if channels is None:
-        # Select all numeric (and nested numeric) channels
+        # Select all numeric (and nested numeric) channels. The ``time`` column
+        # is the x-axis, not a signal, so it is never auto-selected even though
+        # it is a (Duration) numeric column.
         channels = [
             c
             for c in gaze.samples.columns
-            if gaze.samples[c].dtype.is_numeric()
-            or isinstance(gaze.samples[c].dtype, pl.Duration) or (
-                gaze.samples[c].dtype == pl.List and gaze.samples[c].dtype.inner.is_numeric()
+            if c != 'time'
+            and (
+                gaze.samples[c].dtype.is_numeric()
+                or isinstance(gaze.samples[c].dtype, pl.Duration)
+                or (
+                    gaze.samples[c].dtype == pl.List
+                    and gaze.samples[c].dtype.inner.is_numeric()
+                )
             )
         ]
 
@@ -184,6 +191,10 @@ def tsplot(
         if isinstance(time_series.dtype, pl.Duration):
             time_series = duration_to_ms(time_series)
         t = time_series.to_numpy()
+        # Break the line at temporal gaps so that missing samples encoded as
+        # absent rows (tracker gaps, drop_nulls()) show up as gaps instead of
+        # a straight line drawn across them.
+        t, arr = _insert_gap_breaks(t, arr)
     else:
         t = np.arange(arr.shape[1])
     xlims = t.min(), t.max()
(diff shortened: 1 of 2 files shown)


itzzdev09:
**This one is specifically about answering cbueth's request**
 
image-1789481613067
-----------------------------------------------------------------------------------

image-1789481607137


Here are the before/after plots for the case with samples missing from 120–179 ms.

Before (main): the pixel_x and pixel_y traces are connected across the missing interval, and the time column is also plotted as a channel.

After (this PR): the missing interval is preserved as a gap in pixel_x/pixel_y, and the time column is no longer plotted as a channel.

itzzdev09:
@cbueth please review whenever u free

cbueth:
Thank you @itzzdev09 for the illustrative comments with plots. Currently, I am fully loaded with other tasks. Maybe one of the other maintainers can find a moment. Are you having any dependent project that needs this asap?

I suppose this one was not ought to be copied over 🤭 
> **This one is specifically about answering cbueth's request, so put the screenshots directly in the PR comment.**

itzzdev09:
Haha yes, that was a note to myself that slipped in, sorry about that! No rush at all, nothing depends on it. Thanks for taking a look, happy to wait for whoever has time
@cbueth

PR #1735, opened 2026-09-22
Feat/videoreplay fixationcorrection

Di0nigi:
Fixation correction method integration from the videoreplay repo https://github.com/aeye-lab/pymovements-videoreplay/tree/main for a uzh master project, tests still missing.


dkrako:
I am a bit confused about this PR. I think we discussed in the meeting that fixation correction is already implemented in #1625 which was finally merged last week. What does this PR bring to the table?

